Issue #4, opened 2020-08-13
Convert a few GLSL shaders to Rust


repi:
We talked about today that we should convert a few of our existing rendering shaders to Rust to see how well we could express it, even if we can't compile it to SPIRV or run them yet. Having a few GLSL or HLSL shaders here in the repo that we also have a manually converted Rust versions of, initially maybe just rough Rust psuedo code but can also try and build them in a Rust CPU crate project to actually compile it.

Can take for example the Ark mesh and sky shaders, they are quite basic and fine to duplicate to this repo for testing.

This would be a helpful guide to see what functions we would want in a shader core crate, that doesn't exist in Rust std, also for easy porting of existing shaders. And will be a guide to see what areas of Rust are the least ergonomic for shaders where we may want to have Rust language shaders to simplify it.

XAMPPRocky:
I think this can be closed now on the rust-gpu side? We have converted quite a few shaders to Rust.

PR #14, opened 2020-08-19
Sketch simple shaders

Jasper-Bekkers:
These are GLSL shaders we're using in Ark and versions converted into pseudo-rust.

I didn't touch on bindings yet, because I think that'll be something we can debate endlessly about.

This uses a `glam-rs` inspired math library since it seems most likely that we'll use something like that. 

None of this compiles / builds, but it should give a bit of an overview of how these shaders will end up looking. I think many things can be improved so I'll go through and comment on a bunch of things.

Part of #4


Jasper-Bekkers:
Now that we have specific shaders to test with, we should probably close this issue and address ergonomics on a case-by-case basis.

PR #92, opened 2020-10-22
WGPU example

### LINKED ISSUE #65: Add example using gfx-hal or wgpu
In addition to the existing example using `ash`, which is quite explicit and low-level with Vulkan, it would also be very nice to have an example using [`gfx-hal`](https://github.com/gfx-rs/gfx) which would be cross-platform and likely smaller and cleaner example.

@termhn has volunteered :)

VZout:
Up to y'all if you want this in.
Just copy pasted the `wgpu` example and made it rusty.

Mainly just wrote it so I can patch it into main locally for testing.

Fix #65


expenses:
The `Cargo.toml` for the runner seems to be out of date with `main.rs` because `wasm-bindgen-futures` isn't listed as a dependency.

repi:
Wait we don't need a different shader right? Can just use the same one. 

And then maybe call it `example-runner-wgpu` and `example-runner-ash`?

DJMcNab:
I've mostly made it work with the old shader in https://github.com/DJMcNab/rust-gpu/tree/wgpu-example, although the output I get is upside down. I'm not sure why that is 🤔.

Thanks to the people on the discord for helping me get it this far.

VZout:
> Wait we don't need a different shader right? Can just use the same one.
> 
> And then maybe call it `example-runner-wgpu` and `example-runner-ash`?

We don't but I made this more for myself since I want a tiny as possible shader for my own testing. This I can just patch into my branches and test/debug easily.

Cargo.toml
```diff
@@ -1,7 +1,9 @@
 [workspace]
 members = [
     "examples/example-runner",
+	"examples/wgpu-example-runner",
```
bjorn3, line 4:
```suggestion
    "examples/wgpu-example-runner",
```
nit: you used a tab instead of four spaces like the rest of this file.

repi:
On my Windows machine the `wgpu-example-runner` example works, but on my Mac I just get the following when running which is rather odd:

```
[0.902503 ERROR]()(no module): Function retrieval error "Function \'main_vs\' does not exist"
[0.902600 ERROR]()(no module): Invalid shader entry point 'main_vs': InvalidEntryPoint
thread 'main' panicked at 'failed to create graphics pipeline: Failed to create pipeline: Unsupported usage: Implementation specific error occurred', /Users/johan/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-core-0.6.5/src/device/mod.rs:2898:30
```

Not really sure at all why it would be failing to find even the entrypoint of the shader

grovesNL:
@repi this seems to be a bug, we'll fix it in https://github.com/gfx-rs/gfx/issues/3435

grovesNL:
@repi we just published a patch release, so this should work fine after updating with `cargo update` or `cargo update -p gfx-backend-metal`

repi:
@grovesNL awesome thx, will try out!

repi (approved):
Works with Mac now, yay! And merged it with the latest changes.

Think we should get this in! And once it is merged we could do some renaming & moving things around a bit.

Cargo.toml
```diff
@@ -2,7 +2,9 @@
 members = [
     "examples/example-runner-cpu",
     "examples/example-runner",
+    "examples/wgpu-example-runner",
     "examples/example-shader",
+    "examples/wgpu-example-shader",
```
termhn, line 7: Adding a comment here, one thing that I would change but I think could be done after merging, is to rename this shader to something like `hello-world-shader` or `simplest-shader` or something to that effect, since it shouldn't really be specific to wgpu.

termhn (approved):
With the caveat of my one comment, think this seems fine to merge for now... I can take stewardship of making this a little more integrated with the other example(s) afterwards. I would like to get #109 merged as well which then paves the way for the wgpu runner to run the sky example.

Cargo.toml
repi: fully agree!

I was thinking that the current `example-shader` would become more, well simply, `shaders` with a set of non-trivial shaders like the current sky one and more in the same crate.

but also very good then to have a super simple separate crate that is easier to iterate on with fewer dependencies and features. esp. soon when glam gets in #149

.github/workflows/ci.yaml
```diff
@@ -46,6 +46,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - run: sudo apt install libwayland-cursor0 libxkbcommon-dev libwayland-dev
+      - run: brew install spirv-tools
```
khyperia, line 49: wait what? this should have excluded wgpu-example-runner in the clippy run, not installing spirv-tools, right? either that or including example-runner in the clippy run

repi: not sure what you mean?

Did get clippy build failures on this branch on that spirv-tools was missing

bjorn3: Why would you use brew on a linux job?

XAMPPRocky: @bjorn3 It's the only package manager with `spirv-tools` on linux that is available in GitHub Actions.

khyperia: > Why would you use brew on a linux job?

See also https://github.com/EmbarkStudios/rust-gpu/issues/107

PR #109, opened 2020-10-23
Cleanup pass on example shader and vk example-runner

termhn:
* Removes vertex and index buffers; instead, generates position data based on vertex index in the shader
* Removes depth buffer as it's not being used anyway
* Removes the unused vertex_color input from vertex shader
* Renames the wrongly-names out_color output in the vertex shader to `out_pos` with corresponding `in_pos` in the frag shader, which is what it's being used for (that's the data that gets wired to the fragment shader input at location 0, because the builtin "position" output is separate). We could also just scrap this output and use `#[spirv(builtin = "frag_coord")]` in the fragment shader but I think it's helpful to have there in the example.
* Renames a couple of variables in the fragment shader for better clarity
* Uses negative viewport height (flips viewport) which brings clip space to parity with `wgpu`'s (and DirectX's) clip space, useful if we want to directly use the same shader in #92  because otherwise, it will be flipped
* Uses sRGB-encoded surface rather than doing odd sorta-gamma correction in shader
* Replaces the weird tone mapping step and uncharted 2 tone mapping operator with a new (better ;) ) one https://www.desmos.com/calculator/igytv8ck36

image


hrydgard:
Nice! But I think you'll end up needing the depth buffer later, for future more complex examples. Also, vertex input I'd definitely keep around for the same reason - that still needs to work with Rust shaders, so why not keep it so it gets tested?

I do of course agree that removing this stuff makes the example code a lot nicer and shorter so if you intend to have a second bigger example-runner later, keeping this smaller one like this might be worth it.

examples/example-shader/src/lib.rs
```diff
@@ -160,29 +151,37 @@ pub fn main_fs(input: Input<Vec4>, mut output: Output<Vec4>) {
...
 #[allow(unused_attributes)]
 #[spirv(entry = "vertex")]
 pub fn main_vs(
-    in_pos: Input<Vec4>,
-    _in_color: Input<Vec4>,
-    #[spirv(builtin = "position")] mut out_pos: Output<Vec4>,
-    mut out_color: Output<Vec4>,
+    #[spirv(builtin = "vertex_index")] vert_idx: Input<i32>,
+    #[spirv(builtin = "position")] mut gl_pos: Output<Vec4>,
+    mut out_pos: Output<Vec4>,
```
Jasper-Bekkers, line 172: Could we call these `out_pos` and `out_pos_interpolator` or something? I would love us to not have the  `gl_` prefix anywhere if we can avoid it.

Jasper-Bekkers (changes requested):

examples/example-shader/src/lib.rs
```diff
@@ -133,23 +131,16 @@ fn sky(dir: Vec3, sun_position: Vec3) -> Vec3 {
...
-    tex_color += Vec3::new(0.0, 0.001, 0.0025) * 0.3;
+    let tex_color = lin + l0;
 
     // Tonemapping
-    let white_scale = 1.0 / uncharted2_tonemap(TONEMAP_WEIGHTING);
-    let curr = uncharted2_tonemap(((2.0 / LUMINANCE.pow(4.0)).log2()) * tex_color);
-    let color = curr * white_scale;
-
-    color.pow(1.0 / (1.2 + (1.2 * sunfade)))
+    let color = tex_color.min(Vec3::splat(1024.0));
```
Jasper-Bekkers, line 137: I'm not sure about most of what's going on here, for example, I'm not even sure why the sky model chose to embed the tone-mapping operator in here to begin with (or even why it's doing srgb conversions tbh). It might be nicer to hoist all of that out of `sky` all together? 

One thing I do notice is that your change seems to remove the application of `sunfade`. I suspect this was done to make the sun seem brighter.

termhn: > I'm not even sure why the sky model chose to embed the tone-mapping operator in here to begin with
> ...
> It might be nicer to hoist all of that out of sky all together?

Yeah it would... the tonemapping is really just because the sky is being displayed directly... so it makes sense to put it in the main frag shader and not in `sky()`.

and while it does remove sunfade from here, I suspect that the application of `sunfade` here was really just a hacked in "auto brightness slider" (which really shouldnt be done as a gamma anyway) based on a vaguely related value... it's still being used in the calculation of the rayleigh coefficient right after it gets defined.

termhn:
> Nice! But I think you'll end up needing the depth buffer later, for future more complex examples. Also, vertex input I'd keep around for the same reason - that still needs to work with Rust shaders, so why not keep it so it gets tested?
> 
> I do of course agree that removing this stuff makes the example code a lot nicer and shorter so if you intend to have a second bigger example-runner later, keeping this smaller one like this might be worth it.

while I agree that it's probable we'll want these things at some point, I'm a bit skeptical of having a unified example-runner that's capable of doing lots of things at this point... and in any case, it's quite easy to add those back in the same state they are now, since it's just copied from the `ash` examples

khyperia:
Throwing in my two cents, I'm strongly in favor of nuking the `ash` example and just going with `wgpu`. I think the only reason we're still on `ash` is because we were waiting for you to join and do the work, @termhn :P - sorry, probably should have made that more clear, so you didn't put too much work into cleaning the ash example up.

Although, Viktor [already started on the wgpu example](https://github.com/EmbarkStudios/rust-gpu/pull/92), could use that as a base

(So I'm in favor of doing whatever to the ash example right now, because I imagine it'll be gone soon)

termhn:
> Throwing in my two cents, I'm strongly in favor of nuking the `ash` example and just going with `wgpu`. I think the only reason we're still on `ash` is because we were waiting for you to join and do the work, @termhn :P - sorry, probably should have made that more clear, so you didn't put too much work into cleaning the ash example up.
> 
> Although, Viktor [already started on the wgpu example](https://github.com/EmbarkStudios/rust-gpu/pull/92), could use that as a base
> 
> (So I'm in favor of doing whatever to the ash example right now, because I imagine it'll be gone soon)

I'm sorta of the same mind haha. And don't worry, didn't put too much work into the ash one, the main purpose originally was actually just to change it so that the sky shader example works with no friction with the wgpu example Viktor started :D

repi:
> So I'm in favor of doing whatever to the ash example right now, because I imagine it'll be gone soon

I would like the `wgpu` example to be the primary one as it is nicer and faster to use and built on more of a Rust stack.


But do think we'll have to keep a `ash` + MoltenVK example and test suite as we'll need a lower level mechanism where we are not blocked by the `gfx-rs` stack, the WebGPU standard nor `gfx-portability` for Vulkan and SPIRV translation to Metal and MetalSL respectively. MoltenVK is problematic to build but it is a lot more mature than `gfx-portability` and I do think we will need to be able to test on both.

termhn:
Rebased on main and applied changes in order to normalize output between all three example runners. Also renamed `example-shader` to `sky-shader` and `wgpu-example-shader` to `simplest-shader` and addressed previous review comments from @Jasper-Bekkers

khyperia:
Looks like there's still a fair number of references to `example-shader` in docs and comments and whatnot: `README.md`, `examples/wgpu-example-runner/build.rs`, `examples/example-runner/build.rs`, and `docs/src/introduction.md`

Also there are two `Cargo.lock` files not at the root that were accidentally added in #92, can you remove those?

repi:
The `Cargo.lock` files are gone on `main` in e2be0580157d1a73dbaf8b1d86386075518998c7

khyperia:
Also, optional thing, [winit is out of date](https://deps.rs/repo/github/EmbarkStudios/rust-gpu) (`0.22.1` -> `0.23.0`), feel free to update it here, or we can do it afterwards

Jasper-Bekkers:
What do we think about @hrydgard's comment?

> Nice! But I think you'll end up needing the depth buffer later, for future more complex examples. Also, vertex input I'd keep around for the same reason - that still needs to work with Rust shaders, so why not keep it so it gets tested?

I tend to agree with him - keeping it is probably best (which is why I +1'd it), but there may be other opinions here?

termhn:
I think my opinion is https://github.com/EmbarkStudios/rust-gpu/pull/109#issuecomment-715460275 and @khyperia 's is https://github.com/EmbarkStudios/rust-gpu/pull/109#issuecomment-715462782

termhn:
@XAMPPRocky made a couple docs changes in #38d2735 that might be relevant for your review

Jasper-Bekkers:
Oh @termhn this PR should probably update the screenshot in the README.md as well to the new one, would be less confusing for new people.

XAMPPRocky (approved):

termhn:
Pushing the big green button manually 'cause mergify can't merge this one :) 🎉

PR #111, opened 2020-10-23
Fix translation from AtomicOrdering to MemorySemantics

### LINKED ISSUE #100: ordering_to_semantics not correctly implemented
https://github.com/EmbarkStudios/rust-gpu/blob/f4f103197694290c469ad00d8f156ea7d0bfc292/rustc_codegen_spirv/src/builder/builder_methods.rs#L55-L67

According to https://www.khronos.org/blog/comparing-the-vulkan-spir-v-memory-model-to-cs the following must change:

* `Acquire` must use `MakeVisible | Acquire`
* `Release` must use `MakeAvailable | Release`
* `AcqRel` must use `MakeAvailable | MakeVisible | AcquireRelease`
* According to the SPIR-V specification `SeqCst` must not be used for Vulkan.
    https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#_a_id_memory_semantics__id_a_memory_semantics_lt_id_gt
    > If the declared memory model is Vulkan, SequentiallyConsistent must not be used.

khyperia:
Fixes #100

FYI @bjorn3


rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -116,6 +104,40 @@ fn memset_dynamic_scalar(
...
+            AtomicOrdering::AcquireRelease => {
+                MemorySemantics::MAKE_AVAILABLE
+                    | MemorySemantics::MAKE_VISIBLE
+                    | MemorySemantics::ACQUIRE_RELEASE
+            }
+            AtomicOrdering::SequentiallyConsistent => {
+                let emit = self.emit();
+                let memory_model = emit.module_ref().memory_model.as_ref().unwrap();
+                if memory_model.operands[1].unwrap_memory_model() == MemoryModel::Vulkan {
+                    invalid_seq_cst = true;
```
bjorn3, line 124: It is simpler to directly return the zombie here I think.

khyperia: eh, there's some weird RefCell already borrowed shenanigans going on if structured that way - the `memory_model` line takes a ref to the module, and then creating the zombie takes another borrow, so that means the borrow needs to be dropped before that, and it turns into a weird fragile mess.

(really dislike that codegen_ssa requires RefCells everywhere, but, oh well...)

Jasper-Bekkers:
Going over this, it appears that `Unordered` should not exist in `rustc` since it's specifically added to LLVM to support the Java memory model which Rust doesn't support.

Also it looks like `Consume` is commented out in the rustc backend, so we might want to add a comment here as well to promote it to `MakeVisible | Acquire` as per the page referenced if it ever becomes available.

khyperia:
> Going over this, it appears that `Unordered` should not exist in `rustc` since it's specifically added to LLVM to support the Java memory model which Rust doesn't support.

It's definitely [still created](https://github.com/rust-lang/rust/blob/a9cd294cf2775441e713c7ee2918b728733b99f5/compiler/rustc_codegen_ssa/src/mir/intrinsic.rs#L408) though, right? We don't care about whether or not it's reasonable for rustc to emit it (perhaps file a rustc bug for that), we only care if rustc emits it.

bjorn3:
There are intrinsics for the unordered versions of atomic operations, but they aren't used by `Atomic*`.

Jasper-Bekkers:
> It's definitely still created though, right? We don't care about whether or not it's reasonable for rustc to emit it (perhaps file a rustc bug for that), we only care if rustc emits it.

I started filing a rustc ticket under the assumption Rust followed the C++ memory model, turns out it doesn't, so disregard anything I've said about this :-)

Jasper-Bekkers (approved):

PR #114, opened 2020-10-23
Fail unimplemented intrinsics instead of incorrect behavior

### LINKED ISSUE #99: Prefer not implementing intrinsics in a way that breaks in edge cases
For example `saturating_add` isn't implemented in a saturating way:

https://github.com/EmbarkStudios/rust-gpu/blob/f4f103197694290c469ad00d8f156ea7d0bfc292/rustc_codegen_spirv/src/builder/intrinsics.rs#L88-L97

This is more of a lesson I learnt the hard way than an issue with rust-gpu itself. While it may seem like a good idea at first to get it quickly working, it will backfire in the long term. I have spent quite a lot of time debugging miscompilations in cg_clif caused by exactly this kind of shortcuts. At minimum I think you should add a `FIXME` every time you add such a shortcut.

khyperia:
Fixes #99. Zombies are things in libcore, errors are things that aren't.

FYI @bjorn3

Also cleaned up some other minor stuff while in the area.


rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -568,6 +566,18 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
     simple_uni_op! {neg, s_negate}
     simple_uni_op! {fneg, f_negate}
 
+    fn exactudiv(&mut self, lhs: Self::Value, rhs: Self::Value) -> Self::Value {
+        let result = self.udiv(lhs, rhs);
+        self.zombie(result.def, "exactudiv is not supported yet");
```
bjorn3, line 571: You don't need a zombie here. `exactudiv` is the same as `udiv` except that it is UB when there is a remainder and in a few other cases. It is fine to just ignore the UB. You may want to add a FIXME to take advantage of the UB in the future.

khyperia: ah! neat :sparkles:

bjorn3 (approved):

Jasper-Bekkers (approved):

khyperia:
@Mergifyio refresh

Issue #31, opened 2020-09-15
Add bindings for spirv-opt


Jake-Shadle:
We will be relying on [spirv-opt](https://github.com/KhronosGroup/SPIRV-Tools/tree/master/source/opt) for actually making the generated spirv code go brrrrr, but right now there are no rust bindings for it, so we can't call it from Rust like we need to.

There are a couple of complications to this.

1. The code is all C++, and there's no C API
1. CMake, because of course

For the first point, the actual API for running the optimizer seems to be fairly [trivial](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/tools/opt/opt.cpp), so I see 2 paths forward for this.

1. Write a small C wrapper around the C++ API and write manual bindings for it
1. Try out [cxx](https://github.com/dtolnay/cxx) though my guess right now is that it is overkill

As for CMake, just as with our PhysX bindings, we will obviously just build it with cc so that downstream users won't hate us.

I'm putting this issue here for now, but I am thinking that this would probably be its own repo which contains the bindings/wrapper crates in it, and it would initially just be for the spirv-opt part of spirv-tools, but obviously could add the other tools as needed.

Issue #107, opened 2020-10-22
Improve CI build times, esp. on Ubuntu


repi:
These are our current CI build times through our GitHub Action:

- Mac: 3m 9s
- Windows: 8m 22s
- Ubuntu: 13m 22s

Ubuntu is spending 8 min just installing spirv-tools, this we should fix so we at least get down to 8m total build times instead of 13 min.

Just download the spriv-tools executables directly instead of using apt? think was some discussion in chat. cc @Jake-Shadle

Jake-Shadle:
It was originally downloading the executables directly, but was changed in #63 to use brew instead, not sure why exactly, but you can see the CI runs before that PR were around ~6m which is still fairly high, but a lot better.

repi:
Oh, does sound like we should go back!

khyperia:
It looks like we're building at least some crates twice from the split `cargo test`. My guess:

current ci.yaml:

```yaml
      - run: cargo test --workspace --exclude example-runner
      - if: ${{ runner.os != 'macOS' }}
        run: cargo test --workspace example-runner
```

probably better ci.yaml:

```yaml
      - if: ${{ runner.os != 'macOS' }}
        run: cargo test --workspace
      - if: ${{ runner.os == 'macOS' }}
        run: cargo test --workspace --exclude example-runner
```

khyperia:
now that CI is downloading spirv-tools instead of brew installing (and therefore building) them, is there anything left to do in this issue?

current times:

mac: 10m (2m installing spirv-tools, 7m running tests)
ubuntu: 9m30s (45 seconds installing native deps+spirv-tools, 7m30s running tests)
windows: 11m30s (13 seconds installing spirv-tools, 9m running tests)

seems like most time is spent in running tests (mostly building our boatload of dependencies, I think), so not really much we can improve on the CI side.

Jake-Shadle:
We could add support for sccache, but that does add some complexity, I could do a quick and dirty pr to see if the speedup is worth it?

khyperia:
I don't actually know, is sccache significantly better than the github cache action? We used to do that, but it was broken since rustup upgrades wouldn't invalidate the cache, so we removed it. Might be relatively easier to fix that instead. https://github.com/EmbarkStudios/rust-gpu/blob/8c9583f2c6e1fe69a34eac15c5591a1ef54dc166/.github/workflows/ci.yaml#L17-L26

[...]

Issue #48, opened 2020-10-10
"Build system"


khyperia:
This is a tracking issue for what I've been calling the "build system", but in reality the full task list is larger than just that.

In general, this is a group of related issues that would be great for someone other than me to own and drive forwards. These tasks are spread across various "functionality levels", see that issue for more information.

- Create a crate to be used by users build.rs, that takes a path to a shader crate and spits back a spir-v binary that can be `include!`d (or perhaps a better designed version of this). I've created an extremely rough and minimally-featured prototype in the [spirv-builder](https://github.com/EmbarkStudios/rust-gpu/tree/main/spirv-builder) crate - we need to expand it quite a bit, in both robustness and features.
- Integrate spirv-opt and spirv-val into statically linkable libraries that are included in rustc_codegen_spirv. This means the user will not have to install spirv-tools on their system. Tracking issue: https://github.com/EmbarkStudios/rust-gpu/issues/31
- Generate host-side bindings, tracking issue: https://github.com/EmbarkStudios/rust-gpu/issues/10. I like the approach of "SYCL compiles down it's GPU side C++ code along with some side-cart metadata first, this metadata is then converted into regular host side C++". This means that rustc_codegen_spirv would spit out some form of metadata for what bindings are in the module, and then the build.rs helper crate would generate rust source files from that metadata. (Or perhaps rustc_codegen_spirv would spit out rust files directly and the builder would just include them, but, whatever, exact behavior to be determined later. Having multiple backends - e.g. ash and vulkano both - complicates this.)
- Possibly adding any hooks/etc. to that generated code to enable the rendergraph/flamegraph support.

XAMPPRocky:
I've been thinking about this quite a bit in past couple of days, in relation to testing the crate, as right now we use `spirv-builder` for the tests and it complicates the testing process for a lot of the same reasons it's complicated for users. I think there's a few modifications to how we're building that would dramatically simplify the process. First I want to outline who we are targeting with the build system.

- **Users** People or organisations who want to use rust-gpu in their projects but don't want to contribute or modify themselves. Builds should be easy for them to setup and use in their projects.
- **Contributors** People or organisations who want to use and contribute to rust-gpu, they want to be able to build and test their changes. Builds should as close to CI as possible.
- **CI** Continuous Integration wants to be able to easily build and test entirely from scratch. Builds should be fast and easy to re-produce for debugging.

---

[...]

Issue #32, opened 2020-09-16
Do we use spirv-val?


khyperia:
(for end users, for development we're definitely going to use it) - this came up in https://github.com/EmbarkStudios/rust-gpu/issues/31

Say, for example, that the user does something invalid under the current capabilities - e.g. use OpBitcast on a pointer without the Addresses capability, or a storage class that's unsupported. We have two ways of detecting this error:

1) Detect this when we're compiling, and give an error immediately.
2) Run spirv-val on the result binary, and report any errors as rust errors.

There's a trade-off here: 1 takes way more work, and 2 produces lower quality errors. So, opening up a design issue to talk about it.

---

Of course, the answer might be, as always, a mix of the two. For example, we could *try* to detect all errors while we're compiling, but realize that there will be holes, and still run spirv-val on the result - and when spirv-val detects something that we don't, add in a check for that in the compiler.

PR #117, opened 2020-10-23
Replace usage of spirv-* binaries with spirv-tools rust crate

Jake-Shadle:
Making this a draft so that others can leave comments, I think I need to add a wrapper around the Validator as otherwise we will still require having [spirv-tools] installed which is kind of lame when we are already paying the compile time cost for building spirv-tools

## Additions
* Adds [spirv-headers] and [spirv-tools] as git submodules (will add docs about retrieiving/updating them for contributors)
* Adds `spirv-tools-sys` which builds `spirv-tools`, including a small generator script that generates some includes based on [spirv-headers]
* Adds a `spirv-tools-sys/src/c/opt.cpp` which is just a thin (and incomplete) C wrapper around [Optimizer](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/include/spirv-tools/optimizer.hpp) because C++
* Adds a `spirv-tools` crate which provides a slighly nicer API on top of `spirv-tools-sys`

## Changes
* Replaces the usage of the `spirv-opt` binary in link.rs with `spirv_tools::opt::Optimizer`
* Replaces the usage of `spirv-val` for validating resulting binaries in link.rs with `spirv_tools::val::Validator`
* Replaces the usage of `spirv-as` in the linker tests with `spirv_tools::as::Assembler`
* All of the crates that use rustc_codegen_spirv now have a `use-installed-tools` and `use-compiled-tools` feature flags, where `use-compiled-tools` is the default feature for optimal user experience.

## Missing
* There is one change I had to do in our [spirv-tools](https://github.com/EmbarkStudios/SPIRV-Tools/commit/83bb6bacf26160e0b04c092255ac0bbb9ae42e48) fork due some incredible strangeness with libstdc++ streams, so right now the `Assembler` will choke on hexadecimal floats, but we only use the `Assembler` in tests and none of them currently use hexadecimal floats, so figured this would be ok for now, the issue is fixable but doing C++ code was making me sad so I punted until it became an actual issue.

Resolves: #31 

[spirv-headers]: https://github.com/KhronosGroup/SPIRV-Headers
[spirv-tools]: https://github.com/KhronosGroup/SPIRV-Tools


XAMPPRocky:
Personally I would prefer if we could avoid submodules, they are a pain to maintain, and I would even say they're anti-pattern. I think it would be a lot better if we could use `git subtree` for pulling in those repositories instead. They're newer and much easier to maintain as it prevents a lot of foot guns of updating submodules and keeping the repository in sync with them. 

### Subtree docs
- https://github.com/git/git/blob/master/contrib/subtree/git-subtree.txt
- https://www.atlassian.com/git/tutorials/git-subtree

khyperia:
oh no, personally I hate `git subtree` and would much prefer submodules :P

Jasper-Bekkers:
+1 to submodules.

```rust
// The spirv tools use generated code, for now we just replicate the minimum
// generation we need here by calling the *shudders* python script(s) we need
// to in a simple script and commit them to source control, as they only need
// to be regenerated when spirv-headers is updated 
```

This deserves a shout-out here (or anywhere) since this adds a new manual step.

XAMPPRocky:
@khyperia Can I ask the motivation behind that? You're the first person I've met who prefers them. In my experience submodules are hard to use right, and more work to maintain. Every contributor has to keep their submodules in their local version in sync individually, and you can end up in a state where new changes in the submodule break your code, but you don't find out until after both have landed in each. With `git subtree` it's a lot easier to ensure that you can always build your codebase, and makes so that most contributors don't have to care about it since the only time contributor would need to sync is when they want to update `spirv-tools`, and they would also have to ensure that the code continues to pass the tests. From the git docs:

> Unlike submodules, subtrees do not need any special constructions (like .gitmodules files or gitlinks) be present in your repository, and do not force end-users of your repository to do anything special or to understand how subtrees work. A subtree is just a subdirectory that can be committed to, branched, and merged along with your project in any way you want.

Jake-Shadle:
> This deserves a shout-out here (or anywhere) since this adds a new manual step.
Yah I will add docs for this. To be clear, all the generated files are in checked in in spirv-tools-sys/generated so they aren't built every time, but yes, if spirv-headers is changed you will need python etc.

> Is there any way we can keep this SPIRV-Headers and the one used by rspirv in sync easily other then manually checking hashes?
Good question, we can at minimum assert in the build script if their commits differ, or maybe, just the files that we read are the same but...yah.

Jake-Shadle:
So good news, spirv-opt and spirv-val are now compiled as crates and can be run without installing spirv-tools, yay!
...bad news, extremely [slow to compile](https://github.com/EmbarkStudios/rust-gpu/runs/1299856051?check_suite_focus=true), oh and I manage to get an [exception](https://github.com/KhronosGroup/SPIRV-Tools/issues/3992) that seems particular to libstdc++ on both my local machine and the ubuntu runner...

So not sure how we want to handle this.

1. We could install spirv-tools on CI only, and just feature toggle the spirv-tools crate off so we don't take the compile hit, but this introduces a really big diff between the local and CI that would potentially be draining and annoying.
1. We could move our CI to buildkite and throw more cores/sccache etc at it, as it shouldn't really take that long to build (1m11s on my machine), but that has the drawback of being a little more overhead for us and contributors since it's not a common CI, though I don't think that's a strong argument as the important green = good, red = bad, here are some logs are all the same

repi:
Would prefer to stay with GitHub Actions for now at least, it is more accessible for users and has a wider ecosystem and easier if anyone forks or tries out things and this is a very open project so that is a priority.

Still need decent CI build times though (#107)  as we are working on a lot of different areas through small PRs here. Think it would be fine to have CI just download the executables, esp. if it saves massive amount of time. Users likely will want to be able to run locally with pre-built or their own tools as well, but good to have a default that builds everything (if it doesn't take forever)

repi:
Long term wanted future: Rust native versions of esp. the parts of spirv-tools we use. Not sure how realistic

Jake-Shadle:
> Think it would be fine to have CI just download the executables, esp. if it saves massive amount of time. Users likely will want to be able to run locally with pre-built or their own tools as well, but good to have a default that builds everything (if it doesn't take forever)

So I think maybe we can make it slightly less terrible but having a feature flag on spirv-tools for "use executables" vs "use compiled sys crate" so the interface would be the same, just internally it would use one or the other, so would greatly lessen the pain of maintenance, at least for a while.

> Long term wanted future: Rust native versions of esp. the parts of spirv-tools we use. Not sure how realistic

Given the commit frequency (though most of the recent ones seem focused on spirv-fuzz which I don't know if we have any interest in) that might be tough, not necessarily to port what is currently there, but to keep up with features, new versions etc.

repi:
> feature flag on spirv-tools for "use executables" vs "use compiled sys crate"

Agreed, that would likely be a decent compromise

And yeah also think it would hard to keep up with spirv-tools in general for a Rust port. But if we primarily use it to optimize the spirv that is a much smaller scope. fuzzing and validation are more dev features that are needed now but shouldn't be needed by other devs that are using Rust GPU once it is more mature. I think.

Jake-Shadle:
Yah, I guess the good thing is that right now the way the spirv-tools optimizer is structured in passes, it _might_ not be too bad to port passes as needed, and since we control the build system, pick the rust pass or the c++ pass, though I haven't looked a ton at how much data/state is shared between passes (hopefully little).

khyperia:
> And yeah also think it would hard to keep up with spirv-tools in general for a Rust port. But if we primarily use it to optimize the spirv that is a much smaller scope. fuzzing and validation are more dev features that are needed now but shouldn't be needed by other devs that are using Rust GPU once it is more mature. I think.

Some context, we decided in https://github.com/EmbarkStudios/rust-gpu/issues/32 to always run spirv-val on user's machines, and have no plans to stop doing so for a long time.

repi:
> Some context, we decided in #32 to always run spirv-val on user's machines, and have no plans to stop doing so for a long time.

Ah good to know, and makes sense, esp at this early stage.

Would ultimately want users to be able to build shaders without any external C/C++ dependencies or system executables, but that can be a larger target as part of long term design of #48 and potentially other separate issues

Jake-Shadle:
Ok, I've replaced all of the usages of spirv-* binaries, now I just need to add them back behind a feature toggle for the spirv-tools crate so that we can just use those in CI, but make them an optional requirement for users.

XAMPPRocky:
Another potential footgun with submodules that doesn't matter much now, is that having different submodules checked out doesn't count as "dirty" to git/cargo so you could publish accidentally publish the wrong commit. Here's steps how to replicate an example of how to that.

1. Change the branch in `spirv-tools`'s `.gitmodules` to `master`.
2. Run `git submodules update --remote`
3. Run `cargo package`

Screenshot 2020-10-26 at 07 54 42

Jake-Shadle:
I've updated the top of the PR with the one remaining TODO of fixing up documentation, but everything else should be mostly done.

rustc_codegen_spirv/src/link.rs
```diff
@@ -1,27 +1,31 @@
...
+};
+use std::{
+    collections::HashSet,
+    env,
+    ffi::{CString, OsStr},
+    fs::File,
+    io::{Read, Write},
+    path::{Path, PathBuf},
+    sync::Arc,
+};
```
khyperia, line 28: wow, what a rustfmt `merge_imports = true`! is this standard across embark repos, or was this accidental? (not sure if it's anywhere else in this diff, being consistent might be good)

rustc_codegen_spirv/src/link.rs
```diff
@@ -124,97 +128,86 @@ fn link_exe(
...
-            diagnostics.note(&String::from_utf8(output.stderr).unwrap());
-        }
+    let validator = val::create(None);
 
-        diagnostics.emit();
-        None
+    if validator.validate(spv_binary, None).is_err() {
+        let mut err = sess.struct_err("error occurred during validation");
+        err.note(&format!("module {:?}", filename));
+        err.emit();
```
khyperia, line 210: just checking, does this emit the error message? maybe the tool already prints it? (maybe add a comment if it does print it, since this is a little confusing if so)

spirv-tools-sys/generate.rs
```diff
@@ -0,0 +1,115 @@
+// The spirv tools use generated code, for now we just replicate the minimum
+// generation we need here by calling the *shudders* python script(s) we need
+// to in a simple script and commit them to source control, as they only need
+// to be regenerated when spirv-headers is updated
+
+use std::{fs, process::Command};
+
+fn python<S: AsRef<std::ffi::OsStr>>(args: impl IntoIterator<Item = S>) -> Result<(), i32> {
+    Command::new("python")
```
khyperia, line 9: I'm tempted to use PyO3 here to be cute and neat, but considering this is a manual script run by developers and so we can assume python is installed, probably not worth the effort (especially if it gets built by default and so explodes build time due to building PyO3)

spirv-tools/src/cmd.rs
```diff
@@ -0,0 +1,192 @@
...
+    let mut child = cmd.spawn().map_err(CmdError::BinaryNotFound)?;
+
+    if let Some(input) = input {
+        use std::io::Write;
+
+        child
+            .stdin
+            .take()
+            .unwrap()
+            .write_all(input)
```
khyperia, line 89: oh wow, I definitely wouldn't have thought of piping the input, this is super cool!

khyperia (commented):
Holy heck this is a lot, thank you so much for doing this! This is pseudo-approval, but I don't want mergify to shove this in before you have the chance to check optional feedback.

spirv-tools-sys/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+build = "build.rs"
+license = "MIT OR Apache-2.0"
+
+[features]
+# Using this feature disables the compilation in the build script, but
+# preserves the types so that spirv-tools can still work without needing
+# to keep copies of some of the basic enums etc
+use-installed-tools = []
+# Forces compilation of the C++ code, even if `use-installed-tools` is enabled
+use-compiled-tools = []
```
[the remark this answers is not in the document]
Jake-Shadle: They're intentionally not mutually incompatible so that they can be tested in tandem in the future. And (hopefully) progressively replaced.

spirv-tools/src/cmd.rs
Jake-Shadle: Heh yes, a bit nicer than writing a temporary file. :slightly_smiling_face:

spirv-tools-sys/generate.rs
Jake-Shadle: Yah, would be a bit overkill in this case, but could be fun for someone to PR in the future. :wink:

rustc_codegen_spirv/src/link.rs
Jake-Shadle: Hah no, this was just a manual drive by change when I was changing this file, I can revert if you feel like it's noise.

.github/workflows/test.sh
```diff
@@ -0,0 +1,55 @@
...
+# See: https://github.com/EmbarkStudios/rust-gpu/issues/84
+if [[ -z "${CI}" && "$os" != "macOS" ]]; then
+    cargo_test examples/example-runner
+fi
+
+cargo_test examples/wgpu-example-runner
+
+cargo_test_no_features examples/example-runner-cpu
+cargo_test_no_features examples/example-shader
+cargo_test_no_features examples/wgpu-example-shader
```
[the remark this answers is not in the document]
Jake-Shadle: Yah, I think this is fine for now since the "no tests" takes basically 0 time, but still get the full build to validate against.

rustc_codegen_spirv/src/link.rs
khyperia: I thiiink I'd prefer it reverted, if it's not too much work? Like, stylistically, I don't mind either way, but what I do mind is consistency, and everywhere else has it separated. (I realllly want the [`module`](https://github.com/rust-lang/rustfmt/issues/3362#issuecomment-642564290) rustfmt merge_imports option to stabilize, then we could set it in rustfmt.toml...)

Jake-Shadle: No worries, easy to revert, I just get sad when I see duplicates in module imports. :wink:

Jake-Shadle:
Oh, one thing I forgot to ask, should this PR making running the `Optimizer` on by default? I just kept the logic of it being gated behind an environment variable, but that seems dangerous since it doesn't get CI coverage then.

khyperia:
> Oh, one thing I forgot to ask, should this PR making running the `Optimizer` on by default? I just kept the logic of it being gated behind an environment variable, but that seems dangerous since it doesn't get CI coverage then.

In the long-term, I think so, however, it gets a little in-depth - e.g. we want rustc_codegen_spriv tests (currently located at `spirv-builder/src/test/*`) to not run spirv-opt. I thiiink it'd be nicest to do all that configuration plumbing after this PR, since this is already really big.

Jake-Shadle:
Very true! Ok, should I just update the docs in this PR and we'll call it done and merge it in?

khyperia (approved):
yeah!

Jake-Shadle:
Encountered an issue with Windows that I had missed because the tests weren't running where either the use of stdin for input or stdout for output of the program caused problems for the spirv-tools, so until I look into that deeper I just returned to using files for the input and output just so I can get this PR landed.

PR #118, opened 2020-10-23
Add framework for compiler tests

khyperia:


spirv-builder/src/test/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+#![feature(lang_items, register_attr)]
+#![register_attr(spirv)]
+use core::panic::PanicInfo;
+use spirv_std::*;
+#[panic_handler]
+fn panic(_: &PanicInfo) -> ! {
+    loop {}
+}
+#[lang = "eh_personality"]
+extern "C" fn rust_eh_personality() {}
```
bjorn3, line 41: You should be able to avoid this if you pass `-Cpanic=abort` to rustc.

khyperia: There's a whole complex backstory here, I would explain, but, currently dealing with some horrible garbage (see _fltused windows CI failure, which is the latest item on the heap of garbage), so, bleh :(

PR #124, opened 2020-10-23
Enable clippy on spirv-std

repi:
And some smaller rustdoc markdown fixes for clippy.

Fun to run clippy on GPU code :)


khyperia (approved):
woo! I'll probably follow up at some point and fix some of these, e.g. `use_self`, but whatever, this is great for now!

repi:
Yeah I was about to do that also, as I generally like the `Self`, but was quite a bit of stuffs to change in the math

PR #125, opened 2020-10-23
Add cargo-deny config & CI step

repi:
To verify:
- That  we are only using compatible permissive licenses for all of our dependencies
- Don't use multiple duplicate versions of dependencies (have to opt-in to).
- Don't use dependencies with vulnerabilities in them

Also fixed some workspace crates that didn't have our license set


.github/workflows/ci.yaml
```diff
@@ -46,3 +46,8 @@ jobs:
       - run: rustup component add rustfmt clippy rust-src rustc-dev llvm-tools-preview
       - run: cargo fmt --all -- --check
       - run: cargo clippy --workspace --exclude example-runner --all-targets -- -D warnings
+  cargo-deny:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
```
Jake-Shadle, line 52: Should use v2

Jake-Shadle (approved):
Nice!

.github/workflows/ci.yaml
repi: good catch!

PR #126, opened 2020-10-23
Re-enable CI on pushes to all branches

repi:
I think @XAMPPRocky disabled this recently, but generally we want CI to be enabled on all branches (we have in all of our other repos) so one can iterate on branches without having PRs for them and get CI coverage.

Does cause some redundant building, but that is fine.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,8 +1,4 @@
-on:
-  push:
-    branches:
-      - main
-  pull_request:
+on: [push, pull_request]
 
 name: Continuous integration
 


khyperia:
This runs CI *twice* on PRs, though, and is a really unpleasant user experience (I think I requested the change)

repi:
Ah right it does, for our local branch pushes at least, does show up twice in the GitHub PR interface, though does not for PRs on forks.

We do use this for all of our other repos AFAIK, and at least I sometimes use the workflow to be able to push a branch in the background and get CI coverage (which likely will extra useful here when we have larger test suite), but can be fine to just do a draft PR also to get coverage.

We can skip this for now and see how it works, it is not too much of an issue I think, and cleaner GitHub interface is nice also.

Doesn't seem to be a way in GitHub to solve that issue, as the builds are technically building from different git ref. Though wish the build status UI at least would be better.

XAMPPRocky:
GitHub does allow patterns in what branches are enabled, so we could have a naming scheme like `feature-*` to indicate that it's a branch that you want to run CI when you're working on it.

Issue #134, opened 2020-10-24
Add support for using `glam` crate in shaders


repi:
## Goal

At Embark we use [`glam`](https://github.com/bitshifter/glam-rs) as our primary math library in our native and WASM code and are quite happy with it. 

We want to bring up the crate in Rust GPU so it can be built and used with our new `spirv-unknown-unknown` target in the Rust GPU compiler backend, and once that works likely change the example shaders here to use it, or have separate ones that uses and tests it.

In our own Rust engine we plan to convert our shaders to Rust GPU from GLSL and want to use the same math crate on all sides, or at least be able to.

## Bring up

I've talked briefly with @bitshifter (the `glam` maintainer) and he was open to the idea of support, but we'll have to see what is needed for it. 

May be multiple issues to resolve in Rust GPU first in its early stage to get this to work, which is why we are tracking this on our side first here. Once have a first compiling shader we can open a draft PR on glam with it and use it as a git dependency from there. May also take time until it can be merged in and tested as part of glam mainline if non trivial changes are required (or to get it it compiling in glam CI).

@bitshifter do you have any additional thoughts or concerns about how we may approach this? We don't want to take up too much of our time and do think we have a bunch of stuff we need to try and fix first.

## Alternatives

Also note that this work does not mean that `glam` should be the only or the _de facto_ math library for shaders in Rust. 

We hope that most Rust math crates could eventually have support and we'll likely also be interested in exploring specific adaptions to make it adapted and ergonomic for GPU and shader programming (like vector swizzles) or even Rust language changes. But it would be great to have a standard well-designed crate, that we already use, to focus on and bring up first to have as a base.

bitshifter:
Hi @repi, I'm excited that you are taking `glam` along on this journey to bring Rust to GPU! Bringing the `crates.io` experience to shader authoring sounds great and allows other crates, including other math libraries, to innovate in this space.

Certainly working in your own fork to get things off the ground and drafting a PR sounds like a good approach. I don't have any immediate concerns. If there was a need to make breaking API changes it would be good to discuss those if that came up.

The other thing would be if it was possible to build and test GPU code on standard free CI infrastructure, if it wasn't it would make maintenance a bit trickier.

bjorn3:
Mesa has support for software rendering using the llvmpipes backend. It should be possible to use this for testing.

khyperia:
I just looked into this for two minutes, and wanted to throw in my two cents:

[...]

Issue #81, opened 2020-10-21
Intercept calls to (rust) libm and emit a builtin instead of the soft impl


khyperia:
Plan is to go with @bjorn3's comment below and expose the GLOps not in std::intrinsics via intercepting calls to https://github.com/rust-lang/libm, which is used in e.g. `glam`'s `no_std` implementation.

Old title: Expose GLOps not in std::intrinsics

https://docs.rs/spirv_headers/1.5.0/spirv_headers/enum.GLOp.html , same with CLOp.

For example, `acos`.

repi:
Ran into this today with `tan` that glam is using (#134), it is not part of the Rust core intrinsics functions. Any idea of how to add our own intrinsics, do we have to do that in core or as another set of intrinsics for the specific target?

bjorn3:
libstd directly calls the libm functions. Adding an extra intrinsic requires a rustc change to add it to the list of intrinsics of typeck.

repi:
Suspected that, definitely don't want to use `libm` on the GPU for where there is real GPU/SPIRV intrinsics to use

bjorn3:
It should be possible to intercept calls to functions with specific names and instead emit a SPIRV instruction.

khyperia:
Fixed by https://github.com/EmbarkStudios/rust-gpu/pull/264

PR #141, opened 2020-10-25
Add example to evaluate sky shader on CPU 

repi:
Adds a new example that shows how to compile and evaluate a simple fragment shader (our sky shader) directly on the CPU.

For this specific shader as it doesn't use any resources it is very easy as one can just call the fragment shader function manually, later for standard shaders with resource bindings this will be much harder, but maybe still useful to test parts of shaders or reuse some functionality.

Right now it also just works for the current temp math functions that use `core` intrinsic, as all of those are available obviously on the CPU also, later when we start introducing GPU & SPIRV specific intrinsics (#81), this approach won't work, and then we will want to use a higher-level math layer in the shaders for the math instructions (glam: #134) so they can be mapped to the right available target intrinsic.

To make the example very simple, slim and cross-platform, this example uses the `minifb` crate to drive the Window. Could switch to something else later though but it is very basic.

image

### DIFF
--- a/examples/example-runner-cpu/src/main.rs
+++ b/examples/example-runner-cpu/src/main.rs
@@ -6,6 +6,9 @@ use std::time::Instant;
 fn color_u32_from_vec4(v: Vec4) -> u32 {
     let convert = |f: f32| -> u32 { (f.min(1.0).max(0.0) * 255.0).round() as u32 };
 
+    // this later will also do linear -> sRGB gamma correction for output,
+    // but currently our shader output is in sRGB so no conversion needed
+
     convert(v.z()) | convert(v.y()) << 8 | convert(v.x()) << 16 | convert(v.w()) << 24
 }
 
--- a/examples/example-runner-cpu/src/main.rs
+++ b/examples/example-runner-cpu/src/main.rs
@@ -31,7 +31,7 @@ fn main() {
         .map(|i| {
             let screen_pos = vec2(
                 (i % WIDTH) as f32 / WIDTH as f32 * 2.0 - 1.0,
-                (i / HEIGHT) as f32 / HEIGHT as f32 * 2.0 - 1.0,
+                (i / WIDTH) as f32 / HEIGHT as f32 * 2.0 - 1.0,
             );
 
             // evaluate the fragment shader for the specific pixel
--- a/examples/example-runner-cpu/src/main.rs
+++ b/examples/example-runner-cpu/src/main.rs
@@ -1,0 +1,54 @@
+use minifb::{Key, Window, WindowOptions};
+use rayon::prelude::*;
+use spirv_std::{vec2, Vec4};
+use std::time::Instant;
+
+fn color_u32_from_vec4(v: Vec4) -> u32 {
+    let convert = |f: f32| -> u32 { (f.min(1.0).max(0.0) * 255.0).round() as u32 };
+
+    convert(v.z()) | convert(v.y()) << 8 | convert(v.x()) << 16 | convert(v.w()) << 24
+}
+
+fn main() {
+    const WIDTH: usize = 1280;
+    const HEIGHT: usize = 720;
+
+    let mut window = Window::new(
+        "Rust GPU - CPU shader evaluation",
+        WIDTH,
+        HEIGHT,
+        WindowOptions::default(),
+    )
+    .expect("Window creation failed");
+
+    // Limit to max ~60 fps update rate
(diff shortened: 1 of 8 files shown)


examples/example-shader/src/lib.rs
```diff
@@ -168,7 +190,15 @@ pub fn main_fs(input: Input<Vec4>, mut output: Output<Vec4>) {
...
 
-    output.store(color.extend(0.0))
+    color.extend(1.0)
+}
+
+#[allow(unused_attributes)]
+#[spirv(entry = "fragment")]
+pub fn main_fs(input: Input<Vec4>, mut output: Output<Vec4>) {
+    let v = input.load();
+    let color = fs(Vec2::new(v.0, v.1));
```
khyperia, line 200: This makes me think that we should talk about the unwrapping of Inputs/Outputs more - I've flipflopped back and forth on this, there's arguments both ways. Also, we should be able to construct Intputs/Outputs on the CPU, right? (maybe need to make something `pub`, `pub fn construct_from_cpu_var(&mut T) -> Output<T>` or something)

examples/example-runner-cpu/src/main.rs
```diff
@@ -0,0 +1,54 @@
...
+    window.limit_update_rate(Some(std::time::Duration::from_micros(16600)));
+
+    let start_time = Instant::now();
+
+    let buffer = (0..WIDTH * HEIGHT)
+        .into_par_iter()
+        .map(|i| {
+            let screen_pos = vec2(
+                (i % WIDTH) as f32 / WIDTH as f32 * 2.0 - 1.0,
+                (i / HEIGHT) as f32 / HEIGHT as f32 * 2.0 - 1.0,
```
khyperia, line 34: I... thought this should be `(i / WIDTH)`, and I'm confused how this works at all (is the image shifted slightly compared to the GPU version?), but maybe it's just me just waking up.

khyperia (approved):
this is *super* cool, and impressively simple!

> later when we start introducing GPU & SPIRV specific intrinsics (#81), this approach won't work

It *should* work, I plan on implementing it in a way where it should work.

examples/example-runner-cpu/src/main.rs
repi: Just you waking up I think :) this is correct and works. Can likely be made cleaner in a separate functions or so maybe later, but does work

examples/example-shader/src/lib.rs
repi: Yes the current `Input` and `Output` was not ideal for this, but do think it is good to have this use case at least and we can think about the design separately

examples/example-runner-cpu/src/main.rs
bjorn3: It took me a while to wrap my head around this, but what happens is the following: The `i % WIDTH` generates an x coordinate in the range `0..WIDTH`, the `/ WIDTH` turns it into the range `0.0 .. 1.0`, the `* 2.0 - 1.0` then turns it into the range `-1.0 .. 1.0`. Same for the y coordinate.

khyperia: okay no I'm more awake now and it's *not* just me waking up, this *should* be `(i / WIDTH) as f32 / HEIGHT as f32 * 2.0 - 1.0`.

This PR's CPU code:

cpu_original

GPU code:

gpu

This PR's CPU code but with the line changed to `i / WIDTH`:

cpu_changed

it is indeed "the image shifted slightly compared to the GPU version"

repi: Hah awesome, I was overconfident that my code worked, well it did work just not correctly. Thanks!!

Jasper-Bekkers (approved):
I love that we can do this :-)

repi:
Ah some cargo-deny stuff to fix with the minifb dependency

repi:
@Mergifyio rebase

examples/example-runner-cpu/src/main.rs
```diff
@@ -0,0 +1,54 @@
+use minifb::{Key, Window, WindowOptions};
+use rayon::prelude::*;
+use spirv_std::{vec2, Vec4};
+use std::time::Instant;
+
+fn color_u32_from_vec4(v: Vec4) -> u32 {
```
emilk, line 6: Maybe a comment here about why it is OK to ignore sRGB here (or add a TODO about respecting sRGB) ?

repi: done

PR #142, opened 2020-10-26
Clarify breaking changes and deprecation

Jasper-Bekkers:
Came up in the chat; it's useful to clarify these things to our potential users so they know where they stand with this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,10 @@ We meet weekly over a Discord call to discuss design and triage issues. Each mee
 
 We have a [#rust-gpu Discord channel](https://discord.gg/dAuKfZS) for fast discussion and collaboration.
 
+## Backwards compatibility, breaking changes and deprecation
+
+Right now because the project is in a very early state of development, we might introduce temporary changes as stop-gap measures, or implement features or APIs that might not work exactly in a way we end up liking. Therefor it is expected that some (if not most) of the user facing code will change and evolve over time. At the moment this means that we make no guarantees about backwards compatibility and have no formal deprecation model in place. Effectively meaning that in v0.01 we only support the code that is in our repository. We appreciate our early adopters and would ask them to evolve their code along with ours.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,6 +76,10 @@ We meet weekly over a Discord call to discuss design and triage issues. Each mee
 
 We have a [#rust-gpu Discord channel](https://discord.gg/dAuKfZS) for fast discussion and collaboration.
 
+## Backwards compatibility, breaking changes and deprecation
+
+Right now because the project is in a very early state of development, we might introduce temporary changes as stop-gap measures, or implement features or APIs that might not work exactly in a way we end up liking. Therefor it is expected that some (if not most) of the user facing code will change and evolve over time. At the moment this means that we make no guarantees about backwards compatibility and have no formal deprecation model in place. Effectively meaning that in v0.01 we only support the code that is in our repository. We appreciate our early adopters and would ask them to evolve their code along with ours.
```
XAMPPRocky, line 81: Fixed a typo, and clarified the versioning/release, since I believe we aren't currently publishing release artifacts.
```suggestion
Right now because the project is in a very early state of development, we might introduce temporary changes as stop-gap measures, or implement features or APIs that might not work exactly in a way we end up liking. Therefore it is expected that some (if not most) of the user facing code will change and evolve over time. At the moment this means that we make no guarantees about backwards compatibility and have no formal deprecation model in place. Effectively meaning that currently we only support building from source with the latest `main` branch in our repository. We appreciate our early adopters and would ask them to evolve their code along with ours.
```

repi (approved):
Looks good (with the typo fixes).

nit: should it be "stage" instead of "state" in "a very early state of development"? I guess both works

arirawr (approved):

repi:
@mergify rebase

repi:
@Mergifyio rebase

PR #145, opened 2020-10-26
Refactor attributes and add descriptor_set/binding

khyperia:
Whoof, this turned out bigger than expected. A lot of it is necessary refactoring to clean things up, though, not the actual adding of descriptor_set/binding.

Could someone try out this branch on a shader and try setting a uniform? I haven't tested it with `example-runner` since I'm not sure of the arcane vulkan shenanigans needed to bind a uniform on the CPU side.

The syntax looks like this:

```rust
pub fn main_fs(
    input: Input<Vec4>,
    #[spirv(descriptor_set = 2, binding = 5)] uni: Uniform<Vec4>,
    mut output: Output<Vec4>,
) {
```

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -1,3 +1,4 @@
 # Summary
 
 - [Introduction](./introduction.md)
+- [Attribute syntax](./attributes.md)
--- a/docs/src/attributes.md
+++ b/docs/src/attributes.md
@@ -1,0 +1,48 @@
+# Attribute syntax
+
+rust-gpu introduces a number of SPIR-V related attributes to express behavior specific to SPIR-V not exposed in the base rust language.
+
+There are a few different categories of attributes:
+
+## Entry points
+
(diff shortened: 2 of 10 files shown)


repi:
@Mergifyio rebase

Jasper-Bekkers (changes requested):
Maybe you already incorporated this feedback I gave in the discord but I couldn't find it in the PR.

>alright yeah then it's fine; just in case it might be worth while to reserve descriptor set 0 for our use, that way we can auto-gen id's at least into set 0 and avoid potential clashes while we transition over the the actual binding model

XAMPPRocky:
With the refactor, would it be possible to also add a chapter in the `rust-gpu` dev guide that lists these attributes and their basic behaviour?

khyperia:
> Maybe you already incorporated this feedback I gave in the discord but I couldn't find it in the PR.

Oh, I missed this comment, I thought we decided with the explicit system seeing as it's the most flexible _and_ easiest solution. I don't think putting extra effort into making it clean is worth it, that effort should be put into the bindings rfc implementation.

Jasper-Bekkers (approved):

khyperia:
uuh, that was not supposed to be merged, there were still reviews pending?!?

Issue #34, opened 2020-09-18
Vector types


khyperia:
spir-v has vector types that are dramatically useful on e.g. nvidia hardware for optimizing loads/stores. At first, let's just directly expose vector types, instead of trying to do things like auto-vectorization.

1) Rustc already has the notion of vector types, and we directly translate them to spir-v vector types. We need to investigate under what conditions rustc emits vector types, and if we can re-use the same framework. https://github.com/EmbarkStudios/rust-gpu/blob/9697c227401ffdb02f935ddf860060d6f9978700/rustc_codegen_spirv/src/abi.rs#L332-L339
2) If we cannot re-use the same framework, [create an attribute](https://github.com/EmbarkStudios/rust-gpu/blob/9697c227401ffdb02f935ddf860060d6f9978700/rustc_codegen_spirv/src/symbols.rs#L97) to specify a type is a builtin vector, recognize the attribute in `abi.rs` ([example for storage classes](https://github.com/EmbarkStudios/rust-gpu/blob/9697c227401ffdb02f935ddf860060d6f9978700/rustc_codegen_spirv/src/abi.rs#L547)), and emit a vector type. Perhaps do a mix of the two, e.g. if rust supports f32x2 and f32x4, but not f32x3, then we need to define our own f32x3.
3) Investigate the spec: is emitting a spir-v "OpTypeVector" enough, or are there additional decorations/etc. we need to apply to get it to be a proper "vector type"? https://github.com/EmbarkStudios/rust-gpu/blob/9697c227401ffdb02f935ddf860060d6f9978700/rustc_codegen_spirv/src/spirv_type.rs#L166
4) Implement the above in `spirv-std`, adding things like swizzling ops and other shenanigans (can come later).

This seems like a good not-quite-first-but-kinda-first task for someone other than me to do. Step 2 is a little involved and will likely need some mentorship - `abi.rs` is hecking *wild* - but getting more people who know how `abi.rs` works would be really really good.

nipunG314:
Hi, I would be interested with working on this.

From what I understand in this thread, the first thing to do would be to write the operators and impls for f32x4 using macros?

khyperia:
Awesome!

This issue is a bit out of date, though: https://github.com/EmbarkStudios/rust-gpu/pull/71 added quite a few vector items. However, it wasn't very cleanly or thoroughly implemented, and needs significant work. Some tasks remaining are:

[...]

PR #147, opened 2020-10-26
Enable Clippy use_self lint in spirv-std

nipunG314:
As mentioned in #34, the use_self lint is currently disabled in spirv-std.

There is one exception to this. I had to introduce an exception in mod.rs because there is bug that confuses core::intrinsics with f32. I have opened a bug report for it over on the Clippy repo [here](https://github.com/rust-lang/rust-clippy/issues/6221). It seems there a few related errors where this also happens.

```rust
#[allow(clippy::use_self)]
impl MathExt for f32 {
    fn pow(self, factor: f32) -> f32 {
    ...
}
```

I'm not sure whether enabling the use_self lint is a good idea or not. The feature doesn't seem to be bug-free.

### DIFF
--- a/spirv-std/src/lib.rs
+++ b/spirv-std/src/lib.rs
@@ -11,7 +11,7 @@
     clippy::enum_glob_use,
     clippy::pub_enum_variant_names,
     clippy::mem_forget,
-    //clippy::use_self,
+    clippy::use_self,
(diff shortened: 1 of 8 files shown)


khyperia (approved):
Thanks so much! :tada:

PR #158, opened 2020-10-27
Implement copysign for shaders

### LINKED ISSUE #148: Implement copysignf32
We have an implementation of it for OpenCL Kernel compute mode, but not for the "GL Compute" mode where it looks like there is no 1-to-1 mapping for it.

```
  error: TODO: Shader copysign not supported yet
    --> spirv-std\src\math_ext.rs:82:18
     |
  82 |         unsafe { core::intrinsics::copysignf32(self, sign) }
     |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

This is needed by glam (#134) as `f32::signum` uses it

khyperia:
Fixes #148

### DIFF
--- a/rustc_codegen_spirv/src/builder/intrinsics.rs
+++ b/rustc_codegen_spirv/src/builder/intrinsics.rs
@@ -2,6 +2,7 @@ use super::Builder;
 use crate::abi::ConvSpirvType;
 use crate::builder_spirv::SpirvValueExt;
 use crate::codegen_cx::CodegenCx;
+use crate::spirv_type::SpirvType;
 use rspirv::spirv::{CLOp, GLOp};
 use rustc_codegen_ssa::mir::operand::OperandRef;
 use rustc_codegen_ssa::mir::place::PlaceRef;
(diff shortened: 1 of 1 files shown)


repi (approved):
Yay!

repi:
Would be good to have tests for these type of stuffs, later on :)

rustc_codegen_spirv/src/builder/intrinsics.rs
```diff
@@ -240,7 +241,33 @@ impl<'a, 'tcx> IntrinsicCallMethods<'tcx> for Builder<'a, 'tcx> {
                 if self.kernel_mode {
                     self.cl_op(CLOp::copysign, [args[0].immediate(), args[1].immediate()])
                 } else {
-                    self.fatal("TODO: Shader copysign not supported yet https://github.com/EmbarkStudios/rust-gpu/issues/148")
+                    let arg0 = args[0].immediate();
+                    let arg1 = args[1].immediate();
```
bjorn3, line 245: I think it would be more clear if you use `sign`, `sign_val`, `sign_src` or something like that instead of `arg1`.

rustc_codegen_spirv/src/builder/intrinsics.rs
```diff
@@ -240,7 +241,33 @@ impl<'a, 'tcx> IntrinsicCallMethods<'tcx> for Builder<'a, 'tcx> {
...
+                            self.constant_u32(1 << 31),
+                            self.constant_u32(u32::max_value() >> 1),
+                        ),
+                        64 => (
+                            self.constant_u64(1 << 63),
+                            self.constant_u64(u64::max_value() >> 1),
+                        ),
+                        _ => panic!("copysign not supported for width {}", width),
+                    };
+                    let arg0_int = self.bitcast(arg0, int_ty);
```
bjorn3, line 265:
```suggestion
                    let arg0_bits = self.bitcast(arg0, int_ty);
```

rustc_codegen_spirv/src/builder/intrinsics.rs
```diff
@@ -240,7 +241,33 @@ impl<'a, 'tcx> IntrinsicCallMethods<'tcx> for Builder<'a, 'tcx> {
...
+                    let arg1_int = self.bitcast(arg1, int_ty);
+                    let arg0_and = self.and(arg0_int, mask_value);
```
bjorn3, line 267:
```suggestion
                    let arg0_masked = self.and(arg0_int, mask_value);
```

rustc_codegen_spirv/src/builder/intrinsics.rs
```diff
@@ -240,7 +241,33 @@ impl<'a, 'tcx> IntrinsicCallMethods<'tcx> for Builder<'a, 'tcx> {
                 if self.kernel_mode {
                     self.cl_op(CLOp::copysign, [args[0].immediate(), args[1].immediate()])
                 } else {
-                    self.fatal("TODO: Shader copysign not supported yet https://github.com/EmbarkStudios/rust-gpu/issues/148")
+                    let arg0 = args[0].immediate();
+                    let arg1 = args[1].immediate();
+                    let width = match self.lookup_type(arg0.ty) {
+                        SpirvType::Float(width) => width,
+                        other => panic!(
```
XAMPPRocky, line 248: Is panicking appropriate here? Shouldn't this produce a compile error?

bjorn3: Typeck should already ensure that only floats get passed here. Using `unreachable!` or `bug!` makes sense here.

khyperia: Ah yeah, I guess I should have used the `bug!` macro - it's a bug/hard-panic, though, because it's breaking the invariants of what is known today (which is that the arguments to the `copysignf32` and `copysignf64` functions are floats). Similar with sizes that aren't 32 or 64 bits.

khyperia:
> Would be good to have tests for these type of stuffs, later on :)

Yeah! I did smoke-test it real quick, but proper tests would be nice. Also oh my goodness, I spent 5 minutes bending my brain trying to think of what `x = copysign(x, y)` would do for a xy=`-1..1` red/green fullscreen quad, until I realized that `example-runner-cpu` uses the native `copysign` implementation and I can just compare against that!

rustc_codegen_spirv/src/builder/intrinsics.rs
bjorn3: https://github.com/rust-lang/rust/blob/20b1e05a8d0e1773dc840a3286aa37916e87d84b/compiler/rustc_typeck/src/check/intrinsic.rs#L251-L252

PR #159, opened 2020-10-27
Fix compling the wgpu-example-runner on wasm32-unknown-unknown

expenses:
#92 was missing some dependencies needed for wasm compilation, so I've added them here. You can compile it with `RUSTFLAGS=--cfg=web_sys_unstable_apis cargo build --bin wgpu-example-runner --target wasm32-unknown-unknown`. It won't actually render on firefox nightly unless you downgrade to the [`gecko`](https://github.com/gfx-rs/wgpu-rs/tree/gecko) branch, which is lagging behind significantly. Haven't tested on chrome canary yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -313,6 +313,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "console_error_panic_hook"
+version = "0.1.6"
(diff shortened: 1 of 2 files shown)


khyperia (approved):
neat, thanks!

repi:
Cool!

PR #162, opened 2020-10-27
Download spirv-tools directly on ubuntu CI

khyperia:
I just pulled the latest download link, since I don't actually know how to get the proper download link for specific versions. Not sure if that's okay or not.


Jake-Shadle (approved):
Yes, that's what I had as well, as I couldn't find actual properly versioned links that tie to releases, they seem to just push binaries up directly from CI runs (which means each platform gets a unique arbitrary url! YAY!)

PR #168, opened 2020-10-28
Support GLSL450 and Simple memory models

### LINKED ISSUE #167: OpenGL compatibility
OpenGL supports binary SPIR-V shaders as source since 4.6, mostly feature compatible with Vulkan shader requirements.
Right now `rust-gpu` requires the `Vulkan` memory model which is not support. OpenGL only supports the GLSL450 model.
Would be interesting how deep the implementation relays on it. Quickly tweaking the capabilities and extensions towards GLSL results in a binary which can be passed to OpenGL.

(rather a feature request than a bug)

khyperia:
Fixes #167

Only smoke-tested building the example-shader shader on both Simple and GLSL450 models, seems to work. We might want to set up tests for both all the spir-v versions and this going forwards.


spirv-builder/src/lib.rs
```diff
@@ -115,16 +122,32 @@ fn invoke_rustc(
...
+        format!(" -C target-feature={}", target_features.join(","))
     };
     let rustflags = format!(
         "-Z codegen-backend={}{}",
         rustc_codegen_spirv.display(),
-        spirv_version_feture
+        feature_flag,
     );
-    let build = Command::new("cargo")
+    let build = Command::new("/home/khyperia/me/cargo/target/release/cargo")
```
XAMPPRocky: I don't think this going to work on other people's machines. 😄

spirv-builder/src/lib.rs
```diff
@@ -115,14 +122,30 @@ fn invoke_rustc(
...
+            match memory_model {
+                MemoryModel::Simple => "+simple",
+                MemoryModel::Vulkan => "+vulkan",
+                MemoryModel::GLSL450 => "+glsl450",
+            }
+            .to_string(),
+        );
+    }
+    let feature_flag = if target_features.is_empty() {
+        "".to_string()
```
XAMPPRocky, line 141: Pedantic nit:
```suggestion
        String::new()
```

rustc_codegen_spirv/src/builder_spirv.rs
```diff
@@ -71,17 +71,24 @@ pub struct BuilderSpirv {
...
+    ) -> Self {
         let mut builder = Builder::new();
         // Default to spir-v 1.3
         let version = version.unwrap_or((1, 3));
         builder.set_version(version.0, version.1);
+        let memory_model = memory_model.unwrap_or(MemoryModel::Vulkan);
         if kernel_mode {
             builder.capability(Capability::Kernel);
         } else {
             builder.extension("SPV_KHR_vulkan_memory_model");
```
msiglreith, line 87: is this extension actually needed?

khyperia: Yes, the vulkan memory model isn't present for spir-v versions less than 1.5. I'll fix this, though.

Jasper-Bekkers (approved):
I think this is fine to have for now, but I think it would be a mistake to try to support GLSL or anything other then the Vulkan Memory Model if we run into complex memory model issues. For example, if we discuss around subgroup operations, I think that topic might already be tricky enough without the additional implications that OpenGL / GLSL450 brings (if any).

PR #171, opened 2020-10-29
Refactor linker and delete some unused code

khyperia:
Use more efficient solutions than DefAnalyzer, and shuffle linker step order to remove need for ScalarType/AggregateType.

Considering that *not* using DefAnalyzer results in more efficient code in less lines of code (not even counting DefAnalyzer itself), this seems good.


Jasper-Bekkers:
I think this is a sound change; I think the way it import/export type checking used to work was because of the way the deduplicator used to work. Running this after dedup loses some of the type information for error reporting, but because it's significantly faster then that might be a win (and we might be able to recover that information another way).

Jasper-Bekkers (approved):

PR #176, opened 2020-10-29
Rough structurization of loops.

VZout:
This is a rough implementation of loops. There is a lot of duplicate code in functions like: `block_leads_into_break`, `is_parent_of` and more. For now I want to keep it like this until I start optimizing. At the moment this gives me more flexibility if I realize I'm missing requirements for certain cases and need to make changes.

This PR is far from perfect but I like to get this in so I can make more smaller PR's instead of a single HUGE one.

**IMPORTANT**
This PR allows looping control flow. NOT the rust `for i in 0..10`  since iterators are implemented as a switch. meaning you have to result to while loops instead until I introduce structurization for `OpSwitch`.

also run-time complexity is horrible 😅 

### Notes
* `break` is supported in "all" cases except if it results in a fully unreachable merge.
* `continue` is not supported (but is trivial to implement)
* `return` inside a loop not supported.
* Added a few tests.
* Moved `dce` and `structurize` before `mem2reg`

### DIFF
--- a/rustc_codegen_spirv/src/linker/structurizer.rs
+++ b/rustc_codegen_spirv/src/linker/structurizer.rs
@@ -337,7 +337,7 @@ fn block_is_parent_of(parent: Word, child: Word, blocks: &[Block]) -> bool {
     false
 }
 
-// Returns the idx of the branch that loops and the idx to the block that branches to the orignial block.
+// Returns the idx of the branch that loops and the idx to the block that branches to the original block.
 fn get_looping_branch_from_block(
     blocks: &[Block],
     cf_info: &ControlFlowInfo,
--- a/rustc_codegen_spirv/src/linker/structurizer.rs
+++ b/rustc_codegen_spirv/src/linker/structurizer.rs
@@ -1,17 +1,169 @@
 // This pass inserts merge instructions for structured control flow with the assumption the spir-v is reducible.
 
+// TODO: Could i simplify break detection by just checking, hey does the start branch branch to a merge block?
+// TODO: Verify we are never splitting a block that is queued for structurization.
+// TODO: are there any cases where I need to retarget branches or conditional branches when splitting a block?
+
 use super::id;
 use super::simple_passes::outgoing_edges;
 use rspirv::spirv::{Op, Word};
 use rspirv::{
-    dr::{Block, Function, Instruction, ModuleHeader, Operand},
+    dr::{Block, Instruction, Module, ModuleHeader, Operand},
     spirv::SelectionControl,
 };
 use rustc_session::Session;
 use std::collections::VecDeque;
 
-pub fn structurize(sess: Option<&Session>, header: &mut ModuleHeader, func: &mut Function) {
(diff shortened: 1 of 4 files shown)


spirv-builder/src/test/control_flow.rs
```diff
@@ -0,0 +1,162 @@
...
+#[allow(unused_attributes)]
+#[spirv(fragment)]
+pub fn main(i: Input<i32>) {
+    if i.load() > 0 {
+
+    } else if i.load() < 0 {
+
+    } else {
+        
+    }
```
bjorn3, line 144: Rustc may optimize this control flow away. Maybe `-Zmir-opt-level=0` should be used?

VZout: Discussed this with Ashley already :) 

`note they have the #[spirv(fragment)] attribute, so they're entrypoints and held on to by DCE`

`the function isn't inlined (because it only has plain arguments), which means that when considering the call, the call could do side effects or whatever (kinda? not on the gpu, but whatever), and it has no knowledge of the internals, so it pessimistically keeps it`

was her response.

rustc_codegen_spirv/src/linker/structurizer.rs
```diff
@@ -44,59 +310,149 @@ fn get_possible_merge_positions(blocks: &[Block], start: Word) -> Vec<usize> {
...
+        }
+
+        processed.push(front);
         next.extend(new_edges);
     }
 
     false
 }
 
+// Returns the idx of the branch that loops and the idx to the block that branches to the orignial block.
```
expenses, line 340:
```suggestion
// Returns the idx of the branch that loops and the idx to the block that branches to the original block.
```
:wink:

rustc_codegen_spirv/src/linker/structurizer.rs
```diff
@@ -1,17 +1,169 @@
...
-    insert_selection_merge_on_conditional_branch(sess, header, &mut func.blocks);
+pub struct LoopInfo {
+    merge_id: Word,
+    continue_id: Word,
+    header_id: Word,
+}
+pub struct ControlFlowInfo {
+    loops: Vec<LoopInfo>,
+    if_merge_ids: Vec<Word>,
+    switch_merge_ids: Vec<Word>,
```
khyperia, line 25: Should these be HashSets? Quick scan looks like there's no order-dependent stuff going on, and hashset would significantly improve the runtime of the `id_is_loops_merge` functions and friends.

khyperia (approved):
heck yeah! haven't actually gone in and reviewed all the algorithms, I trust you on that (and we can always come back and fix things). :+1: to the strategy of smaller pulls with some unimplemented functionality.

rustc_codegen_spirv/src/linker/structurizer.rs
bjorn3: It may be faster to do something like `IndexVec<Word, Option<ControlFlowKind>>` assuming that any `Word` can only be one of loop merge, loop continue, loop header, if merge, switch merge or none of the former.

PR #179, opened 2020-10-29
Add spirv-tools and spirv-tools-sys crates to wrap usage of spirv-tools

Jake-Shadle:
Replaces the use of spirv-tools' binaries (as, opt, and val) with the ability use the compiled C++ code instead, or in addition to the spirv binaries.

## Additions
* Adds [spirv-headers] and [spirv-tools] as git submodules (will add docs about retrieiving/updating them for contributors)
* Adds `spirv-tools-sys` which builds `spirv-tools`, including a small generator script that generates some includes based on [spirv-headers]
* Adds a `spirv-tools-sys/src/c/opt.cpp` which is just a thin (and incomplete) C wrapper around [Optimizer](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/include/spirv-tools/optimizer.hpp) because C++
* Adds a `spirv-tools` crate which provides a slighly nicer API on top of `spirv-tools-sys`

## Changes
* Replaces the usage of the `spirv-opt` binary in link.rs with `spirv_tools::opt::Optimizer`
* Replaces the usage of `spirv-val` for validating resulting binaries in link.rs with `spirv_tools::val::Validator`
* Replaces the usage of `spirv-as` in the linker tests with `spirv_tools::as::Assembler`
* All of the crates that use rustc_codegen_spirv now have a `use-installed-tools` and `use-compiled-tools` feature flags, where `use-compiled-tools` is the default feature for optimal user experience.

## Missing
* There is one change I had to do in our [spirv-tools](https://github.com/EmbarkStudios/SPIRV-Tools/commit/83bb6bacf26160e0b04c092255ac0bbb9ae42e48) fork due some incredible strangeness with libstdc++ streams, so right now the `Assembler` will choke on hexadecimal floats, but we only use the `Assembler` in tests and none of them currently use hexadecimal floats, so figured this would be ok for now, the issue is fixable but doing C++ code was making me sad so I punted until it became an actual issue.

Resolves: #31 

[spirv-headers]: https://github.com/KhronosGroup/SPIRV-Headers
[spirv-tools]: https://github.com/KhronosGroup/SPIRV-Tools


Jake-Shadle:
Sorry, there were a couple of issues with #117, this PR redoes it with fixed docs and (hopefully soon) fixed windows.

Jake-Shadle:
This fixes the docs deployment.

PR #184, opened 2020-10-30
Make all example windows 720p by default

### LINKED ISSUE #178: standardize example runner windows sizes to 720p
Currently, all the example runners have different sizes, with the ash runner making a pretty unwieldy full 1080p window. Should standardize them to 720p

nolanderc:
Resolves #178.

Regarding using a scaled resolution for 4K monitors: as far as I can tell, using `winit`'s `LogicalSize` respects monitor scaling, so it should be fine even for larger resolutions.

### DIFF
--- a/examples/runners/ash/src/main.rs
+++ b/examples/runners/ash/src/main.rs
@@ -564,7 +564,7 @@ fn main() {
     let options = Options::from_args();
 
     unsafe {
-        let (base, events_loop) = ExampleBase::new(1920, 1080, &options);
+        let (base, events_loop) = ExampleBase::new(1280, 720, &options);
         let renderpass_attachments = [vk::AttachmentDescription {
             format: base.surface_format.format,
             samples: vk::SampleCountFlags::TYPE_1,


examples/wgpu-example-runner/src/main.rs
```diff
@@ -126,8 +126,17 @@ async fn run(event_loop: EventLoop<()>, window: Window, swapchain_format: wgpu::
...
 
 fn main() {
+    const WIDTH: u32 = 1280;
+    const HEIGHT: u32 = 720;
+
     let event_loop = EventLoop::new();
-    let window = winit::window::Window::new(&event_loop).unwrap();
+    let window = winit::window::WindowBuilder::new()
+        .with_inner_size(winit::dpi::LogicalSize::new(
+            f64::from(WIDTH),
```
termhn, line 135: I think I would rather just put f64 literals here (`1280.0`) than defining separate u32 consts.

termhn (changes requested):
Thanks and welcome to the project!! <3

Just a heads up, when #109 gets merged, this will need to be rebased because the folder structure of the examples got changed!

examples/wgpu-example-runner/src/main.rs
nolanderc: Yup, I was mirroring `examples/example-runner-cpu`, but here we don't actually don't use the consts for anything other than initializing the window size.

termhn (changes requested):
Great! All that's left is a rebase on latest `main` :D

termhn (approved):

PR #187, opened 2020-10-30
Add initial platform support docs

### LINKED ISSUE #177: Documenting Platform Support
As discussed in the last meeting, we want to clarify what platforms, vendors, and versions we support, and what we're focusing on right on now. the outcome of this would be a page in the GPU dev guide that provides an overview the support and what that kind of support means for the platforms.

XAMPPRocky:
### [Rendered](https://github.com/EmbarkStudios/rust-gpu/blob/platform-support/docs/src/platform-support.md)

closes #177

### DIFF
--- a/docs/src/platform-support.md
+++ b/docs/src/platform-support.md
@@ -38,5 +38,5 @@ Currently we don't have specific generations of GPUs for support, as long they s
 [linux-intel]: https://www.intel.com/content/www/us/en/support/articles/000005520/graphics.html
 [macOS-intel]: https://www.intel.com/content/www/us/en/support/articles/000022440/graphics.html
 [windows-intel]: https://downloadcenter.intel.com/product/80939/Graphics
-[`vulkaninfo`]: https://vulkan.lunarg.com/doc/view/1.2.148.1/windows/vulkaninfo.html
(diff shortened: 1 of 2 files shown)


docs/src/platform-support.md
```diff
@@ -0,0 +1,42 @@
...
+- **Primary —** Built and tested on CI.
+- **Secondary —** Built but *not fully tested* on CI.
+- **Tertiary —** Present in the codebase but not built or tested.
+
+## Operating System
+
+| Operating System | Version | Support | Notes
+|-------------------|---------|---------|-------
+| Windows | 10+ | Primary | |
+| Linux | ??? | Primary | |
```
khyperia, line 13: maybe the ubuntu CI version we use here?

khyperia (approved):

docs/src/platform-support.md
repi: That's a good idea and could be a base

Jasper-Bekkers (approved):
We probably should just link to `https://vulkan.lunarg.com/doc/view/latest/windows/vulkaninfo.html` instead of a specific version of `vulkaninfo`. Otherwise this looks good.

XAMPPRocky:
@Jasper-Bekkers Good catch, I just got the URL from a search. Updated now.

repi (approved):

PR #188, opened 2020-10-30
Fix val error output

Jake-Shadle:
This fixes a bug pointed out by @VZout, I was accidentally linking to the wrong C function so we never got a diagnostic from validation, this fixes that bug, and adds a test using a shader that spirv-val produces an error message on so that I can refine the error diagnostics in the future (eg, using the cmd line tool is lossy compared to the native code).


spirv-tools/tests/val.rs
```diff
@@ -0,0 +1,44 @@
+const SPIRV_BIN: &[u8] = include_bytes!("wgpu_example_shader.spv");
+
+fn validate_compiled(_input: &[u8]) -> Option<Result<(), spirv_tools::Error>> {
+    #[cfg(feature = "use-compiled-tools")]
```
XAMPPRocky, line 4: It might be worth adding `cfg_if` at some stage, lets you write this in a much nicer fashion.

```rust
cfg_if::cfg_if! {
    if #[cfg(feature = "use-compiled-tools")] {
       Some(..)
    } else {
       None
    }
}
```

Jake-Shadle: Yah, I was planning on fleshing out testing for this eventually, that would definitely be a part of the push.

VZout (approved):

Issue #180, opened 2020-10-29
Compute shader support & basic example


repi:
We are focusing first now on fragment & graphics shaders, and the binding for that. But do want and need to add support for basic compute shaders soon also. This is tracking that.

charles-r-earp:
Features relevant to compute:

- [x]  LocalSize
- [ ] Buffer inputs / outputs
- [ ] Builtins like GlobalInvocationID / LocalInvocationID
- [ ] PushConstants
- [ ] Shared Workgroup Memory
- [ ] Memory Barriers
- [ ] Loops
- [ ] Math / Approx Ops
- [ ] Atomic Ops  

With #195, you can compile and execute a compute shader, but it can't do anything.

Cazadorro:
Are these the math operations you want to support those found here? https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_C.html#math-functions

I don't know what math functions SPIR-V actually supports underneath, but I have a private repo I abandoned that had approximations of a lot of ei, e1, erf erfc, erfcinv erfinv.  I just un-privated it so you [can see](https://github.com/Cazadorro/sfal), but I made it very early on when I was using rust, and I'm "newish" to rust development in general, so I probably didn't use best practices.  But what it does have is several approximations for several of the main "special functions" (except lgamma an tgamma), with sources to where I got the implementations from.  The origional purpose ironically was to demo possible approximations for functions missing from OpenCL for GLSL in rust, but found a way around using the special function I wanted to use.  

Another resource I found that was great was this:

https://developer.download.nvidia.com/cg/cos.html

or well the rest of the math functions it showed on the side, some of them show example approximations used, not sure if any of the math ops you can't get builtins for are there.

charles-r-earp:
Yes, I wasn't sure if those work in shaders generally or just kernels. OpenCL has __FAST_RELAXED_MATH__, I'm not sure if that can be specified some how. 

In particular fma, exp, ln, sqrt, etc.

Cazadorro:
I would complain to Khronos group about the fast math stuff, though I think for most usecases it shouldn't matter.  To better explain, I don't think there's a way to say *don't* optimize math in SPIR-V, and because of this, you get odd behavior like [this](https://stackoverflow.com/questions/60079088/problem-with-vulkan-floating-point-behavior).   Essentially Nvidia is using FAST_RELAXED_MATH behind the scenes after consuming your SPIR-V code, at least in some instances. 

[...]

PR #195, opened 2020-10-30
Compute Shaders + ExecutionMode's

charles-r-earp:
Adds a wgpu compute shader example, "wgpu-example-compute-runner" and "wgpu-example-compute-shader". At minimum to compile and execute a compute shader, ie GLCompute, the execution mode LocalSize must be set. So this PR additionally adds a mechanism to do that:

```
#[spirv(gl_compute(local_size_x=64, local_size_y=16, local_size_z=4))]
pub fn main_cs() {}
```
Emulating GLSL, dimensions not specified default to 1 (gl_compute only). 

The execution modes specified within a execution model, are collected into a Vec along with their arguments, and then this is sent as a SpirvAttribute::Entry, which is now special Entry struct. In entry.rs, the stub functions submit these execution modes. 

I also added the fn compute_shader_entry_stub as a duplicate of shader_entry_stub, which in the future may handle function parameters differently. 

I added the complete list of execution modes. There may be some additional logic for default values that I'm not familiar with, notably you can now specify OriginLowerLeft to vertex shaders, (OriginUpperLeft is the default). There isn't any error handling about which modes are valid per execution model, or specifying multiple conflicting modes. In general they are collected and submitted left to right. 

All the tests passed.


rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -241,5 +249,78 @@ impl<'tcx> CodegenCx<'tcx> {
...
 
         emit.entry_point(execution_model, fn_id, name, &[]);
+        
+        entry.execution_modes.iter()
+            .for_each(|(execution_mode, execution_mode_extra)| {
+                emit.execution_mode(fn_id, *execution_mode, execution_mode_extra);
+            });
+    }
+    
+    fn compute_shader_entry_stub(
```
khyperia, line 259: There's a decent amount of duplication between this and `shader_entry_stub` (and the existing duplication with kernel too), it might be nice to refactor things into something a little better structured - but probably in a later PR (I can do so), this is fine as-is as a temp solution.

repi:
wow thanks for the contribution! 

linking to #180

rustc_codegen_spirv/src/symbols.rs
```diff
@@ -279,15 +414,169 @@ pub fn parse_attrs(
...
+                                                    );
+                                                }
+                                                    
+                                            });
+                                    }
+                                    match e.execution_model {
+                                        Fragment => {
+                                            //FIXME: Is this the correct default?
+                                            let origin_mode = origin_mode.unwrap_or(OriginUpperLeft);
+                                            e.execution_modes.push((origin_mode, ExecutionModeExtra::new([])));
```
khyperia, line 547: This is a little convoluted code so I might be missing something, but it might be nice to only specify this if the user didn't do so themselves. (Considering we didn't even allow users to specify it before this PR, this is fine as-is for now, mostly just making a note that we should follow up on this later!)

rustc_codegen_spirv/src/symbols.rs
```diff
@@ -279,15 +414,169 @@ pub fn parse_attrs(
...
+                            }
+                            a
+                        })
+                };
+                if result.is_none() {
+                    cx.tcx
+                        .sess
+                        .span_err(name.span, "unknown argument to spirv attribute");
                 }
+                result
```
khyperia, line 579: This whole thing is super nested, it'd be really nice if we could split it into some smaller functions.

examples/wgpu-example-compute-runner/src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+    use rspirv::binary::Disassemble;
+    let mut loader = rspirv::dr::Loader::new();
+    rspirv::binary::parse_words(binary, &mut loader).unwrap();
+    loader.module()
+        .disassemble()
+}
+
+fn main() {
+    let spirv = include_bytes_align_as!(u32, env!("wgpu_example_compute_shader.spv"));
+    println!("{}\n", disassemble_spirv(bytemuck::cast_slice(spirv)));
```
khyperia, line 54: prooobably shouldn't include this, as all the other examples don't? but shrug, doesn't really matter either way.

examples/wgpu-example-compute-runner/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "wgpu-example-compute-runner"
+version = "0.1.0"
+authors = ["charles-r-earp <charles.r.earp@gmail.com>"]
```
khyperia, line 4: this feels weird to ask (@repi maybe could provide feedback here?), but should this be embark's email that's in every other crate, or yours?

khyperia (commented):
Oh my *goodness*, I was gearing up to implement this, and then you just come along and, do it??? Thank you so so so very much, this is incredible! I'm very happy taking this as-is right now (with the `-> bool` rustc nightly version fixed) and addressing the other feedback in a follow-up (either me or you, either is fine by me).

rustc_codegen_spirv/src/symbols.rs
charles-r-earp: This isn't a breaking change. The origin_mode variable is None, and replaced with OriginUpperLeft or OriginLowerLeft if specified. Before it was hardcoded in entry.rs for fragment shaders.

Kamayuq (approved):

khyperia:
I'm not sure why CI isn't running on this - perhaps the merge conflicts need to get resolved first?

charles-r-earp:
Yes, merge conflicts are fixed.

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -199,23 +313,23 @@ impl Symbols {
             .map(|&(a, b)| (a, SpirvAttribute::StorageClass(b)));
         let execution_models = EXECUTION_MODELS
             .iter()
-            .map(|&(a, b)| (a, SpirvAttribute::Entry(b)));
-        let custom = std::iter::once((
-            "really_unsafe_ignore_bitcasts",
-            SpirvAttribute::ReallyUnsafeIgnoreBitcasts,
-        ));
```
khyperia, line 206: why split this out into its own thing?

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -254,61 +407,70 @@ pub fn parse_attrs(
...
+                                "#[spirv(..)] attribute argument must be single identifier",
+                            );
+                            return None;
+                        }
+                    };
+                    cx.sym
+                        .attributes
+                        .get(&name.name)
+                        .map(|a| match a {
+                            SpirvAttribute::Entry(entry) => SpirvAttribute::Entry(
```
khyperia, line 460: I would prefer this gets handled in the same way `descriptor_set` and `binding` is handled, but whatever, that can be cleaned up later.

khyperia (approved):
thanks so much!

crates/rustc_codegen_spirv/src/symbols.rs
charles-r-earp: So you're saying add additional if else statements for each entry / execution_model? Wouldn't putting everything in the attributes HashMap scale better?

crates/rustc_codegen_spirv/src/symbols.rs
charles-r-earp: I figured it was commonly used. And this logic matches descriptor_set, binding. But it probably should be reverted to how it was, in attributes.

PR #201, opened 2020-11-02
Move tests from spirv-builder to their own crate

### LINKED ISSUE #170: Crate, Example, Documentation and Test Organisation
The amount of crates, examples documentation, and tests are expanding pretty rapidly, and some of the layout is already is mildly confusing. I think we should figure out some sort of structure for each.

### Crates
Currently we have three dev crates (five with #117) which are at the top level, the top level is already becoming pretty crowded with documentation, and configuration files.  I think we should have a `crates/` folder that houses the main development crates. 

### Documentation
Currently we have two directories (`assets` and `rfcs`) which are at the top level, but I think since these will be less frequently used I would propose that we should move `assets` into `docs` and move `rfcs` into `docs/src` and make it part of the rust GPU dev guide.  This has the advantage of making the RFCs viewable in the rust gpu guide directly.

### Tests
The integration tests for `rustc_codegen_spirv` are currently located inside `spirv-builder` in `spirv-builder/src/test`. This isn't where I would expect these kinds of tests to be. I propose that we create a new separate crate for holding integration/end-to-end tests (e.g. `test_suite`, `codegen_test`, etc).

### Examples
We have four different example crates at the moment, all on the same level in `examples`, however it's not really clear which examples are related, or use each other. I would like to propose adding another level to examples that covers the basic domain of example (e.g. a `wgpu` directory for wgpu examples), and remove the `example` prefix from the name (e.g. `example-runner` → `runner`).

----
I've included a `tree` output of what this proposed directory structure would look like if all of these changes implemented below.
```
.
├── CODE_OF_CONDUCT.md
├── CONTRIBUTING.md
├── Cargo.lock
├── Cargo.toml
├── LICENSE-APACHE
├── LICENSE-MIT
├── README.md
├── crates
│   ├── rustc_codegen_spirv
│   │   ├── Cargo.toml
│   │   └── src
│   ├── spirv-builder
│   │   ├── Cargo.toml
│   │   └── src
│   ├── spirv-std
│   │   ├── Cargo.toml
│   │   └── src
│   └── test-suite
│       ├── basic.rs
│       └── mod.rs
├── deny.toml
├── docs
│   ├── assets
│   │   └── sky.jpg
│   ├── book.toml
│   └── src
│       ├── SUMMARY.md
│       ├── attributes.md
│       ├── introduction.md
│       └── rfcs
├── examples
│   ├── basic
│   │   ├── cpu-runner
│   │   ├── runner
│   │   └── shader
│   └── wgpu
│       ├── runner
│       └── shader
├── rust-toolchain
├── setup.bat
└── setup.sh

20 directories, 21 files
```

XAMPPRocky:
Final part of and fixes #170


crates/testsuite/tests/basic.rs
```diff
@@ -1,4 +1,4 @@
-use super::{dis_fn, val};
+use testsuite::{dis_fn, val};
```
khyperia, line 1: Still would prefer this to be in src, not tests, especially considering now that the crate name is literally called testsuite, but oh well.

(Context for others, having it be in tests instead of src builds 1 crate per file. I'm not sure if they're run simultaneously, but if they are, that's a problem, because the global lock is only per-process.)

khyperia (approved):

crates/testsuite/tests/basic.rs
XAMPPRocky: One of the advantages when structured like this though, is that all of the testing apparatus code is compiled once and shared with all tests, so adding or changing a test doesn't cause the entire testsuite to need to recompile, where it would if all of the tests were in `src`.

khyperia: Just want to check, I can't actually find any documentation on this, is it guaranteed each file/crate is run serially?

XAMPPRocky: I'm pretty sure in this case cargo is locking across processes, but I've reverted those changes for now and I'll write a bigger thing about testing in #48 as I think two are related.

khyperia: Cargo's lock does not help here, as tests can still modify files of the project outside of cargo's lock. In an earlier iteration of this framework, we were hitting failures due to that, which is why GLOBAL_MUTEX exists.

PR #212, opened 2020-11-05
Add changes to compile rust-gpu on the latest nightly

XAMPPRocky:
This PR along with allows rustc to be able to pull in and compile rust-gpu as a codegen backend. There's also some changes that are needed on rustc side, for which I'll create a follow up PR that contains a way to checkout and patch rust-lang/rust.

### DIFF
--- a/crates/rustc_codegen_spirv/src/lib.rs
+++ b/crates/rustc_codegen_spirv/src/lib.rs
@@ -151,23 +151,19 @@ fn is_blocklisted_fn(symbol_name: &str) -> bool {
 fn target_options() -> Target {
     Target {
         llvm_target: "no-llvm".to_string(),
-        target_endian: "little".to_string(),
         pointer_width: 32,
-        target_c_int_width: "32".to_string(),
-        target_os: "unknown".to_string(),
-        target_env: String::new(),
-        target_vendor: "unknown".to_string(),
         data_layout: "e-m:e-p:32:32:32-i64:64-n8:16:32:64".to_string(),
         arch: "spirv".to_string(),
-        linker_flavor: LinkerFlavor::Ld,
         options: TargetOptions {
+            allows_weak_linkage: false,
+            crt_static_allows_dylibs: true,
             dll_prefix: "".to_string(),
             dll_suffix: ".spv".to_string(),
-            panic_strategy: PanicStrategy::Abort,
-            emit_debug_gdb_scripts: false,
-            allows_weak_linkage: false,
             dynamic_linking: true,
-            crt_static_allows_dylibs: true,
+            emit_debug_gdb_scripts: false,
+            linker_flavor: LinkerFlavor::Ld,
+            panic_strategy: PanicStrategy::Abort,
+            target_os: "unknown".to_string(),
             // TODO: Investigate if main_needs_argc_argv is useful (for building exes)
             main_needs_argc_argv: false,
             ..Default::default()
--- a/crates/rustc_codegen_spirv/src/builder/mod.rs
+++ b/crates/rustc_codegen_spirv/src/builder/mod.rs
@@ -245,19 +245,16 @@ impl<'a, 'tcx> DebugInfoBuilderMethods for Builder<'a, 'tcx> {
     fn dbg_var_addr(
         &mut self,
         _dbg_var: Self::DIVariable,
-        _scope_metadata: Self::DIScope,
+        _scope_metadata: Self::DILocation,
         _variable_alloca: Self::Value,
         _direct_offset: Size,
         // NB: each offset implies a deref (i.e. they're steps in a pointer chain).
         _indirect_offsets: &[Size],
-        _span: Span,
     ) {
         todo!()
     }
 
-    fn set_source_location(&mut self, _scope: Self::DIScope, _span: Span) {
-        todo!()
-    }
+    fn set_dbg_loc(&mut self, _: Self::DILocation) { todo!() }
 
     fn insert_reference_to_gdb_debug_scripts_section_global(&mut self) {
         todo!()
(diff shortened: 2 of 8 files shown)


MarijnS95:
@XAMPPRocky The next sources changed `Scalar` handling which I've patched to see if the latest rustc would resolve another issue, perhaps this is useful to you: https://github.com/EmbarkStudios/rust-gpu/compare/pub-changes...MarijnS95:nightly

bjorn3:
If you symlink it to `compiler/rustc_codegen_spirv`, it should be possible to build it using `codegen-backends = ["llvm", "spirv"]`. The bootstrap changes to compile cg_clif should work with any codegen backend that can be found at `compiler/rustc_codegen_$name`.

crates/rustc_codegen_spirv/src/lib.rs
```diff
@@ -206,7 +206,7 @@ impl MetadataLoader for SpirvMetadataLoader {
 }
 
 #[derive(Clone)]
-struct SpirvCodegenBackend;
+pub struct SpirvCodegenBackend;
```
khyperia, line 209: Why are these pub now?

khyperia (commented):
Still need to wait for `rustfmt` to build on a new nightly and whatnot, but looks good pending that.

XAMPPRocky:
@bjorn3 Thanks for that, I tried using that method, however when trying that it fails on building the compiler_artifacts with a seemingly unrelated error.

```
error[E0107]: wrong number of type arguments: expected 2, found 1
   --> /Users/src/rust/compiler/rustc_data_structures/src/graph/iterate/mod.rs:271:10
    |
271 |     ) -> ControlFlow<Self::BreakVal> {
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 2 type arguments

error[E0107]: wrong number of type arguments: expected 2, found 1
   --> /Users/src/rust/compiler/rustc_data_structures/src/graph/iterate/mod.rs:276:51
    |
276 |     fn node_settled(&mut self, _node: G::Node) -> ControlFlow<Self::BreakVal> {
    |                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 2 type arguments

error[E0107]: wrong number of type arguments: expected 2, found 1
   --> /Users/src/rust/compiler/rustc_data_structures/src/graph/iterate/mod.rs:299:10
    |
299 |     ) -> ControlFlow<Self::BreakVal> {
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 2 type arguments
```

bjorn3:
https://github.com/rust-lang/rust/blob/f58ffc93815f76576eb56df4bdeec2fe8f12b766/library/core/src/ops/control_flow.rs#L6

Huh, the second type argument has a default. Can you post the full build log? Maybe that helps?

XAMPPRocky:
[Build Log](https://paste.rs/OEc)

bjorn3:
> command did not execute successfully: "/Users/src/rust/build/x86_64-apple-darwin/stage0/bin/cargo" "build" "--target" "x86_64-apple-darwin" "-Zbinary-dep-depinfo" "-j" "16" "--release" "--manifest-path" "/Users/src/rust/compiler/**rustc_codegen_spirv**/Cargo.toml" "--message-format" "json-render-diagnostics"

Did you add a dependency on `rustc_data_structures` or other rustc crates to `Cargo.toml` of `rustc_codegen_spirv`? That is not necessary and will cause issues.

XAMPPRocky:
@bjorn3 I added it because not having it gives the following error.

```
   Compiling spirv-builder v0.1.0 (/Users/src/rust/compiler/rustc_codegen_spirv/crates/spirv-builder)
error[E0463]: can't find crate for `rustc_ast`
  --> crates/rustc_codegen_spirv/src/lib.rs:57:1
   |
57 | extern crate rustc_ast;
   | ^^^^^^^^^^^^^^^^^^^^^^^ can't find crate

For more information about this error, try `rustc --explain E0463`.
error: could not compile `rustc_codegen_spirv`
```

https://paste.rs/8DF

bjorn3:
Only `crates/rustc_codegen_spirv` needs to be symlinked I think. (likely requires a `Cargo.toml` change to fix the `spirv-builder` path I think) Even the way you do it right now that error shouldn't happen I think.

XAMPPRocky:
@bjorn3 Well we do actually need crates as `rustc_codegen_spirv` depends on `spire-tools`, however symlinking just `rustc_codegen_spirv` and changing the path to the dependency manually in spirv's Cargo.toml does cause it to start working, I'm running through compiletest right now and will report more when it finishes.

MarijnS95:
Good news, 2020-11-09 [includes `rustfmt` again](https://rust-lang.github.io/rustup-components-history/) :tada: 

Fixes for 2020-11-06 and 2020-11-09 pushed to https://github.com/EmbarkStudios/rust-gpu/compare/pub-changes...MarijnS95:nightly

XAMPPRocky:
@MarijnS95 Nice, I ended up implementing the changes before I saw your comment. 😄

MarijnS95:
@XAMPPRocky I figured as much, saw your rebase and had to act "quick" as I had already done these changes over the weekend, should have mentioned that before :grimacing: 

At least we both sorted the members in `TargetOptions` :tada:

XAMPPRocky:
> EDIT: Btw these properties were moved from Target to TargetOptions because the latter one has sensible defaults. Most of our target_ fields match up, do you think we should omit some?

I've updated it now so that any that matched were removed. The only one remaining I believe is `target_os` since I don't know if we want the target name to be `spirv-unknown-none` or `spirv-unknown-unknown`, the former feels more accurate but I don't it matters much.

MarijnS95:
> I've updated it now so that any that matched were removed. The only one remaining I believe is `target_os` 

Yeah I came to the same conclusion, looks good this way :ok_hand: 

> since I don't know if we want the target name to be `spirv-unknown-none` or `spirv-unknown-unknown`, the former feels more accurate but I don't think it matters much.

Hmm, that's a tough one. Only `wasm32` seems to be using `unknown` for the OS, and searching around Clang appears to use `spir64-unknown-unknown`?

bjorn3:
You could say that the OS spirv code is running on is the driver or firmware, so in that case `unknown` makes sense, as you aren't running it bare bones without any driver at all, nor is the spirv the firmware itself.

khyperia (approved):
:tada:

PR #214, opened 2020-11-05
It is no longer necessary to provide supported_target_features

bjorn3:
This hasn't been necessary since rust-lang/rust#77795.


khyperia (approved):
neat, thanks!

Issue #190, opened 2020-10-30
Tracking Issue for Android Support


XAMPPRocky:
As mentioned in #177 we eventually want to be able to support android platforms, this is a tracking issue to track that progress.

PR #130, opened 2020-10-24
feat: implemented swap chain recreation #103

jirislav:
As requested in #103

PR #215, opened 2020-11-05
Add Android support to wgpu example

MarijnS95:
Closes: #190

As expected cross-compiling rust-gpu for Android works just fine :tada:, tested on both 10 and 11. This PR introduces the necessary changes to `example-runner-wgpu` to make it deal with the (imo rather obnoxious) suspend-resume handling. But hey, at least you can switch between apps without it breaking! [Don't ask](https://github.com/rust-windowing/winit/blob/3a077ff21125e9b130ab18596e9a363b5d9dcb86/src/platform_impl/android/mod.rs#L218) for the "back" button to work, though :sweat_smile: 

The same winit changes can be applied to `example-runner-ash` after #130 lands.

In other good news (also expected) compiling `rustc_codegen_spirv` on a native aarch64 device (Android phone with chroot) and using it to compile the `sky-shader` works OOTB; C++ spirv-tools didn't even make it break a sweat :slightly_smiling_face: 
It is unfortunately impossible to do this from termux yet (for the portable devs) as there are [no nightlies available](https://rust-lang.github.io/rustup-components-history/aarch64-linux-android.html) for `aarch64-linux-android`; manually attempting to compile this is a can of worms I'd rather leave closed.

EDIT: Do we want to build-test Android on the CI as well? We can set up something [like android-nk-rs](https://github.com/rust-windowing/android-ndk-rs/blob/master/.github/workflows/rust.yml#L90-L96).


examples/runners/wgpu/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+    window::Window,
+};
+
+async fn run(event_loop: EventLoop<()>, window: Window, swapchain_format: wgpu::TextureFormat) {
+    let size = window.inner_size();
+    let instance = wgpu::Instance::new(wgpu::BackendBit::PRIMARY);
+
+    // Wait for Resumed event on Android; the surface is only needed early to
+    // find an adapter that can render to this surface.
+    let mut surface = if !cfg!(target_os = "android") {
```
khyperia, line 13: nit: `!cfg!` is a little gross, could we swap the true/false branches if this statement?

i.e. `if cfg!(..) { None } else { Some(..) }`

examples/runners/wgpu/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+        .as_ref()
+        .map(|surface| device.create_swap_chain(&surface, &sc_desc));
+
+    event_loop.run(move |event, _, control_flow| {
+        // Have the closure take ownership of the resources.
+        // `event_loop.run` never returns, therefore we must do this to ensure
+        // the resources are properly cleaned up.
+        let _ = (&instance, &adapter, &module, &pipeline_layout);
+
+        *control_flow = ControlFlow::Wait;
```
khyperia, line 93: I have no idea what practical differences are (just what the raw functionality does), is there a reason you changed this from Poll to Wait?

examples/runners/wgpu/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+            Event::WindowEvent {
+                event: WindowEvent::CloseRequested,
+                ..
+            } => *control_flow = ControlFlow::Exit,
+            Event::WindowEvent {
+                event:
+                    WindowEvent::KeyboardInput {
+                        input:
+                            KeyboardInput {
+                                virtual_keycode: Some(VirtualKeyCode::Escape),
```
khyperia, line 151: neat, thanks~

examples/runners/wgpu/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+    let window = winit::window::WindowBuilder::new()
+        .with_title("Rust GPU - wgpu")
+        .with_inner_size(winit::dpi::LogicalSize::new(1280.0, 720.0))
+        .build(&event_loop)
+        .unwrap();
+
+    #[cfg(not(target_arch = "wasm32"))]
+    {
+        #[cfg(not(target_os = "android"))]
+        wgpu_subscriber::initialize_default_subscriber(None);
```
khyperia, line 175: looks like `wgpu_subscriber` is not used on android, could we cfg it out for android in Cargo.toml? (it's already cfg'd out for wasm32)

khyperia (commented):
For others reviewing this, github's UI is a little annoying to see the actual changes due to the moved file. If you check out this branch on your local repo, though, this will show you the diff:

`git diff origin/main:examples/runners/wgpu/src/main.rs HEAD:examples/runners/wgpu/src/lib.rs`

---

this looks hecking fantastic! pseudo-approval from me, just want to give you a chance to respond to the nits before mergify does its thing.

examples/runners/wgpu/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+#[cfg_attr(target_os = "android", ndk_glue::main(backtrace = "full"))]
+pub fn main() {
+    let event_loop = EventLoop::new();
+    let window = winit::window::WindowBuilder::new()
+        .with_title("Rust GPU - wgpu")
+        .with_inner_size(winit::dpi::LogicalSize::new(1280.0, 720.0))
+        .build(&event_loop)
+        .unwrap();
+
+    #[cfg(not(target_arch = "wasm32"))]
```
XAMPPRocky, line 172: I would add `cfg_if` and replace these blocks with that. We're already using it in the `ash` example.

XAMPPRocky:
> EDIT: Do we want to build-test Android on the CI as well? We can set up something like android-nk-rs.

Definitely

MarijnS95:
> For others reviewing this, github's UI is a little annoying to see the actual changes due to the moved file. If you check out this branch on your local repo, though, this will show you the diff:
> 
> `git diff origin/main:examples/runners/wgpu/src/main.rs HEAD:examples/runners/wgpu/src/lib.rs`

That's git rename detection going bonkers. I recall it being pretty good at understanding that there was a move from `main.rs => lib.rs` (with similarity index 98% in the move commit 0655b4911b1c6a17e7188782822ba0723387e22e) and `main.rs` reintroduced as a new file, but that's not happening here (anymore?). You can however use copy detection (`-C` flag on `show`/`diff`) so that the output is shown as if `lib.rs` is a copy from `main.rs`.

> this looks hecking fantastic! pseudo-approval from me, just want to give you a chance to respond to the nits before mergify does its thing.

Thanks!

> > EDIT: Do we want to build-test Android on the CI as well?
> 
> Definitely

@XAMPPRocky I'll get it set up in the next push round :+1:

examples/runners/wgpu/src/lib.rs
MarijnS95: Yeah I noticed it looks weird this way. I'm a fan of keeping `else { None };` as the last block but that doesn't really work out here, unless using `cfg!(not(target_os = "android"))`. That's more characters as well so I'll flip them around :+1:

examples/runners/wgpu/src/lib.rs
MarijnS95: That's explained in the relevant commit making this change: 54752c289c7d1b3a48488db3278344fddffcab0f

> The image currently does not change and the OS will notify us when to
> redraw (ie. after window resizes). This is going to save power
> especially on mobile devices.

> As soon as interactive or animating visuals are introduced to this
> example redraws can be requested with `window.request_redraw()`.

examples/runners/wgpu/src/lib.rs
MarijnS95: This caused a crash on-device but I haven't looked at the problem yet. Would be nice to have some form of tracing when necessary I guess?

MarijnS95: The problem turns out to be wgpu attaching a log tracer which conflicts with the Android logger _if_ set up. That Android logger has intentionally been left out as we do not print any logs anywhere in this example, and we'd need to wait for https://github.com/rust-windowing/android-ndk-rs/pull/88 to land in order to not clutter `Cargo.toml` with some "unused" crate.

examples/runners/wgpu/src/lib.rs
MarijnS95: Oh that's really nice, I didn't know `cfg_if` existed, thanks for the suggestion!

.github/workflows/ci.yaml
```diff
@@ -64,6 +63,31 @@ jobs:
         shell: bash
         run: .github/workflows/test.sh ${{ runner.os }}
 
+      - if: runner.os == 'Linux'
```
XAMPPRocky, line 66: Is it worth splitting this out into a separate job so that in the CI UI it says ubuntu, macOS, windows, android?

MarijnS95: Yeah I wasn't sure about this. The build:

- Breaks on Mac because of the wrong NDK name;
- Breaks on Windows because `export` is not a thing there.

The former one should be fixed now, the latter one I don't immediately know how to resolve (without writing separate scripts for separate os'es). Does Windows understand `$HOME` as used elsewhere in this file, shouldn't that be `%UserProfile%`?

The main reason it' stuffed in `check` is to leverage existing setup and build artifacts: is that okay or is there a better solution in the pipeline with the changes discussed yesterday?

MarijnS95: Should be fixed now: we don't need any of these ugly download steps as the GH Actions image comes with the NDK preinstalled :tada: 

Still an issue with spaces in paths on Windows, which I hope to get to the bottom of here: https://github.com/rust-windowing/android-ndk-rs/pull/92

MarijnS95: Spacing issue should be solved now, cc https://github.com/rust-windowing/android-ndk-rs/pull/92#issuecomment-723455010.

khyperia (approved):
:tada:

PR #216, opened 2020-11-09
Add barrier proposal

Tobski:
This proposal is designed to primarily address #8, though it also interacts with a few other issues, notably #19 and PR #29.

Lots of discussion on #8 integrated into this proposal, notably including that it uses explicit load/store functions.

Main questions from my perspective are around whether we want to enable a more stable unsafe API first, and what people think of usability of the proposal.


XAMPPRocky:
Thank you for your PR! One small thing is that the proposal is under `tree/main/docs/src/rfcs` instead of `docs/src/rfcs`.

### [Rendered](https://github.com/Tobski/rust-gpu/blob/patch-1/tree/main/docs/src/rfcs/002-fearless-concurrency.md)

Tobski:
@XAMPPRocky whoops, didn't trim enough of the path when I added it 🤦 - will fix!

docs/src/rfcs/002-fearless-concurrency.md
```diff
@@ -0,0 +1,191 @@
...
+
+##### Drop
+
+The `Mutable` variants of the wrapper types implement a custom `drop()` to ensure correct synchronization.
+
+```rust
+fn drop(&mut self);
+```
+
+In addition to destruction when it goes out of scope, an `OpMemoryBarrier` instruction is emitted with a `Scope` equal to that of the scope of the wrapper type, and `Release | UniformMemory` memory semantics.
```
XAMPPRocky, line 67: As others mentioned on Discord, [you can't rely on `drop` being run for safety.](https://doc.rust-lang.org/reference/destructors.html#not-running-destructors) If you want to guarantee destruction and replicate Rust's ownership semantics I would use closures. With closures you can pass it temporarily or give it to the closure but require that the user returns it.

```rust
struct Buffer;

impl Buffer {
    // Version A (mutable ref)
    fn borrow_mut_for_device(&mut self, callback: impl FnOnce(&mut DeviceVisibleMutableBufferValue)) {
        let buffer_value = _; // create DeviceVisibleMutableBufferValue
        callback(&mut buffer_value);
        self.destroy(buffer_value);
    }

    // Version B (passing ownership)
    fn borrow_mut_for_device(&mut self, callback: impl FnOnce(DeviceVisibleMutableBufferValue) -> DeviceVisibleMutableBufferValue) {
        let buffer_value = _; // create DeviceVisibleMutableBufferValue
        let buffer_value = callback(buffer_value);
        self.destroy(buffer_value);
    }
}

fn main() {
    // Version A
    buffer.borrow_mut_for_device(|buffer_value| {
        let value = buffer_value.load();
        // Use value...
    })
    // Version B
    buffer.borrow_mut_for_device(|buffer_value| {
        let value = buffer_value.load();
        // Use value...
        buffer_value
    })
}
```

Tobski: Ok yea with fresh eyes I now see why it's considered "safe" to effectively leak objects - link was helpful too, thanks!

Closures probably are a decent option for at least more advanced use of buffers - though "simple" load/store might suffer here. Going to have a think about this one, but it's given me some good ideas at least.

Tobski: Right, I've taken this feedback on board and flipping run with it - I've updated the proposal to V2.

docs/src/rfcs/002-fearless-concurrency.md
```diff
@@ -0,0 +1,220 @@
...
+Bitonic sort example (unrolled, incomplete...):
+
+```rust
+fn bitonic_sort_workgroup(buffer: MutDeviceStorage<u32>) {
+    buffer.map_workgroup_joined<2>(|array| {
+        [
+            max(array[0],array[1]),
+            min(array[0],array[1]),
+        ]
+    });
```
XAMPPRocky, line 91: Can't this length be inferred by the compiler? I would assume the below would work. Also the syntax isn't `<2>` it's `::<2>`
```suggestion
    buffer.map_workgroup_joined(|array| {
        [
            max(array[0],array[1]),
            min(array[0],array[1]),
        ]
    });
```

docs/src/rfcs/002-fearless-concurrency.md
```diff
@@ -0,0 +1,220 @@
...
+    data: StorageBuffer<T>,
+};
+```
+Workgroup storage is wrapped with a single type:
+
+```rust
+struct MutWorkgroupStorage<T> {
+    data: Workgroup<T>,
+};
+```
```
XAMPPRocky, line 35: Nit: In rust the convention is put `Mut` at the end of the type name. So these would be `DeviceStorageMut` and `WorkgroupStorageMut` respectively.

Some examples:
- https://doc.rust-lang.org/stable/std/collections/btree_map/struct.IterMut.html
- https://doc.rust-lang.org/stable/std/ops/trait.DerefMut.html
- https://doc.rust-lang.org/stable/std/borrow/trait.BorrowMut.html

docs/src/rfcs/002-fearless-concurrency.md
Tobski: Hm yep you're right. I'll fix this.

docs/src/rfcs/002-fearless-concurrency.md
Tobski: Ah yes, I will update. Thanks!

khyperia:
Closing this due to inactivity, if someone would like to start pushing on this again, feel free to reopen.

PR #219, opened 2020-11-10
Make SpirvValue contain an enum for its value

khyperia:
This is a prerequisite for all sorts of work (panics, printf extension, builtin functions, nested constants, all sorts of fun stuff) - `SpirvValueKind` is eventually going to have a decent number of cases. However, just for a proof-of-concept, I've added one "extra" case: `ConstantPointer`. Not only does this work towards my goal of removing the horrible `zombie_even_in_user_code` hack, it significantly improves error messages, since we're able to report violations of using the constant pointer directly in the compiler, not the linker, and so we have span information.

Before:

```rust
  error: dynamic use of address of constant
    |
    = note: Stack:
            main_fs
            Unnamed function ID %343
```

After:

```rust
  error: Cannot use this pointer directly, it must be dereferenced first
     --> examples/shaders/sky-shader/src/lib.rs:184:5
      |
  184 |     f(&2);
      |     ^^^^^
```

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -24,7 +24,7 @@ macro_rules! simple_op {
             assert_ty_eq!(self, lhs.ty, rhs.ty);
             let result_type = lhs.ty;
             self.emit()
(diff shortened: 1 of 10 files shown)


khyperia:
Because this is such a huge change, it's going to merge conflict with pretty much any compiler change, so I'm going to just merge this now.

PR #220, opened 2020-11-10
Update glam

XAMPPRocky:


examples/shaders/sky-shader/src/lib.rs
```diff
@@ -163,12 +163,12 @@ pub fn fs(screen_pos: Vec2) -> Vec4 {
     let clip_pos = screen_pos.extend(1.0).extend(1.0);
     let world_pos = {
         let p = clip_to_world.mul_vec4(clip_pos);
-        p.truncate() / p.w()
+        Vec3::from(p.truncate() / p.w())
```
khyperia, line 166: silly thing you didn't change so it's not super relevant here, but I just read glam's 0.10 changelog, and `transform_point3` now divides by `w` as of `0.9.4`, so we could change this to be `let word_pos = clip_to_world.transform_point3(screen_pos.extend(1.0));`.

khyperia (approved):

examples/shaders/sky-shader/src/lib.rs
```diff
@@ -160,15 +160,11 @@ pub fn fs(screen_pos: Vec2) -> Vec4 {
...
-    let world_pos = {
-        let p = clip_to_world.mul_vec4(clip_pos);
-        p.truncate() / p.w()
-    };
+    let world_pos = clip_to_world.transform_point3(screen_pos.extend(1.0));
     let dir = (world_pos - eye_pos).normalize();
 
     // evaluate Preetham sky model
-    let color = sky(dir, sun_pos);
+    let color = sky(dir.into(), sun_pos);
```
MarijnS95, line 167: I don't seem to need this `.into()` here, `dir` is a `Vec3` which `sky` expects.

Completely unrelated note: I tried using the new `let eye_pos = const_vec3!([0.0, 0.0997, 0.2]);` macro here which works fine on the host but only in SPIR-V with `const VARIABLE: Vec3 = const_vec3!([...]);`:

```rust
     --> examples/shaders/sky-shader/src/lib.rs:154:19
      |
  154 |     let eye_pos = const_vec3!([0.0, 0.0997, 0.2]);
      |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      |
      = note: from: *{Function} [u8; 16]
      = note: to: *{Function} [f32; 3]
```

That's a deliberate error from `zombie_bitcast_ptr`: I don't fully grasp what's going on, is this something you want documented in an issue to pursue later on?

XAMPPRocky: File an issue, at the very least maybe the error message can be improved.

termhn:
Hm, I'm not sure I like the approach of publicly re-exporting glam from `spirv_std`, especially if we want the separation of concerns discussed in #223. I think this also makes it harder to integrate with `macaw` to add extensions on top of `glam` that way. I might go back to the old way in #222 but with the newer `glam` version

XAMPPRocky:
I don't have strong opinions, I re-exported from glam because when glam is updated having it as a `[crates-io.patch]` means that when the version is incompatible cargo just warn just warn that the patch is unused and continue trying to build with the version on crates.io, which I was encountering a lot. Having it as a direct dependency prevents that, and it allows people to use the in progress spirv fork of glam without finding a PR on GitHub. Once there's a published version on crates.io I think we can pull it back out.

PR #222, opened 2020-11-10
Add trunc and fract to MathExt

termhn:
Does what it says on the tin ;)

### DIFF
--- a/crates/spirv-std/src/math_ext.rs
+++ b/crates/spirv-std/src/math_ext.rs
@@ -11,6 +11,8 @@ pub trait MathExt {
     fn ceil(self) -> Self;
     fn exp(self) -> Self;
     fn saturate(self) -> Self;
+    fn trunc(self) -> Self;
+    fn fract(self) -> Self;
 
     fn signum(self) -> Self;
     fn copysign(self, sign: Self) -> Self;
@@ -69,6 +71,14 @@ impl MathExt for f32 {
 
     fn saturate(self) -> f32 {
         self.max(0.0).min(1.0)
+    }
+
+    fn trunc(self) -> f32 {
+        unsafe { core::intrinsics::truncf32(self) }
+    }
+
+    fn fract(self) -> f32 {
+        self - self.trunc()
     }
 
     fn signum(self) -> f32 {


khyperia (approved):
neat! this commit is included in the other PR, so github might get a little confused here, but yeah~

also looks like the commit hash in <https://github.com/EmbarkStudios/rust-gpu/blob/bfa0f50f9be875f6e059172cb29f86c4f09bdd81/Cargo.toml#L19> is absolutely borked - I think you're going to have to update that to whatever the `spirv` branch is for CI to pass (I'm guessing someone force-pushed the branch)

repi (approved):

Cargo.toml
```diff
@@ -16,4 +16,4 @@ members = [
 
 [patch.crates-io]
 # https://github.com/bitshifter/glam-rs/pull/85
-glam = { git = "https://github.com/EmbarkStudios/glam-rs", rev = "d55ff75e91aab01c4fd9808e3ca9c079dfa7315c" }
+glam = { git = "https://github.com/EmbarkStudios/glam-rs", branch = "spirv" }
```
XAMPPRocky, line 19: Probably want to rebase onto #220 since that will conflict with this change.

crates/spirv-std/src/lib.rs
```diff
@@ -39,8 +39,6 @@
 mod math_ext;
 pub use math_ext::MathExt;
 
-pub use glam;
```
khyperia, line 42: @XAMPPRocky explicitly added this in https://github.com/EmbarkStudios/rust-gpu/pull/220, so we should figure out if we want this or not (I said :+1: to her adding it, so I think we should keep it).

Also, there's some issues with overriding versions in cargo.toml, in particular people referencing crates through a git reference will not pick up the workspace override, so keeping the override in the crate instead of the workspace is good.

termhn:
Okay, so based on meeting convo, I reverted back to pub use glam for now, pending figuring out how we can upstream to glam. Will just be using patching in ark/macaw for now.

khyperia (approved):

khyperia:
@Mergifyio refresh

PR #224, opened 2020-11-11
Use branch over commit for glam.

XAMPPRocky:


khyperia (changes requested):
Because https://github.com/EmbarkStudios/rust-gpu/pull/220 required rust-gpu changes alongside the update to glam, that scenario can happen again in the future: so, we should not auto-update our branch, because that has the potential to break everyone working on rust-gpu in the time between the glam branch being updated, and the rust-gpu update being landed.

PR #228, opened 2020-11-11
Synthesize Session for tests, instead of Option<Session>

khyperia:
Also remove LinkerError and use rustc error reporting directly instead.

Generally various other random clean-up things too (e.g. lib.rs imports were annoying me, so fixed those up)


VZout (approved):
lgtm!

PR #229, opened 2020-11-11
Put shader modules at the top of the example runners

XAMPPRocky:
This moves the shader module that the runners use to the top of the file to make it easier to quickly change between different shaders.


examples/runners/wgpu-compute/src/main.rs
```diff
@@ -1,3 +1,8 @@
+
```
bjorn3, line 1:
```suggestion
```

khyperia (approved):

examples/runners/wgpu-compute/src/main.rs
XAMPPRocky, line 1:
```suggestion
```

PR #231, opened 2020-11-12
Shader hot-reloading in WGPU example

repi:
This is the the test a did 1-2 weeks ago on hot-reloading of shaders, haven't had time to continue on it but do think we should clean it up and get it in to the all the example runners

It does have a problem with spirv-builder where that somehow causes every iteration in cargo-watch to rebuild `core` and everything, it seems like.  I did test before to not have the spriv-builder build path at all and just load .spv directly from the target folder and that worked flawlessly when iterating, could potentially disable the spriv-builder path when one has the `hot-reload` feature enabled as it is not needed then

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -47,6 +47,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anymap"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


khyperia:
https://github.com/EmbarkStudios/rust-gpu/pull/423 does the same thing as this and isn't a draft, so closing this!

PR #235, opened 2020-11-13
Linker pass to match storage class / pointer type for access chains.

charles-r-earp:
When casting a pointer of a type to the pointer of a field using access chain, the result storage class will now match the storage class of the base, which is required with OpAccessChain.

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -1215,11 +1215,8 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
     }
 
     fn pointercast(&mut self, val: Self::Value, dest_ty: Self::Type) -> Self::Value {
-        let (val_pointee, val_storage_class) = match self.lookup_type(val.ty) {
-            SpirvType::Pointer {
-                pointee,
-                storage_class,
-            } => (pointee, storage_class),
+        let val_pointee = match self.lookup_type(val.ty) {
+            SpirvType::Pointer { pointee, .. } => pointee,
             other => self.fatal(&format!(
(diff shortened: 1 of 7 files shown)


khyperia (changes requested):
Unfortunately this is not correct. Other code depends on the result of `pointercast` actually being the type specified in `dest_ty`, and now they mismatch. While this may happen to pass tests now, this is not the solution, and will cause subtle, confusing issues down the line. The correct fix would be to generate the correct type that is eventually passed into `dest_ty` (generally an extremely difficult problem).

However, as far as I can tell, the code right now fails extremely confusingly via spirv-val if you pointercast between two address spaces. It'd be great to turn changing storage classes in pointercast into a compiler error, instead of letting it get to `spirv-val`.

charles-r-earp:
Agreed. It is a deeper problem. I looked into trying to fix it upstream in abs.rs, but this actually leads to more issues. Basically, if we have say a Buffer<Block>, the Block pointer gets declared first and has no way to know that it is within the Buffer ie has a storage class. And, the interior type may be reused for different storage classes, so its not unique. 
What I realized is that there is a separation, the entry function will see the Buffer<Block> and associate it with the specified storage class, but methods on it will not. 
Instead, it may be simpler to fix this prior to assembly, after functions have been inlined and reduced. Since, as far as I can tell, this issue tends to only fail validation. It should be fairly easy to locate all the access chains or other ops and lookup the storage class of the base.

khyperia:
Is it possible for you to join the discord? I don't fully understand what's happening here, but it seems like something that should probably go through the MCP process, this is a fairly large fundamental change to how address spaces are handled, which is an extremely difficult and complex topic that has far-reaching implications. If not, could you file an MCP about what the problem is, what the plan is to fix it, and possibly how that fits into the addressing model story (in addition to things like the bindings RFC)?

PR #240, opened 2020-11-13
wgpu-runner: Added clap to switch between shaders

VZout:
Added `clap` so you can choose what shader to run with. i.e: `cargo run --bin wgpu-example-runner -- --shader=Simplest`.
Should make adding examples easier and removes the need (in most cases) for adding a new runner.


termhn (approved):
LGTM!

PR #241, opened 2020-11-13
Merged wgpu runners

VZout:
Combines the `wgpu` and `wgpu-compute` runners making use of the new clap options.
This one might be a bit more controversial.

`cargo run --bin wgpu-example-runner -- --shader=Compute`


termhn (approved):
Nice! LGTM

termhn:
re-ran CI jobs because the tests seem to have failed due to network error unrelated to this PR

repi:
The other ash runner uses structopt instead of clap (or rather on top of clap) :)  but clap v3 should be able to replace structopt in general so fine with testing it here

examples/runners/wgpu/src/compute.rs
```diff
@@ -27,19 +25,19 @@ fn create_device_queue() -> (wgpu::Device, wgpu::Queue) {
...
-        return futures::executor::block_on(create_device_queue_async());
-    };
+        futures::executor::block_on(create_device_queue_async())
+    }
     #[cfg(target_arch = "wasm32")]
     {
-        return wasm_bindgen_futures::spawn_local(create_device_queue_async());
-    };
+        wasm_bindgen_futures::spawn_local(create_device_queue_async())
+    }
```
XAMPPRocky, line 33: This could be reduced to the following since you're using `cfg_if` already. For some reason I can't provide it as a suggestion on GitHub.

```rust
cfg_if::cfg_if! {
     if #[cfg(target_arch = "wasm32")] {	
        wasm_bindgen_futures::spawn_local(create_device_queue_async())
    } else {
        futures::executor::block_on(create_device_queue_async())
   }
}
```

examples/runners/wgpu/src/graphics.rs
```diff
@@ -0,0 +1,211 @@
...
+    let render_pipeline = device.create_render_pipeline(&wgpu::RenderPipelineDescriptor {
+        label: None,
+        layout: Some(&pipeline_layout),
+        vertex_stage: wgpu::ProgrammableStageDescriptor {
+            module: &module,
+            entry_point: "main_vs",
+        },
+        fragment_stage: Some(wgpu::ProgrammableStageDescriptor {
+            module: &module,
+            entry_point: "main_fs",
```
XAMPPRocky, line 65: I would define these entrypoint string literals as `const`s at the top of the file.

VZout: Good suggestion but not really part of this PR. This change should also be made to ash not just wgpu. Its green because I changed the filename.

examples/runners/wgpu/src/compute.rs
VZout: done

XAMPPRocky (approved):

XAMPPRocky:
With two approvals and since this entirely touch the examples, I think we have enough to merge.

DJMcNab:
@Mergifyio refresh

PR #245, opened 2020-11-16
Sky shader Improvements.

VZout:
* Sky shader now responds to resizing properly.
* Added a width, height and time constant to all shaders.

### DIFF
--- a/examples/runners/ash/src/main.rs
+++ b/examples/runners/ash/src/main.rs
@@ -650,7 +650,7 @@ fn main() {
             .stage_flags(vk::ShaderStageFlags::all())
             .build();
         let layout_create_info = vk::PipelineLayoutCreateInfo::builder()
-            .push_constant_ranges(&vec![push_constant_range])
+            .push_constant_ranges(&[push_constant_range])
             .build();
 
         let pipeline_layout = base
--- a/examples/runners/ash/src/main.rs
+++ b/examples/runners/ash/src/main.rs
@@ -32,6 +32,17 @@ macro_rules! offset_of {
         }
     }};
 }
+
+struct ShaderConstants {
+    width: u32,
+    height: u32,
+    time: f32,
+}
+
+unsafe fn any_as_u8_slice<T: Sized>(p: &T) -> &[u8] {
+    ::std::slice::from_raw_parts((p as *const T) as *const u8, ::std::mem::size_of::<T>())
+}
+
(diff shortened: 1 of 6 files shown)


VZout:
image
 Clippy noooo.

examples/runners/wgpu/src/lib.rs
```diff
@@ -11,6 +11,13 @@ pub enum RustGPUShader {
     Compute,
 }
 
+#[allow(dead_code)]
+pub struct ShaderConstants {
```
XAMPPRocky, line 15: Instead of redefining `ShaderConstants` couldn't the `wgpu` and `ash` runners pull in the shader as a crate dependency like `cpu` does? That would remove a quite a bit of duplicate code.

VZout: It could but in the future it would be super annoying. Any other shader would have to include the sky shader just to get that struct. (I intend to use it for other shaders as well, similar to shadertoy). Could create yet another package as a "shared code" thing but I also don't really like that sound of that if its just a single struct.

VZout: if we end up with more shared code that would be the way to go though adding a "shared code" crate.

examples/runners/ash/src/main.rs
```diff
@@ -32,6 +32,18 @@ macro_rules! offset_of {
...
+#[allow(dead_code)]
+struct ShaderConstants {
+    width: u32,
+    height: u32,
+    time: f32,
+}
+
+unsafe fn any_as_u8_slice<T: Sized>(p: &T) -> &[u8] {
+    ::std::slice::from_raw_parts((p as *const T) as *const u8, ::std::mem::size_of::<T>())
+}
```
khyperia, line 45: Not relevant for this PR, but we should start developing a best practices guide in the rust-gpu dev guide on how to handle stuff like this. Just letting people run around with this level of terrifying unsaftey is sure to end in horror. (stuff like `#[repr(C)]`, zero-copy serialization crates(?), etc.)

khyperia (approved):
I am incredibly surprised push constants just worked without compiler support. Neat?

XAMPPRocky (commented):
Looks good to me code wise, though I'm trusting you that the math code is what's its supposed to do. Probably also want to make `ShaderConstants` `#[repr(packed)]` so that those `any_as_*_slice` methods are safe, currently if there was padding, it could be uninitialised.

XAMPPRocky (approved):

PR #254, opened 2020-11-18
implement asm!

### LINKED ISSUE #230: Support `asm!`
This blocks a surprising number of issues, so filing an issue to explicitly track and mention in other issues as "this is blocked on that".

Supporting `asm!` currently blocks:

- texture support: https://github.com/EmbarkStudios/rust-gpu/issues/204
- additional shader intrinsics: https://github.com/EmbarkStudios/rust-gpu/issues/197
- barrier proposal (well, the implementation of the proposal, not the proposal itself): https://github.com/EmbarkStudios/rust-gpu/pull/216

There _are_ alternatives to implement the above without `asm!`, but considering so many things would be resolved really easily if `asm!` was supported instead of the alternative of significant custom compiler support (I imagine quite a few future things will need it too), it'd be super great if we could implement `asm!`.

Supporting `asm!` itself is currently blocked by https://github.com/rust-lang/rust/pull/78950, which once it's in (hopefully it gets in!), we'll have to wait for a nightly to drop with it.

khyperia:
fixes #230

it's pretty fragile right now and likely needs follow-up PRs for complex shenanigans, but, it probably works for all our use cases!

... also, the `declare.rs` changes are unrelated, but they snuck into this diff and I'm too lazy to split them into another PR (it's refactoring/cleanup made possible due to some changes a while ago that didn't notice it could be simplified)


VZout (approved):
lgtm!

PR #255, opened 2020-11-18
Moved some functions to a shared crate

VZout:
Added a "shared" crate with some functions that I want to use in other shaders as well. and allows cpu and gpu code to share `ShaderConstants`

### DIFF
--- a/examples/shaders/shared/Cargo.toml
+++ b/examples/shaders/shared/Cargo.toml
@@ -1,14 +1,10 @@
 [package]
-name = "sky-shader"
+name = "shared"
 version = "0.1.0"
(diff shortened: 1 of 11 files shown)


examples/shaders/shared/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "shared"
```
bjorn3, line 2: This will probably need to be renamed if you want to publish on crates.io in the future.

khyperia: This is in `examples`, which will not be published.

khyperia (approved):

Issue #197, opened 2020-11-01
Support additional shader intrinsics


Jasper-Bekkers:
Some of these are only available in specific shader stages and should error out in other stages (for example, calculating screenspace derivatives might be unavailable outside pixel shaders). Also some of these may need to add additional capablities (such as SubgropuBallotKHR).

## Generic math functions
 - Already documented in #81
 - Possibly: OpDot & OpOuterProduct

## Bit instructions
 - OpBitFieldInsert
 - OpBitFieldSExtract / OpBitFieldUExtract
 - OpBitReverse
 - OpBitCount

## Fragment shader functions

 - OpDPdx (done)
 - OpDPdy (done)
 - OpFwidth (done)
 - OpDPdxFine (done)
 - OpDPdyFine (done)
 - OpFwidthFine (done)
 - OpDPdxCoarse (done)
 - OpDPdyCoarse (done)
 - OpFwidthCoarse (done)
 - OpKill (we may also want to take a look at OpDemoteToHelperInvocationEXT instead which is probably a nicer alternative depending on platform support; https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/EXT/SPV_EXT_demote_to_helper_invocation.asciidoc and https://github.com/microsoft/DirectXShaderCompiler/issues/2823)

## Compute
 - OpControlBarrier
 - OpMemoryBarrier (see also https://github.com/EmbarkStudios/rust-gpu/pull/216)

### Subgroup instructions
 - Probably all of "3.37.21. Group and Subgroup Instructions" and "3.37.24. Non-Uniform Instructions", but if we want to start with a more limited set the following three are probably fine.
 - OpSubgroupBallotKHR
 - OpSubgroupFirstInvocationKHR
 - OpSubgroupReadInvocationKHR

khyperia:
> Possibly: OpDot & OpOuterProduct

I'm confused. I thought you had the opinion that we should not add vector operation support (like add) because the componentwise operations are good enough - how are these different?

Jasper-Bekkers:
> I'm confused. I thought you had the opinion that we should not add vector operation support (like add) because the componentwise operations are good enough - how are these different?

Yeah that's why I marked them as optional; I think there might be some hw accelerated OpDot's out there but I think they're rare (old desktop parts or mobile gpus).

khyperia:
[...]

PR #256, opened 2020-11-18
Implement ddx, ddy, fwidth

khyperia:
Does things relevant to https://github.com/EmbarkStudios/rust-gpu/issues/197 (uh, avoiding the github magic word that closes the issue, since this doesn't deal with all of it)

Now that `asm!` is in, we don't even have to touch the compiler to implement this! wow!

bikesheddables:

- Function names? I copied the names from the spir-v spec.
- What to do on CPU? Currently panics, because the alternative of cfg'ing them out makes the IDE experience horrible (since IDE uses host target)

Things that are technically bikesheddables but I feel strongly about:

- Where do we put this? I think we should keep it in spirv-std in a specific extension trait, because it actually is extremely spirv-specific, and that's what spirv-std is for.

### DIFF
--- a/crates/spirv-std/src/lib.rs
+++ b/crates/spirv-std/src/lib.rs
@@ -1,5 +1,6 @@
 #![no_std]
 #![feature(register_attr, repr_simd, core_intrinsics)]
+#![cfg_attr(target_arch = "spirv", feature(asm))]
 #![register_attr(spirv)]
 // Our standard Clippy lints that we use in Embark projects, we opt out of a few that are not appropriate for the specific crate (yet)
 #![warn(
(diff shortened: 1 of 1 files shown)


crates/spirv-std/src/lib.rs
```diff
@@ -97,6 +98,68 @@ pointer_addrspace!(incoming_ray_payload_khr, IncomingRayPayloadKHR, true);
...
+        }
+    };
+}
+
+// "must be a scalar or vector of floating-point type. The component width must be 32 bits."
+deriv_impl!(f32);
+// TODO: Fix rustc to support these
+// deriv_impl!(glam::Vec2);
+// deriv_impl!(glam::Vec3);
+// deriv_impl!(glam::Vec4);
```
eddyb, line 161: Just confirmed that these lines (uncommented) stop erroring with https://github.com/rust-lang/rust/pull/79171.

repi (approved):
Think it is a great fit to have this in `spirv-std`, or at least it is the best place for it now

PR #259, opened 2020-11-18
Add redraw gap print to wgpu example

DJMcNab:


Jasper-Bekkers:
I think that if we measure anything we should be measuring milliseconds per frame, not frames per second. For arguments, see http://www.mvps.org/directx/articles/fps_versus_frame_time.htm and https://iliaskapouranis.com/2018/11/04/why-fps-is-a-bad-performance-metric/

DJMcNab:
Yeah, I was thinking about doing something similar.

At the moment I don't know how to ask wgpu when we have finished drawing the frame, which is unfortunate. I know this needs a lot of reworking, but winit is all a bit of a black box to me. I'm going to experiment with this a little bit.

DJMcNab:
I'm not planning on updating this further - there was talk of using https://github.com/khyperia/Clam/blob/clam5/src/fps_counter.rs in Discord, but currently the perfomance metrics are not very important; the maths in that fps counter look more optimised for giving a better approximation when sampled each frame when frame times are variable, but I would expect the current naïve implementation to be sufficient for the stable frame times our current shaders should be achieving.

khyperia:
Thanks for implementing this and putting in all the effort to make it nice, but it seeems that this probably doesn't add much value. I'll close this for now, but we can always revisit this and reopen it if someone wants to.

PR #266, opened 2020-11-20
Use a demangled symbol name for OpName.

### LINKED ISSUE #102: Fix issue with spirv-cross colliding names with builtins
Found in https://github.com/EmbarkStudios/rust-gpu/pull/86#issuecomment-714378730

Investigated a bit, turns out spirv-cross uses the `OpName` of functions directly in its output (rude! against spec!), which then can of course collide with builtins and other items. Some options:

- Disable OpName generation when targeting spirv-cross
- Collect builtin names in spirv-cross, and rename things in the compiler when there's a collision
- Make sure OpNames are fully qualified and unique (presumably there might be collisions between two functions named the same in separate crates or something...). Either the `rustc-demangle` crate or `tcx.def_path_debug_str` could help here, as well as `-Zsymbol-mangling-version=v0` to include all generic parameters in the symbol.

eddyb:
Fixes #102, using @khyperia's suggestion of demangling `v0` symbols to get an absolute path (with all of the generic parameters). This required #265 so that `-Z symbol-mangling-version=v0` works at all.

If we want to avoid `spirv-builder` passing `-Z symbol-mangling-version=v0` explicitly, I could make a small change to `rustc` that would allow overriding the symbol mangling version from within the codegen backend (e.g. via overriding queries).

Also undid https://github.com/EmbarkStudios/rust-gpu/pull/86#issuecomment-714378730, but will require testing on macOS to confirm the plain `smoothstep` works now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1964,6 +1964,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "rustc-demangle"
+version = "0.1.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6e3bad0ee36814ca07d7968269dd4b7ec89ec2da10c4bb613928d3077083c232"
+
+[[package]]
 name = "rustc_codegen_spirv"
 version = "0.1.0"
 dependencies = [
@@ -1971,6 +1977,7 @@ dependencies = [
  "pipe",
  "pretty_assertions",
  "rspirv",
+ "rustc-demangle",
  "spirv-tools",
  "tar",
  "tempfile",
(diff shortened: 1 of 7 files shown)


XAMPPRocky:
I got this error on macOS 11.0.1

```

failures:

---- test::basic::asm stdout ----
thread 'test::basic::asm' panicked at 'No function with that name found', crates/spirv-builder/src/test/mod.rs:123:10

---- test::basic::asm_add_two_ints stdout ----
thread 'test::basic::asm_add_two_ints' panicked at 'No function with that name found', crates/spirv-builder/src/test/mod.rs:123:10

---- test::basic::add_two_ints stdout ----
thread 'test::basic::add_two_ints' panicked at 'No function with that name found', crates/spirv-builder/src/test/mod.rs:123:10


failures:
    test::basic::add_two_ints
    test::basic::asm
    test::basic::asm_add_two_ints

test result: FAILED. 24 passed; 3 failed; 1 ignored; 0 measured; 0 filtered out
```

eddyb:
Sorry about that, forgot to run `cargo test`, I assume everything works now? The `sky_shader` example was the main thing to check, for the record, since it's the one that was failing with a `fn smoothstep` at the top-level of a crate.

XAMPPRocky (approved):

khyperia (approved):

Issue #192, opened 2020-10-30
support proc-macros


dvc94ch:
```rust
#[unroll::unroll_for_loops]
fn unroll_loop(a: Vec3) -> Vec3 {
    let mut b = Vec3::default();
    for i in 0..3 {
        b[i] + 1.0;
    }
    b
}
```
```sh
cargo +nightly build                                       (base) 
   Compiling wgpu-example-runner v0.1.0 (/home/dvc/implicit/marching_rays/runner)
error: failed to run custom build command for `wgpu-example-runner v0.1.0 (/home/dvc/implicit/marching_rays/runner)`

Caused by:
  process didn't exit successfully: `/home/dvc/implicit/marching_rays/target/debug/build/wgpu-example-runner-fbbe94bc606d1aa7/build-script-build` (exit code: 1)
  --- stderr
     Compiling wgpu-example-shader v0.1.0 (/home/dvc/implicit/marching_rays/shader)
  error[E0463]: can't find crate for `unroll`
    --> shader/src/lib.rs:45:3
     |
  45 | #[unroll::unroll_for_loops]
     |   ^^^^^^ can't find crate
     |
     = note: extern location for unroll is of an unknown type: /home/dvc/implicit/marching_rays/target/release/deps/libunroll-75b6e6a2c714c829.so
     = help: file name should be lib*.rlib or *..spv

  error: aborting due to previous error

  For more information about this error, try `rustc --explain E0463`.
  error: could not compile `wgpu-example-shader`

  To learn more, run the command again with --verbose.
  Error: BuildFailed
```

nolanderc:
I'm going to look into this!

nolanderc:
I traced the issue down to some weird behavior in rustc. The gist of the issue is that when building with the spir-v backend we configure the compiler to generate `.spv` files:
https://github.com/EmbarkStudios/rust-gpu/blob/afe3e1f83fe68b35d57efe23322f896aae6aa2ae/crates/rustc_codegen_spirv/src/lib.rs#L163-L166

However, for some reason, rustc uses the same postfix and suffix (`*.spv`) when considering which libraries it can use for plugins proc-macros, instead of using the default for the system it's running on (`*.so` on Linux, like in the error above). I tried changing the lines `dll_suffix` and `dll_prefix` above to match my system (MacOS) but got a similar error:

```
[...]

PR #267, opened 2020-11-20
spirv-builder: pass through non-JSON stdout lines even after an error.

eddyb:
While investigating #192, I couldn't get `RUSTC_LOG` to work for the nested build, and it turns out that it was a combination of it outputting on stdout (which https://github.com/rust-lang/rust/pull/79238 fixes) and `spirv-builder` not showing any of the stdout on failure.

This addresses the latter, in case a similar problem to the former, arises in the future.


khyperia (approved):
neat, I actually repeatedly hit this and was like "but wait I thought I passed through invalid-json-stdout" but never bothered to put in the time to debug why it wasn't printing. thanks!

PR #269, opened 2020-11-22
Add compile error when both use-compiled-tools and use-installed-tools are enabled

### LINKED ISSUE #261: Error when both use-compiled-tools and use-installed-tools are enabled
Similar to EmbarkStudios/spirv-tools-rs#4 except for when both are enabled. Currently if you enable both you can get linker errors.

Hentropy:
Resolves #261 

The compile error is located in the build.rs for rustc-codegen-spirv so that the error happens early, rather than after compiling spirv-tools.  If having a build.rs is an issue, it could be moved to lib.rs.


XAMPPRocky (approved):
Thank you for your PR! This is looks good to me. I'd rather have a quicker build. Congrats on your first contribution! 🎉

khyperia:
This should have gone into spirv-tools, not rust-gpu

XAMPPRocky:
@khyperia The original issue (#261) was on `spirv-tools` and moved here because it's valid to have both features enabled in `spirv-tools` but not in `rustc_codegen_spirv`. 

Screenshot 2020-11-23 at 09 03 01

PR #275, opened 2020-11-26
Move lang_items into spirv-std

XAMPPRocky:
This PR moves the lang_items that shader crates need to define into `spirv-std` itself, so `spirv-std` behaves a little more like `std`, and reduces the amount of code you need to get started. It's included under a default feature so if for some reason you still need to define your own `lang_items` you still can.


khyperia (approved):
neat! last time I tried doing this there were absolutely garbage build issues, but if it somehow magically works now, :tada:

Issue #204, opened 2020-11-03
Texture support


khyperia:
Implement texture support in the compiler. Exposing all of this is a *lot*, and needs a decent amount of UX thought, so posting this issue so people can chime in with ideas in addition to tracking the implementation.

Here's an overview of how textures work in the SPIR-V spec:

- OpTypeImage: an opaque data blob that contains some form of image. The type contains:
  - the underlying sampled type
  - dimensions
  - if it's a depth image (as in depth testing), can be yes/no/unknown
  - "arrayed"
  - whether it's multisampled
  - whether it can be used without a sampler (I'm guessing capabilities are required here?), values are "unknown, with sampler, without sampler"
  - image format, can be unknown
  - if it's read, write, or readwrite.
- OpTypeSampler/OpConstantSampler: does *not* contain image data, merely information on how to access an image. Data is in the OpConstantSampler, not the type (the type has no parameters at all).
  - None/ClampToEdge/Clamp/Repeat/RepeatMirrored
  - if it's normalized
  - sampler filter mode: Nearest/Linear
- OpTypeSampledImage/OpSampledImage: combines an image and OpConstantSampler into one meta-object.
  - The result of OpSampledImage has absolutely wild restrictions: it can only be referenced in the same BB that it's declared in, can't be used in OpSelect/OpPhi, if the image is in a composite then all indices getting the image must be dynamically-uniform, etc.

Then, *lots* of image read/write instructions (not sure what Dref is used for, "depth-comparison reference value"):

and edit, to be clear, nearly all of these (except for ImageRead/ImageWrite) take the combined TypeSampledImage. There is no avoiding using the combined version. The combined type has insane restrictions that pretty much guarantees that it will not be exposed to users as a struct (e.g. must be used in the same basic block it is declared in), but we cannot get around using it, it is how texture reads work in SPIR-V.

- OpImageFetch - directly on OpTypeImage with sampled=1, not OpTypeSampledImage, integer indices
- OpImageRead - directly on OpTypeImage with sampled=0 or 2, integer indices
- OpImageWrite - directly on OpTypeImage with sampled=0 or 2, integer indices
- OpImageGather - fetches one color from four pixels I think? weirdly, takes OpTypeSampledImage, despite name. float indices
- OpImageDrefGather - fetches depth from four pixels I think? weirdly, takes OpTypeSampledImage, despite name. float indices
[...]

PR #276, opened 2020-11-26
Textures!

khyperia:
Partially does https://github.com/EmbarkStudios/rust-gpu/issues/204 (just the compiler parts and the minimal library part to make sure the compiler bits work - library bits can be done in a follow-up)

I have a local change to wgpu runner based off https://github.com/gfx-rs/wgpu-rs/blob/e59ea495a66ce9606c3a2bbfc2efe8c0c05d413c/examples/cube/main.rs that makes sure textures work, but, considering it changes the shader significantly (adds a texture/sampler parameter and hardcode-replaces the sky shader), I'm not sure what I should do with it.

### DIFF
--- a/crates/rustc_codegen_spirv/src/abi.rs
+++ b/crates/rustc_codegen_spirv/src/abi.rs
@@ -305,6 +305,13 @@ impl<'tcx> ConvSpirvType<'tcx> for TyAndLayout<'tcx> {
 }
 
 fn trans_type_impl<'tcx>(cx: &CodegenCx<'tcx>, ty: TyAndLayout<'tcx>, is_immediate: bool) -> Word {
+    if let TyKind::Adt(adt, _) = *ty.ty.kind() {
+        for attr in parse_attrs(cx, cx.tcx.get_attrs(adt.did)) {
+            if let Some(image) = trans_image(cx, ty, attr) {
+                return image;
+            }
+        }
+    }
     // Note: ty.layout is orthogonal to ty.ty, e.g. `ManuallyDrop<Result<isize, isize>>` has abi
     // `ScalarPair`.
     // There's a few layers that we go through here. First we inspect layout.abi, then if relevant, layout.fields, etc.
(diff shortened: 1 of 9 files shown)


crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -493,6 +529,136 @@ pub fn parse_attrs(
...
+    let args = match attr.meta_item_list() {
+        Some(args) => args,
+        None => {
+            cx.tcx
+                .sess
+                .span_err(attr.span(), "image attribute must have arguments");
+            return None;
+        }
+    };
+    if args.len() != 6 && args.len() != 7 {
```
VZout, line 542: I think we can set a lot of defaults values. Having to set depth to 0 for 90% of textures seems a bit annoying.

VZout (approved):

crates/rustc_codegen_spirv/src/symbols.rs
khyperia: I don't imagine this attribute being used out of `spirv-std` (hence me leaving it undocumented), so user convenience and defaulting and whatnot isn't necessary.

VZout:
@khyperia don't forget to add it to the guide :) (or poke one of the community team to write documentation for you 🙊)

crates/rustc_codegen_spirv/src/abi.rs
```diff
@@ -705,3 +712,49 @@ fn name_of_struct(ty: TyAndLayout<'_>) -> String {
...
+            multisampled,
+            sampled,
+            image_format,
+            access_qualifier,
+        } => {
+            // see SpirvType::sizeof
+            if ty.size != Size::from_bytes(4) {
+                cx.tcx.sess.err("#[spirv(image)] type must have size 4");
+                return None;
+            }
```
eddyb, line 735: Just leaving this as a note, my preference is encoding it as `&Image` where
```rust
extern {
    #[spirv(image)]
    type Image;
}
```

But I have no idea if this is workable, so we should leave it for future (if ever) experimentation.

<hr/>

With a hardcoded size, one fun way to stop `rustc` from letting you nest the layout in anything else (with any more data), is to make it a newtype of `[u8; (1 << 47) - 1]`. That's because you can't have a size of 2<sup>47</sup> (or larger), on a 64-bit target (and there's a similar limitation on 32-bit targets but it's closer to 2<sup>32</sup>).

crates/rustc_codegen_spirv/src/symbols.rs
VZout: Oh okay nvm then :)

crates/rustc_codegen_spirv/src/abi.rs
eddyb: The other way to think about this is we could represent them as small as `1`-byte structs, *but no less*, so that the offset bijectively maps to the field, and generally to avoid any ZST special-casing.

The only reason to use pointer-size is the analogy with WASM, which has e.g. "function pointers" (and more generally `externref`), that are abstract values which can be referred to by an `usize` index into a global "table", if need be.
(Is that a plausible concern/direction with SPIR-V? Could we actually have a global array of e.g. `Image`s?)

We can also avoid any `Abi` special-casing by using `[u8; 1]`, IIRC, since that *should* create `Abi::Aggregate`.
(Either way we should assert the `.abi` is what we expect, in case it *does* change in the future)

crates/rustc_codegen_spirv/src/symbols.rs
eddyb: Could we specify some of these with `const` generics instead of attributes?

eddyb (commented):
I guess I can't "approve" it since it was already merged, but LGTM!

VZout:
Why was it merged..? It should have waited for your review.

Issue #128, opened 2020-10-24
Suggestion: Don't name everything after spirv.


lachlansneff:
# Proposal
I'm suggesting that the rust-gpu project doesn't name all the attributes and crates after "spirv". There are many gpu languages, and restricting this project to just spirv so early one might not be a good idea. And if it expands later to other languages, but the attributes and crates are still called spirv, it could pretty easily get confusing.

Instead, name things `shader` or `gpu`. e.g. `#[shader(...)]` and `use std_shader::...`.

repi:
Agreed! Which is why we also overall called the project/repo "Rust GPU" and have https://shader.rs :) 
Do hope that the current `spirv-std` will evolve to a `std-shader` (or `std-gpu` or similar). 

There are some SPIRV-specific intrinsics and types also though on the lower layers, but now when we can do `#[cfg(target_arch = "spirv")]` that will be very helpful in our own crates as well as others crates to compile in SPIRV-specifics but building general crates that can be compile for multiple targets, including CPU.

XAMPPRocky:
We discussed at the meeting, and the consensus was that we are in favour of changing the name, but want to wait until project is further along before renaming anything. So I'm going to close this for now.

PR #277, opened 2020-11-26
[RFC] Input / output proposal

Jasper-Bekkers:


Jasper-Bekkers:
[Rendered](https://github.com/EmbarkStudios/rust-gpu/blob/stage-interpolants/docs/src/rfcs/002-shader-stage-interpolants.md)

docs/src/rfcs/002-shader-stage-interpolants.md
```diff
@@ -0,0 +1,188 @@
...
+    VertexOutput {
+        position: world_to_clip.mul(input.position),
+        normal: input.normal,
+        color: input.color,
+    }
+}
+
+struct PixelOutput {
+    // albedo and material id packed into same RGBA8 render target
+    #[render_target(0)]
```
khyperia, line 162: Should (all) these attributes be `#[spirv(render_target = 0)]` to be consistent with [the rest of the attributes](https://embarkstudios.github.io/rust-gpu/book/attributes.html)?

docs/src/rfcs/002-shader-stage-interpolants.md
```diff
@@ -0,0 +1,188 @@
...
+    material_idx: u8,
+
+    // normal and exponent packed into RGBA8 render target
+    #[render_target(1)]
+    normal: Vec3,
+    #[render_target(1)]
+    specular_exponent: f32
+}
+
+fn ps_main(#[stage_in] input: &VertexOutput, shading: &ShadingInputs, indirect_lighting: &IndirectLighting) -> PixelOutput {
```
khyperia, line 174: Should these be by-value instead of by-ref? Having them be by-ref significantly complicates the implementation and generates suboptimal SPIR-V.

docs/src/rfcs/002-shader-stage-interpolants.md
```diff
@@ -0,0 +1,188 @@
...
+## Proposal
+
+```rust
+// these `attribute`'s would match the input assembler setup on the CPU
+struct VertexInput {
+    #[attribute(0)]
+    position: Vec3,
+    #[attribute(1)]
+    normal: Vec3,
+    #[attribute(2)]
```
khyperia, line 137: What behavior do you imagine when attributes are missing? Error, implicit default value, something else?

docs/src/rfcs/002-shader-stage-interpolants.md
```diff
@@ -0,0 +1,188 @@
...
+    color: Vec3,
+}
+
+// naming this is hard, since vertex shader won't really know what shader
+// stages will run after it until the pipeline state object gets created
+// in vulkan. a tesselation shader may run after this for example,
+// and that decision won't be known by our compiler.
+struct VertexOutput {
+    #[position]
```
khyperia, line 146: Just checking, does this replace the semantics for [builtins](https://embarkstudios.github.io/rust-gpu/book/attributes.html#builtins), and all of those builtins will be supported, in both input+outputs?

docs/src/rfcs/002-shader-stage-interpolants.md
```diff
@@ -0,0 +1,188 @@
...
+}
+
+struct PixelOutput {
+    // albedo and material id packed into same RGBA8 render target
+    #[render_target(0)]
+    albedo: Vec3,
+    #[render_target(0)]
+    material_idx: u8,
+
+    // normal and exponent packed into RGBA8 render target
```
khyperia, line 167: Does SPIR-V support doing this packing like this built-in, or is this something we need to implement ourselves? If ourselves, what's the justification? Seems like a really obscure feature.

docs/src/rfcs/002-shader-stage-interpolants.md
```diff
@@ -0,0 +1,188 @@
...
+// in vulkan. a tesselation shader may run after this for example,
+// and that decision won't be known by our compiler.
+struct VertexOutput {
+    #[position]
+    position: Vec3,
+    normal: Vec3,
+    color: Vec3,
+}
+
+fn vs_main(#[stage_in] input: &VertexInput) -> VertexOutput {
```
khyperia, line 152: Is it valid to not have a `#[spirv(stage_in)]` attribute? No return value?

khyperia (commented):
Thanks for doing this, this looks really awesome!

docs/src/rfcs/002-shader-stage-interpolants.md
Jasper-Bekkers: I thought we had consensus that we shouldn't do that #128 and I was under the impression that most of those `#[spirv]` attributes were either temporary or to get at specific implementation specific things from spirv. Not for higher level features.

docs/src/rfcs/002-shader-stage-interpolants.md
Jasper-Bekkers: I think by-ref here is an omission on my part, I think by-value doesn't make this code any more or less usable from a user's point of view, so by all means if it makes life easier on our end by-value it is.

docs/src/rfcs/002-shader-stage-interpolants.md
Jasper-Bekkers: Only a relatively small amount of those builtins actually make sense in this context - for example one can't output a `VertexId` and some of those are shader-stage specific like `TessLevelOuter` or `LocalInvocationId` so I would suggest we curate a list of attributes that we allow as inputs and outputs per shader stage.

docs/src/rfcs/002-shader-stage-interpolants.md
Jasper-Bekkers: Vec4 -> RGBA8 already happens automatically of course. And it's relatively common to be (bit) packing data into render-targets (example: gbuffer / vbuffer laydown). However, judging by the sketchy support (shader-playground)[http://shader-playground.timjones.io/2a474cd68f38f6891cadeebe41f394b3] we probably should just only support f32 (and vectors thereof) types in fragment shader outputs.

docs/src/rfcs/002-shader-stage-interpolants.md
khyperia: The conclusion of 128 is that we're keeping `spirv` for now, and perhaps we rename it to `#[shader(render_target = 0)]` in the future. Besides, there's significant technical difficulties in supporting `#[render_target(0)]`, and is the reason `#[spirv(...)]` was invented in the first place.

docs/src/rfcs/002-shader-stage-interpolants.md
Jasper-Bekkers: We should probably require both for now, I think there might be some obscure use-cases for vertex shaders so we can try to list them here and see.

- Procedural full-screen quad will require `VertexId`, and so will need to declare that in `#[stage_in]` through something like a `#[vertex_id]` attribute. This would imply that at least `#[attribute(N)]` would be optional.
 - It might be wanted occasionally would want to store out vertex data to a buffer (think debugging tools doing mesh visualization) - in this case you probably wouldn't want to run a pixel shader directly after (or maybe you would). However, this would probably require stream-out support (which we don't have, and probably won't have for a long time since it's quite an obscure feature).
 - Another (much more likely) case would be shadow-map rendering. In this case you don't bind a pixel shader, but still need to output position from the vertex shader.

I think it's safe to require them both, unless somebody has another valid use-case that I didn't think of?

docs/src/rfcs/002-shader-stage-interpolants.md
Jasper-Bekkers: I think it's most likely that all of the members of a `#[stage_in]` struct will have to have some kind of decoration - either `#[position]`, `#[vertex_id]`, etc or `#[attribute(..)]`. However, I don't think it's required for `#[stage_in]` to have to have an `#[attribute(..)]` (see below).

docs/src/rfcs/002-shader-stage-interpolants.md
khyperia: I'm confused, the shader-playground link doesn't have the same packing thing here, and so I'm not sure if you're saying that SPIR-V supports doing this packing built-in, or if we have to implement it ourselves. To be clear, I'm talking about having >1 field mapping to the same render target, and concatenating those fields into a single output.

docs/src/rfcs/002-shader-stage-interpolants.md
XAMPPRocky: Yeah, I agree even if it's not called `spirv` there should be at least one wrapper attribute around the inner attribute. It would get confusing as a user as to who is defined the attribute.

Jasper-Bekkers: Ah right, I was thinking of these along the same lines as the regular rustc [built-in attributes](https://doc.rust-lang.org/reference/attributes.html#built-in-attributes-index). However, if we feel putting these in their own namespace is the best way forward then I think we should do that. I would suggest putting them in another namespace then the `spriv` namespace though, mostly to keep them separate and avoid potential name clashes, but also to keep the new attribute clean and only add things there after some deliberation (instead of straight up forwarding spirv_headers symbols) and so we can avoid naming things after spirv-isms such as `gl_compute` or the confusion between workgroup and subgroup (two separate but very similar concepts in spirv).

docs/src/rfcs/002-shader-stage-interpolants.md
Jasper-Bekkers: > To be clear, I'm talking about having >1 field mapping to the same render target, and concatenating those fields into a single output.

Ah - I thought you meant the part about packing stuff together with u8 types and such. From the looks of it, it looks like the `Component` decoration isn't valid for render-target outputs and so all packing should be done manually.

In an earlier draft of this I actually had the pixel shader return a fixed-length array of (float|vec) from the pixel shader, something along these lines, which I feel is quite a bit simpler, but might run into trouble when outputting to render-targets with different components, or when outputting depth for example. 

```rust
fn ps_main(#[stage_in] input: &VertexOutput, shading: &ShadingInputs, indirect_lighting: &IndirectLighting) -> [Vec4; 2] {
    let mut T = brdf(&shading);
    T += input.color;
    T += gi(&indirect_lighting);

    [
        pack_gbuffer0(T, material_idx),
        pack_gbuffer1(normal, specular_exponent),
    ]
}
```

I think maybe the best option could be to have the ps output a struct where each member represents one render target. What do you think?

docs/src/rfcs/002-shader-stage-interpolants.md
charles-r-earp: Potentially #[spirv(attribute="position")] or #[spirv(attribute=0)] instead?

Jasper-Bekkers: The reason suggested much sorter syntax for most of these things is because they are quite common in shader cod and pop up all the time.

@charles-r-earp the word `attribute` here refers to a specific kind of attribute which is modeled in Vulkan by something known as the [vertex input attributes](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkVertexInputAttributeDescription.html), adding position in there would complicate things I feel because that's one of the outputs of a vertex shader.

Jasper-Bekkers: This may be an indicator that the word `attribute` is badly chosen for this case since it's already such an overloaded term, maybe `vertex_attribute` would be better, what do you think?

khyperia: > The reason suggested much sorter syntax for most of these things is because they are quite common in shader cod and pop up all the time.

We've already decided to go with `#[spirv(...)]` in previous discussions, bikeshedding over this isn't useful. If we want to change that decision, let's open a separate MCP with drawbacks/advantages/alternatives and discuss it there.

> However, I don't think it's required for `#[stage_in]` to have to have an `#[attribute(..)]` (see below).

I don't quite understand this comment, could you explain this more? What do you mean by `#[stage_in]` having an `#[attribute(..)]`?

docs/src/rfcs/002-shader-stage-interpolants.md
khyperia: > I would suggest putting them in another namespace then the `spriv` namespace though, mostly to keep them separate and avoid potential name clashes

I strongly disagree, the conclusion of #128 is that we're keeping things in `#[spirv(..)]` _for now_. If you believe the current system is unclean and has issues like potential name clashes, then let's address those in another MCP, instead of patching on half-fixes like new namespaces.

> so we can avoid naming things after spirv-isms such as `gl_compute`

I am fully aware of the issues with this and have asked feedback on names, and multiple times the conclusion has been that the current system is fine. If we want to change those decisions, let's open another thread.

khyperia:
Considering this has been stuck at "yeah I'll do this in a couple months" for... half a year, I don't think this is ever going to get worked on, and we're going to have to live with https://github.com/EmbarkStudios/rust-gpu/issues/416

PR #280, opened 2020-11-27
Refactor and document storage classes

XAMPPRocky:
This PR changes how storage classes where defined to their own module, refactors the macro to allow you to define multiple storage classes in a single macro invocation and more importantly allows you to add documentation to the storage classes. With the refactor I also changed the DSL you use to define storage so that's a more descriptive, and hopefully more intuitive, for example here is now what the definition of the `Input` and `Output` storage classes look like.

```rust
storage_class! {
    /// Input from pipeline.
    ///
    /// Visible across all functions in the current invocation. Variables
    /// declared with this storage class are read-only, and must not
    /// have initializers.
    #[spirv(input)] storage_class Input;

    /// Output to pipeline.
    ///
    /// Visible across all functions in the current invocation.
    #[spirv(output)] writeable storage_class Output;
}
```

### DIFF
--- a/crates/spirv-std/src/storage_class.rs
+++ b/crates/spirv-std/src/storage_class.rs
@@ -1,0 +1,203 @@
+//! # Storage Classes
+//!
+//! Class of storage for declared variables. These types act as pointers to
+//! values either contained in the GPU's memory. For example; `Input<f32>` points to a
+//! `f32` that was provided as input from the pipeline, and `Private<f32>`
+//! points to a `f32` in the GPU's global memory. Intermediate values do not
+//! form a storage class, and unless stated otherwise, storage class-based
+//! restrictions are not restrictions on intermediate objects and their types.
+
+macro_rules! storage_class {
+    ($(#[$($meta:meta)+])* storage_class $name:ident ; $($tt:tt)*) => {
+        $(#[$($meta)+])*
+        #[allow(unused_attributes)]
+        pub struct $name<'value, T> {
+            value: &'value mut T,
+        }
+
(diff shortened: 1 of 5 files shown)


crates/spirv-std/src/storage_class.rs
```diff
@@ -0,0 +1,193 @@
+//! # Storage Classes
+//!
+//! Class of storage for declared variables. Intermediate values do not form a
+//! storage class, and unless stated otherwise, storage class-based restrictions
+//! are not restrictions on intermediate objects and their types.
```
khyperia, line 5: Pointing out that they're aliases for pointers is important I think. I've seen some confusion over this in the rust-gpu discord, I don't think it's very clear right now that "`Input<T>`" means "a *pointer* to T, where that pointer is of style "input"" - and considering storage classes are literally "pointer flavors", seems important to document.

crates/spirv-std/src/storage_class.rs
```diff
@@ -0,0 +1,203 @@
+//! # Storage Classes
+//!
+//! Class of storage for declared variables. These types act as pointers to
+//! values either contained in the GPU's. For example; `Input<f32>` points to a
```
XAMPPRocky, line 4:
```suggestion
//! values either contained in the GPU's memory. For example; `Input<f32>` points to a
```

khyperia (approved):

XAMPPRocky:
I've also now also added a conivence function for "load and store" operations. For example multiplying and storing the result.

```rust
// Before
input.store(input.load() * 5);
// Now
input.then(|i| i * 5);
```

PR #281, opened 2020-11-27
Use new toolchain file

DJMcNab:
I'm not quite sure where is best to document that we require nightly rust, but in that list of instructions is definitely not it.

### DIFF
--- a/docs/src/introduction.md
+++ b/docs/src/introduction.md
@@ -11,22 +11,6 @@ how to use and develop on Rust-GPU.
     git clone --recurse-submodules https://github.com/EmbarkStudios/rust-gpu
     ```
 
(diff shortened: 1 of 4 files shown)


DJMcNab:
Realistically, blocked on https://github.com/rust-lang/rustup/issues/2579.

XAMPPRocky:
I don't see how that blocks this? I think removing the setup scripts and automatically installing components is still valuable even if it installs an extra component on CI, rustup is not where we're spending most of our time in CI anyway.

DJMcNab:
Yeah, it also seems like the GitHub actions rustup was not up to date, and I don't have time to work out how to work around that at the moment

Also I did find a workaround in that thread, so we should use that

I might reopen this in the next few days, but it might be worth whoever does the next rust update just doing it anyway

Issue #249, opened 2020-11-18
API Design: StorageBuffers


charles-r-earp:
# Summary 
This proposal attempts to address #232, #180, and #8. StorageBuffers, aka Buffer Blocks, are the primary inputs / outputs to compute shaders, though they can also be used in other stages. Iterators are one of Rust's highlights, and I propose emulating them in gpu code. 

# Example: Scaled Add aka Saxpy
```
// Rust Cpu
fn scaled_add(x: &[f32], mut y: &mut [f32], alpha: f32) {
    for (x, mut y) in x.iter().copied().zip(y) {
        *y += alpha * x;
    }
}

// OpenCL 
__kernel void scaled_add(__global const float* x, __global float* y, float alpha, uint n) {
    uint gid = get_global_id(0);
    if gid < n {
        y[gid] += alpha * x[gid];
    }
}

// GLSL 
#version 450

layout(set=0, binding=0) buffer Input {
    float x[];    
}

layout(set=0, binding=1) buffer Output {
    float y[];    
}

layout(push_constant) uniform PushConsts {
    float alpha;
    uint n;
}

void main() {
    uint gid = gl_GlobalInvocationID.x;
    if gid < n {
        y[gid] += alpha * x[gid];
    }
}
```

Saxpy is trivially parallel, that is, it can be separated into n independent operations. There are plenty of other similar kinds of operations common in CUDA / OpenCL code, which do not require any synchronization / barriers. In fact, this is probably the most common case. 

## Possible Naive Implementation in rust-gpu

```
#[allow(unused_attributes)]
#[spirv(gl_compute)]
pub fn scaled_add(x: Buffer<[f32]>, mut y: BufferMut<[f32]>, alpha: f32, n: u32) {
    let gid = spirv_std::global_x();
    if gid < n {
        unsafe {
            *y.get_mut_unchecked(gid) = x.get_unchecked(gid);
        }    
    } 
} 
```

This is the most straightforward translation of the above to rust gpu code. If we neglect concerns about aliasing with other shaders, then the only potential failure mode would be that the user provided n is outside the bounds of either x or y. The programmer is responsible, by using unsafe, to ensure no aliasing within the shader. 

The rayon crate allows for easy parallel processing on cpu's, with its ParallelIterator trait. It looks like this:

```
// rayon
fn scaled_add(x: &[f32], mut y: &mut [f32], alpha: f32) {
    use rayon::iter::ParallelIterator;
[...]

Issue #232, opened 2020-11-12
Support "interface blocks"


termhn:
This is the primary missing functionality needed for Ark shaders that I've encountered so far (other than textures) 

https://www.khronos.org/opengl/wiki/Interface_Block_(GLSL)

khyperia:
Fixed by https://github.com/EmbarkStudios/rust-gpu/pull/295

PR #282, opened 2020-11-27
[RFC] Add GlobalBuffer proposal

charles-r-earp:
This proposal stems from #249, and attempts to address #232. It does not intend to conflict with #8 but barriers may be inherently necessary to allow mutable access. It may be possible to enable specific access paradigms for trivial cases where no synchronization is necessary, but that is intended to be a potential next step. 

[Rendered](https://github.com/charles-r-earp/rust-gpu/blob/global-buffer-rfc/docs/src/rfcs/00X-global-buffer.md)


docs/src/rfcs/00X-global-buffer.md
```diff
@@ -0,0 +1,324 @@
...
+
+Global and GlobalMut are essentially either &(mut)T or &(mut)\[T\]. They are only to be provided to user code when they are not aliased. Safe abstractions should ensure that this is the case, and then yield these objects. 
+
+Due to the nature of GlobalMut, it is critical that the compiler prevents GlobalMut from being passed in as a parameter, since it marks a safe to access reference or slice. Global, since it is immutable, is not problematic, but might as well be banned as well, as it isn't meant to be used this way, and will not be portable. Further, the spirv attribute, at least for "storage_buffer" and "cross_workgroup", etc, should be limited only to within the crate.
```
Jasper-Bekkers, line 96: I'm not sure I follow exactly why `GlobalMut` shouldn't be passed to functions, would you mind elaborating this a bit?

docs/src/rfcs/00X-global-buffer.md
```diff
@@ -0,0 +1,324 @@
...
+    use spirv_std::{Buffer, BufferMut};
+    
+    type T = f32;
+    const N: usize = 1024;
+    
+    #[allow(unused_attributes)]
+    #[spirv(gl_compute(local_size=64)]
+    pub fn scaled_add(
+        #[spirv(descriptor_set=1, binding=0)] x: GlobalBuffer<[T; N]>,
+        #[spirv(descriptor_set=1, binding=1)] mut y: GlobalBufferMut<[T; N]>,
```
Jasper-Bekkers, line 293: I have some slight trouble understanding this as written - does this mean that each thread gets 1024 items available to itself to iterate over, or does this mean that `N` here is the amount of total threads launched? In the latter case, shouldn't this be `GlobalBuffer<[T]>` instead from a usability point of view? I see in your proposal that this is used to create a slice of N long; however this would prose problems for most use-cases as they won't know the size of the `GlobalBuffer` at compile-time right?

It's kind of unfortunate that we can't rely on vulkan's built in [robust buffer access](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#features-robustBufferAccess) support since it just silently ignores out of bounds access

Jasper-Bekkers (commented):
> It does not intend to conflict with #8 but barriers may be inherently necessary to allow mutable access.

The reason I brought up #8 was mostly because it's spiraled a bit out of control in terms of scope; and one of it's goals has become to add proper acquire and release semantics to all buffer loads so we know memory is in a well-defined state at all times. Therefor I would really like to see @Tobski's feedback on this PR. Also because the ideas in this PR map pretty closely to a private discussion @Tobski and I had about iterator-style access to buffers.

docs/src/rfcs/00X-global-buffer.md
charles-r-earp: Not functions, but as a parameter to entry points. All invocations share GlobalBuffer, but the intent is that Global is then a safe, per invocation reference. If the user can just declare a GlobalMut as a parameter, it can freely be accessed without protections against multiple mutation that GlobalBufferMut would provide.

docs/src/rfcs/00X-global-buffer.md
charles-r-earp: The intent is to have both arrays and runtime arrays, with GlobalBuffer<[T; N]> and GlobalBuffer<[T]>. Right now, slices can't be used in entry parameters. 
In this simple example, the buffers have 1024 f32's. I haven't proposed any way to validate that the buffers actually have that length. You are right that generally these would be runtime arrays with some runtime length. 
The compute shader sees 1024 items in each buffer. It is up to GlobalBufferMut, through some safe interface, to ensure that each thread / invocation only sees some exclusive piece of it. In this case, the zip_mut_with function just indexes both buffers by the global index and applies the provided function, which operates on a single pair of scalars. If the global index is out of bounds, the function is not applied.
I think I will edit this to show using slices because the array is confusing.

docs/src/rfcs/00X-global-buffer.md
Jasper-Bekkers: ✅

khyperia:
Closing this due to inactivity, if someone would like to start pushing on this again, feel free to reopen.

PR #284, opened 2020-11-30
Update to use rust-toolchain toml

XAMPPRocky:


khyperia (approved):

rust-toolchain
```diff
@@ -1 +1,3 @@
-nightly-2020-11-24
+[toolchain]
+channel = "nightly-2020-11-24"
+components = ["rust-src", "rustc-dev", "llvm-tools-preview"]
```
eddyb, line 3: Missing newline at end of file.

rust-toolchain
```diff
@@ -1 +1,3 @@
-nightly-2020-11-24
+[toolchain]
```
eddyb, line 1: I've come up with this trick:

```suggestion
# If you see this, run `rustup self update` to get rustup 1.23 or newer.

# NOTE: above comment is for older `rustup` (before TOML support was added),
# which will treat the first line as the toolchain name, and therefore show it
# to the user in the error, instead of "error: invalid channel name '[toolchain]'".

[toolchain]
```

The result of that is that if you're on older `rustup`, you get this error:
```
error: invalid channel name '# If you see this, run `rustup self update` to get rustup 1.23 or newer.' in '/home/eddy/Projects/rust-gpu/rust-toolchain'
error: caused by: invalid toolchain name: '# If you see this, run `rustup self update` to get rustup 1.23 or newer.'
```
instead of the much less informative:
```
error: invalid channel name '[toolchain]' in '/home/eddy/Projects/rust-gpu/rust-toolchain'
error: caused by: invalid toolchain name: '[toolchain]'
```

eddyb:
The macOS failure is https://github.com/rust-lang/rustup/issues/2601 and caused by using `rustup component add rust-src`, *on top of* a "broken" `rust-src` installed by `rustup toolchain install ... --components rust-src` *or* through the TOML `rust-toolchain` file.
I'll try removing the `rustup component add` commands, let's see what happens.

eddyb:
All 3 "Test" CI runs (ubuntu, windows and macOS) succeeded, I'll go fix the rest now, rebase the PR, etc.

eddyb:
Got a bit stuck on `cargo-deny`, [it relies on `rust:1.47-alpine3.12`](https://github.com/EmbarkStudios/cargo-deny-action/blob/19e6cc3cc2ee9e9ef4c0bb7daa95996653aa24de/Dockerfile#L1) and even [latest Alpine doesn't have updated `rustup`](https://pkgs.alpinelinux.org/package/edge/community/x86_64/rustup).

But I think I can tweak the `Dockerfile` to install rustup through the same method this PR used.

eddyb:
Just to cross-link the threads, this is why I think using the TOML `rust-toolchain` shouldn't break as often as `setup.sh`/`setup.bat`: https://github.com/EmbarkStudios/rust-gpu/pull/332#issuecomment-741759083 - at least for RLS, it's down to a coincidence, but it is better than the previous situation where it was easier to break it than to make it work.

.github/workflows/ci.yaml
```diff
@@ -4,7 +4,7 @@ on:
       - main
   pull_request:
 
-name: Continuous integration
+name: CI
```
khyperia, line 7: :tada:

.github/workflows/deploy_docs.yml
```diff
@@ -8,7 +8,8 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v2
-      - run: rustup component add rustfmt clippy rust-src rustc-dev llvm-tools-preview
+      - run: sh <(curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs) -y
+        shell: bash
```
khyperia, line 12: is this needed? I guess it's a little hard to test since deploy_docs doesn't run until it's pushed to main

khyperia (approved):

.github/workflows/deploy_docs.yml
eddyb: Oh wait I forgot to remove it from `deploy_docs.yml` :facepalm:.

As for `rustfmt` and `clippy`, they're not part of the `default` profile AFAICT, so unless we want to add them to `rust-toolchain`, they won't be available.

PR #287, opened 2020-11-30
New structurizer: now with ∞% more φ!

eddyb:
Not sure this is ready, but it seems to handle everything I've thrown at it.

In an effort to test it, I came up with this silly example:
<details>
<summary>(click to open example code + graph rendering)</summary>

```rust
pub fn marker() {}

pub fn shortcircuit_and(a: bool, b: bool) {
    if a && b {
        marker();
    }
}

pub fn nested_while_loop(a: bool, b: bool) {
    while a {
        let mut i = 0;
        loop {
            marker();
            match i {
                0 | 1 | 2 | 4 => {}
                5 => break,
                _ => continue,
            }
            marker();
            i += 1;
        }
        if a & b {
            break;
        }
        if a | b {
            return;
        }
        while b {
            marker();
        }
    }
}
```

[](http://build.lyken.rs/~eddy/tmp/8Ng3/_.svg)

</details>

I was inspired by the loop-specific solution of "threading all exits through one `break` (with a tag indicating which to take when *outside* the loop)" and tried to extend it to handle non-structural acyclic branches as well - the result is this PR.

You can see all the `OpPhi`s, which are used to decide whether a "deferred exit" executes, when it's finally "fully accounted for" (or "owned"? not sure what the best terminology is) by a "region".

Most of the implementation is a SEME -> SESE propagation (with the aforementioned "deferred exits"), which is then used to handle both acyclic and cyclic "regions".

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/mod.rs
+++ b/crates/rustc_codegen_spirv/src/linker/mod.rs
@@ -7,6 +7,7 @@ mod duplicates;
 mod import_export_link;
 mod inline;
 mod mem2reg;
+mod new_structurizer;
 mod simple_passes;
 mod structurizer;
 mod zombies;
@@ -26,6 +27,7 @@ pub struct Options {
     pub inline: bool,
     pub mem2reg: bool,
     pub structurize: bool,
+    pub use_new_structurizer: bool,
 }
 
 fn id(header: &mut ModuleHeader) -> Word {
@@ -136,7 +138,11 @@ pub fn link(sess: &Session, mut inputs: Vec<Module>, opts: &Options) -> Result<M
 
     let mut output = if opts.structurize {
         let _timer = sess.timer("link_structurize");
-        structurizer::structurize(sess, output)
+        if opts.use_new_structurizer {
+            new_structurizer::structurize(output)
+        } else {
+            structurizer::structurize(sess, output)
+        }
     } else {
         output
     };
--- a/crates/rustc_codegen_spirv/src/linker/simple_passes.rs
+++ b/crates/rustc_codegen_spirv/src/linker/simple_passes.rs
@@ -44,8 +44,18 @@ pub fn block_ordering_pass(func: &mut Function) {
             .iter()
             .find(|b| b.label_id().unwrap() == current)
             .unwrap();
+        let mut edges = outgoing_edges(current_block);
+        // HACK(eddyb) treat `OpSelectionMerge` as an edge, in case it points
+        // to an otherwise-unreachable block.
+        if let Some(before_last_idx) = current_block.instructions.len().checked_sub(2) {
+            if let Some(before_last) = current_block.instructions.get(before_last_idx) {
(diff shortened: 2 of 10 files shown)


crates/rustc_codegen_spirv/src/linker/mod.rs
```diff
@@ -136,7 +137,11 @@ pub fn link(sess: &Session, mut inputs: Vec<Module>, opts: &Options) -> Result<M
 
     let mut output = if opts.structurize {
         let _timer = sess.timer("link_structurize");
-        structurizer::structurize(sess, output)
+        if false {
```
khyperia, line 140: `if env::var("OLD_STRUCTURIZER").is_ok()` >:)

or replace the old structurizer, not sure if this is intended to be a replacement or not?

crates/rustc_codegen_spirv/src/linker/simple_passes.rs
```diff
@@ -70,6 +80,7 @@ pub fn block_ordering_pass(func: &mut Function) {
     assert_eq!(func.blocks[0].label_id().unwrap(), entry_label);
 }
 
+// FIXME(eddyb) use `Either`, `Cow`, and/or `SmallVec`.
```
khyperia, line 83: :+1: :white_check_mark: :ok: :100:

khyperia (commented):
I do not believe I am qualified to review `new_structurizer.rs`, nor am I going to try to other than a quick glance-over, because wow okay that's a lot of algorithm. Other than that, though, I think this is good to merge! (not hitting the approve button because mergify scares me)

crates/rustc_codegen_spirv/src/linker/mod.rs
VZout: I mean it looks like this is better in everyway might as well delete my work? or rename mine to `old_structurizer` and this new one to `structurizer` if you want compare it if we run into issues with the new one.

VZout (approved):
I assume cargo test was successful. Seems way better than my code& algorithm. Haven't ran spirv-cfg to see if I can see any weird things.

crates/rustc_codegen_spirv/src/linker/simple_passes.rs
DJMcNab: `Either<Cow, SmallVec>`

crates/spirv-builder/src/test/basic.rs
```diff
@@ -97,11 +97,7 @@ OpFunctionEnd"#,
     );
 }
 
-// TODO: While the "checked sub is not implemented yet" issue is fixed and so this repro should be
-// fixed, a further underlying issue gets triggered instead, which is that the current structurizer
-// doesn't handle this case. Remove `#[ignore]` once fixed.
 #[test]
-#[ignore]
```
khyperia, line 104: heck yee, love yeeting `#[ignore]`s

khyperia (approved):

PR #290, opened 2020-12-01
Split introduction into two separate sections

XAMPPRocky:
This splits up the introduction into two separate sections, one focused on users who want to write shader crates, and the other who want to contribute to Rust-GPU directly.

### DIFF
--- a/docs/src/writing-shader-crates.md
+++ b/docs/src/writing-shader-crates.md
@@ -1,0 +1,75 @@
+# Writing Shader Crates
+
+This is section is going to walk you through writing a shader in Rust and
+setting up your shader crate.
+
+Be aware that this project is in a very early phase, please [file an
+issue](https://github.com/EmbarkStudios/rust-gpu/issues) if there's something
+not working or unclear.
+
+## Setup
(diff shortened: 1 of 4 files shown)


docs/src/writing-shader-crates.md
```diff
@@ -0,0 +1,72 @@
...
+target = "spirv-unknown-unknown"
+rustflags = ["-Zcodegen-backend=<path_to_librustc_codegen_spirv>"]
+```
+
+Now we can build our crate with cargo as normal. We have to provide
+`-Zbuild-std` as the `spirv-unknown-unknown` sysroot is not currently
+available in the default installation.
+
+```bash
+cargo build -Zbuild-std=core
```
khyperia, line 68: This can be specified in `.cargo/config` as well, we should put it there. I think it's something like

```toml
[unstable]
build-std = ["core"]
```

but my memory is fuzzy so we should double-check.

XAMPPRocky: Yep, I was looking for that, but couldn't find until I search for unstable specifically.

khyperia (approved):

PR #292, opened 2020-12-01
No longer reserved descriptor set 0

VZout:


khyperia (approved):

Jasper-Bekkers:
For posterity: this was set 0 reserved as part of #19 - so we could do easy automatic slot and binding assignment. The automatic binding proposal isn't going away, however we've determined that introducing it would be a breaking change regardless of the reservation of set 0 or not and at the moment the reservation is getting in the way of some internal progress.

Jasper-Bekkers (approved):

PR #295, opened 2020-12-01
`#[spirv(block)]` for `Block` decorations on `struct`s.

### LINKED ISSUE #247: sky shader example has validation eror

eddyb:
Fixes #247, but sadly overlaps with #289.

I added a Vulkan validation test to `spirv-builder` tests, hope reusing `spirv_tools` through `rustc_codegen_spirv` is okay.
(Do we care whether that validation is Vulkan 1.2 vs 1.1?)

Also confirmed with `spirv-val --target-env vulkan1.1` that the error is gone from `sky_shader.spv`.

### DIFF
--- a/crates/spirv-builder/src/test/basic.rs
+++ b/crates/spirv-builder/src/test/basic.rs
@@ -1,4 +1,4 @@
-use super::{dis_fn, val};
+use super::{dis_fn, val, val_vulkan};
 
 #[test]
 fn hello_world() {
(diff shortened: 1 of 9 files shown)


crates/spirv-builder/src/test/basic.rs
```diff
@@ -126,3 +126,52 @@ pub fn main() {
...
+pub struct ShaderConstants {
+    pub width: u32,
+    pub height: u32,
+    pub time: f32,
+}
+
+#[allow(unused_attributes)]
+#[spirv(fragment)]
+pub fn main(
+    #[spirv(push_constant)] constants: PushConstant<ShaderConstants>,
```
khyperia, line 145: I believe `#[spirv(push_constant)]` does nothing and was erroneously included in the example shader, which I'm assuming you copied this from. Doesn't really matter though.

khyperia (approved):
:tada:

crates/spirv-builder/src/test/basic.rs
eddyb: Oh haha whoops.

PR #313, opened 2020-12-03
Run Cargo with --locked on CI.

### LINKED ISSUE #303: CI should disallow stale Cargo.lock.
E.g. #301 changed a dependency but didn't include the `Cargo.lock` change.

This should be easy to fix by just adding `--locked` to the CI `cargo` commands, which will error if `Cargo.lock` isn't up to date.

eddyb:
Fixes #303.

Decided to not remove `Cargo.lock` for now given we haven't gotten around to discussing that.

I'm opening this PR without updating `Cargo.lock` at first, so we can confirm the check works.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -57,8 +57,8 @@ jobs:
         run: rustup component add rust-src rustc-dev llvm-tools-preview
       # Fetch dependencies in a separate step to clearly show how long each part
       # of the testing takes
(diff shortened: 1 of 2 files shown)


eddyb:
Success! https://github.com/EmbarkStudios/rust-gpu/actions/runs/398975517 failed. Pushing `Cargo.lock` update now.

khyperia (approved):

Issue #300, opened 2020-12-02
Investigate inferring storage class


khyperia:
Instead of having explicit pointer-wrapping types in `spirv-std` like `Input<T>`, instead, have the compiler not emit storage classes at all (using `Generic` as a placeholder or something, even though that's not what `Generic` storage class means). Then, in the linker, do a whole-program-analysis and propagate storage classes throughout the program. Essentially, treat any place where a pointer type appears as a "generic type parameter" except it's a "generic storage class parameter", then do monomorphization (so duplicating functions that are called with multiple storage classes, etc.).

This would solve many problems:

- Constants could be emitted to OpVariables of Private storage class, and pointers to them could be passed to functions.
- OpTypeRuntimeArray would be able to be supported really easily, as the current `Input<T>` system requires loading a value of `T`, but with OpTypeRuntimeArray, `T` is unsized. Rust's normal rules around this work great, if it were `&T`, you could write `obj.the_unsized_array[5]` and it'd generate the right code with the right storage classes.
- This looks really really nice:
  ```rust
  fn main (#[spirv(input)] input: &f32, #[spirv(output)] output: &mut f32) {}
  ```

charles-r-earp:
Function is the default right now, so just removing the storage class attributes on structs would be enough. I don't think there needs to be a placeholder, just leave them as Function and then in a linker pass it should be possible to propagate the correct storage class. Otherwise all the code that asserts that types (and storage classes) are equal would have to be changed to allow for Generic == Other. 

I would suggest handling block decorations this way as well.

Firstly, is it even possible to have a slice as an argument to a pub fn in a dylib? I thought that Rust doesn't have a stable abi for slices and so this causes issues.

So, do you want to just declare the entry variable with the storage class (but not the pointer?), assuming this doesn't cause issues? This would tag the variable for later, will the pointer storage class would be Function and operate normally. 

This was more or less what I was trying to do in #235. While doable, it would probably need to gather a graph of all the deps of a variable, rather than just a few hashmaps because searching is recursive and deep. This interacts with a whole host of operations, not just access chains and loads / stores, but OpPhi as well. Ideally this would occur earlier than I did it and before a lot of optimizations and instructions, but that requires more reflection. 

Not only does this look nice, it also makes it easier to use "non-gpu" code, with the only real constraint being no_std. Otherwise that code would have to be generic over the storage class, or written specifically for a particular storage class, or purely copy in copy out, can't interact with slices.

[...]

PR #315, opened 2020-12-04
Added new method for specifying storage classes.

charles-r-earp:
Related to #300.  This is a preliminary step that adds the means to do:

    fn main (#[spirv(input)] input: &f32, #[spirv(output)] output: &mut f32) {} 

In this case, the variables input and output would be declared with input and output storage class. In order to actually work, addition logic in the linker stage will have to be added.

This does not conflict and does not remove the current logic, and so is not a breaking change. 

Potentially work related to #300 could be applied to a new "infer-storage-class"  branch, since this this is only useful to users when and if additional functionality is added.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1988,6 +1988,8 @@ dependencies = [
  "pretty_assertions",
  "rspirv",
  "rustc-demangle",
+ "serde",
+ "serde_json",
  "spirv-tools",
  "tar",
  "tempfile",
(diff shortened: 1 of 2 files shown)


khyperia:
This on its own will cause lots of subtle bugs and issues, and cannot be accepted until the full inference engine is in place. The inference engine will be a rather large task to do, and @eddyb has already designed and thought through a lot of the issues in #300, and so the task of doing so is currently assigned to them. So I'm really sorry, but we can't accept this right now without that in place - which will not be in place until 2021, so I'm going to close this for now.

Issue #161, opened 2020-10-27
Create CODEOWNERS


XAMPPRocky:
Talking with @khyperia we were discussing that it would be good to have a CODEOWNERS file to make it easier to identify who can or should merge a given PR.

arirawr:
Sounds like a good call.

repi:
Fixed in #318

PR #318, opened 2020-12-04
Make maintainers explicit with CODEOWNERS

lpil:
Hello!

We're making maintainers explicit. I've guessed the maintainers here, let me know if this is not right. Thanks!

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* khyperia XAMPPRocky
+* @khyperia @XAMPPRocky
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,8 @@
-* @khyperia @XAMPPRocky
+* @khyperia
+
+crates/* @khyperia @eddyb 
+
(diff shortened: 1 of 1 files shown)


termhn:
I think the CODEOWNERS structure should look something like this (not sure if @VZout wants to be included as an examples/ owner or not, and you could probably put @khyperia as one of the owners on every one of these categories if she wants :P)

```
* @khyperia

crates/* @khyperia @eddyb 

examples/* @khyperia @termhn @VZout(?)

docs/* @XAMPPRocky
*.md @XAMPPRocky
```

lpil:
Wonderful, thank you so much @termhn ! I've updated this PR with your suggestion.

khyperia:
Linking https://github.com/EmbarkStudios/rust-gpu/issues/161 because this will close that issue

lpil:
@khyperia Erin is off for a month so we won’t be able to get a review from her. Is there anything else to do here or can we merge?

khyperia:
rrright, yes, sorry, completely forgot about that!

khyperia (approved):

lpil:
Thank you everyone!! 💜

PR #320, opened 2020-12-04
Add support for sampled images

msiglreith:
Implements https://github.com/EmbarkStudios/rust-gpu/issues/204#issuecomment-738634388

Combined image samplers are allowed as resources and require a generic parameter indicating the underlying image type.
Therefore, no constructor ala `let sampled_image = image.combine(sampler);` for `Image2d` right now.

The way the internal spirv type of the underlying image is created feels a bit handwavy to be honest - ideally I would somehow directly check for all constraints (exactly an image type, possible some trait?).


crates/rustc_codegen_spirv/src/abi.rs
```diff
@@ -818,6 +818,25 @@ fn trans_image<'tcx>(
...
+            if ty.size != Size::from_bytes(4) {
+                cx.tcx.sess.err("#[spirv(sampled_image)] type must have size 4");
+                return None;
+            }
+
+            // We use a generic to indicate the underlying image type of the sampled image.
+            // The spirv type of it will be generated by querying the type of the first generic.
+            if let TyKind::Adt(_, substs) = *ty.ty.kind() {
+                // TODO: enforce that the generic param is an image type?
+                let image_ty = substs.type_at(0);
```
khyperia, line 834: Nit: I think this might ICE if it's not a generic type (as in doesn't have a type parameter 0), but considering `#[spirv(sampled_image)]` is only intended to be used in `spirv-std`, that's probably okay.

crates/rustc_codegen_spirv/src/spirv_type.rs
```diff
@@ -440,6 +444,10 @@ impl fmt::Debug for SpirvTypePrinter<'_, '_> {
                 .field("access_qualifier", &access_qualifier)
                 .finish(),
             SpirvType::Sampler => f.debug_struct("Sampler").field("id", &self.id).finish(),
+            SpirvType::SampledImage { image_type } => f
+                .debug_struct("SampledImage")
+                .field("image_type", &self.cx.debug_type(image_type))
```
khyperia, line 451: Nit: this should have `.field("id", &self.id)` too (I've found including the id of self helps a lot when debugging type shenanigans)

crates/rustc_codegen_spirv/src/abi.rs
```diff
@@ -818,6 +818,29 @@ fn trans_image<'tcx>(
...
+                let image_type = trans_type_impl(cx, span, cx.layout_of(image_ty), false);
+                Some(SpirvType::SampledImage { image_type }.def(span, cx))
+            } else {
+                cx.tcx
+                    .sess
+                    .err("#[spirv(sampled_image)] generic type must be an image");
```
khyperia, line 840: Nit: I think this might actually be a `bug!()`? I don't think it's possible to get a non-ADT here (see callsite of `trans_image`, where it's already matching on `TyKind::Adt` - I guess you could grab out the `substs` there and pass them into `trans_image`, if you'd like)

khyperia (commented):
Wow, that was hecking quick, thanks so much! This looks super super good~

I would hit approve, but then mergify would merge it without letting you respond to my tiny lil nitpicks, so doing a comment.

khyperia:
> The way the internal spirv type of the underlying image is created feels a bit handwavy to be honest - ideally I would somehow directly check for all constraints (exactly an image type, possible some trait?).

yeah, haha, there's so much handwaviness right now that's very similar... we probably want to check that somehow eventually and produce a nice error (right now it'll fail `spirv-val` and produce its wacky message), but, considering the amount of similar handwaviness elsewhere, not really worth it to do that now.

msiglreith:
thanks for the quick and super helpful review!

khyperia (approved):
:tada:

PR #321, opened 2020-12-04
Get closer to testing NO_DCE in CI

khyperia:
The only remaining validation issues are due to the SROA pass not being implemented yet

### DIFF
--- a/crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
@@ -212,9 +212,16 @@ impl<'tcx> CodegenCx<'tcx> {
 
         // Create these up front instead of on demand in SpirvValue::def because
         // SpirvValue::def can't use cx.emit()
-        let global_var =
-            self.emit_global()
-                .variable(ty, None, StorageClass::Function, Some(initializer));
+        let global_var = self.emit_global().undef(ty, None);
+
+        // In all likelihood, this zombie message will get overwritten in SpirvValue::def_with_span
+        // to the use site of this constant. However, if this constant happens to never get used, we
+        // still want to zobmie it, so zombie here.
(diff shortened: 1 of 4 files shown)


crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
```diff
@@ -212,9 +212,16 @@ impl<'tcx> CodegenCx<'tcx> {
 
         // Create these up front instead of on demand in SpirvValue::def because
         // SpirvValue::def can't use cx.emit()
-        let global_var =
-            self.emit_global()
-                .variable(ty, None, StorageClass::Function, Some(initializer));
+        let global_var = self.emit_global().undef(ty, None);
```
eddyb, line 215: Is the change to the actual value needed to make this work?

crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
```diff
@@ -212,9 +212,16 @@ impl<'tcx> CodegenCx<'tcx> {
...
+
+        // In all likelihood, this zombie message will get overwritten in SpirvValue::def_with_span
+        // to the use site of this constant. However, if this constant happens to never get used, we
+        // still want to zobmie it, so zombie here.
```
eddyb, line 219: Oops, we probably want a set of `Span`s, I dropped the ball on that.

crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
khyperia: Oh, right, I don't think so no, accidentally included this because I was like "wait why isn't this OpVariable getting removed? uuh, let's try OpUndef? that didn't work either, hmm, oh right it's not getting zombied if it's not used" and then I never reverted it. I'll do so now.

eddyb (approved):

PR #139, opened 2020-10-25
Pin Rust Nightly to 2020-10-25

repi:
Fix #137

PR #332, opened 2020-12-09
Simplify setup.{sh,bat} and work around rust-lang/rustup#2601.

eddyb:
`rustup component add` will install the toolchain if it isn't already, and will respect `rust-toolchain`, so hardcoding the value from `rust-toolchain` into `setup.sh` and `setup.bat` isn't necessary.

It also works around https://github.com/rust-lang/rustup/issues/2601, which I've seen a handful of people hit (including myself), usually due to IDE integration installing the `rust-src` component using `rustup component add`, before `./setup.sh` is ran.

The inclusion of `rust-std` is a second level of workaround, it *somehow* avoids the problem described in https://github.com/rust-lang/rustup/issues/2601#issuecomment-741592459, where running the new `setup.sh`/`setup.bat` for the same toolchain that the old `setup.sh`/`setup.bat` was already ran (and it executed `rustup toolchain install ... --component rust-src`, putting the components list in a broken/suboptimal state).

More specifically, installing `rust-std` (even on its own) makes `rustup` decide to remove the (wrong) `rust-src-$host` component and install the (correct) `rust-src` one, if the former happens to be installed.
(The two versions of `rust-src` are the same component "package", only their names in `lib/rustlib/components` differ)

### DIFF
--- a/setup.bat
+++ b/setup.bat
@@ -1,3 +1,6 @@
 setlocal
 
-rustup toolchain install nightly-2020-11-24 --component rust-src rustc-dev llvm-tools-preview
+rem HACK(eddyb) including `rust-std` works around this `rustup` bug:
+rem https://github.com/rust-lang/rustup/issues/2601
+
+rustup component add rust-std rust-src rustc-dev llvm-tools-preview
--- a/setup.sh
+++ b/setup.sh
@@ -1,3 +1,6 @@
 #!/usr/bin/env bash
 
-rustup toolchain install nightly-2020-11-24 --component rust-src rustc-dev llvm-tools-preview
+# HACK(eddyb) including `rust-std` works around this `rustup` bug:
+# https://github.com/rust-lang/rustup/issues/2601
+
+rustup component add rust-std rust-src rustc-dev llvm-tools-preview


eddyb:
@khyperia mentioned we've gone back & forth on this, and it looks like the short form was suggested in https://github.com/EmbarkStudios/rust-gpu/pull/139#discussion_r511646308, but then later reverted in https://github.com/EmbarkStudios/rust-gpu/commit/d9838a0a5f5c4e817097b37c7124b6a6376bc313 and I don't know why.

If there is a bug with `rustup component add` that I haven't managed to reproduce yet during my https://github.com/rust-lang/rustup/issues/2601 testing, I want to know what it is before we merge this or do anything else with `rustup`.

eddyb:
Oh, #139 getting merged and d9838a0a5f5c4e817097b37c7124b6a6376bc313 were very close together, and I believe I found [the context on Discord](https://discord.com/channels/750717012564770887/750717499737243679/770275479609409546).

Looks like for @repi `rustup component add` wasn't respecting `rust-toolchain`? I haven't ran into that myself.
This is what happens when you don't have the toolchain installed (with the `setup.sh` from this PR):
```console
$ rustup toolchain uninstall nightly-2020-11-24
info: uninstalling toolchain 'nightly-2020-11-24-x86_64-unknown-linux-gnu'
info: toolchain 'nightly-2020-11-24-x86_64-unknown-linux-gnu' uninstalled
$ ./setup.sh
info: syncing channel updates for 'nightly-2020-11-24-x86_64-unknown-linux-gnu'
info: latest update on 2020-11-24, rust version 1.50.0-nightly (d9a105fdd 2020-11-23)
```

And when the toolchain exists, I can use verbose mode to see it downloads the right version:
```console
$ rustup component remove rust-src
info: removing component 'rust-src'
$ rustup -v component add rust-std rust-src
verbose: read metadata version: '12'
info: component 'rust-std' for target 'x86_64-unknown-linux-gnu' is up to date
info: downloading component 'rust-src'
verbose: downloading file from: 'https://static.rust-lang.org/dist/2020-11-24/rust-src-nightly.tar.xz'
```

repi:
Think it was simply failing in CI before I did revert that and went back on installing explicit version. But may work now so that would be good!

eddyb:
Current `main` [uses `rustup component add` on CI](https://github.com/EmbarkStudios/rust-gpu/blame/a942b6c4252b0ba1011295c2ada3031fb7442a80/.github/workflows/ci.yaml#L57-L109) that looks like it hasn't changed since #63.

So I don't see how `setup.sh`/`setup.bat` could affect CI? Is there CI logic outside of that `ci.yaml` file?

eddyb:
Looks like the whole `rust-src` bug is still going to be a problem once we switch to TOML `rust-toolchain` https://github.com/rust-lang/rustup/issues/2601#issuecomment-741739612 - but potentially less so, assuming nothing runs `rustup component add rust-src` directly (RLS probably won't, it checks first).

Also, `rust-std` can't be used as a workaround with the TOML because the workaround only causes `rustup` to fix itself when ran in a *separate* `rustup` invocation than the one which installed the broken `rust-src-$host`.

eddyb:
Looks like TOML works with RLS, but only because of a coincidence: https://github.com/rust-lang/rustup/issues/2601#issuecomment-741757662

I guess I'll go the full TOML route, and I've opened https://github.com/NixOS/nixpkgs/pull/106443 so that I can have a TOML-capable `rustup`.

PR #335, opened 2020-12-10
stable spirv-std 

VZout:
Allow compiling spirv-std on stable.

### DIFF
--- a/crates/spirv-std/src/lib.rs
+++ b/crates/spirv-std/src/lib.rs
@@ -1,7 +1,10 @@
 #![no_std]
-#![feature(register_attr, repr_simd, core_intrinsics, lang_items)]
+#![cfg_attr(
+    target_arch = "spirv",
+    feature(register_attr, repr_simd, core_intrinsics, lang_items)
+)]
 #![cfg_attr(target_arch = "spirv", feature(asm))]
-#![register_attr(spirv)]
+#![cfg_attr(target_arch = "spirv", register_attr(spirv))]
 // Our standard Clippy lints that we use in Embark projects, we opt out of a few that are not appropriate for the specific crate (yet)
 #![warn(
     clippy::all,
--- a/crates/spirv-std/src/lib.rs
+++ b/crates/spirv-std/src/lib.rs
@@ -1,10 +1,9 @@
 #![no_std]
 #![cfg_attr(
     target_arch = "spirv",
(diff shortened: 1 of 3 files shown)


crates/spirv-std/src/storage_class.rs
```diff
@@ -36,7 +36,7 @@ macro_rules! storage_class {
             /// Store the value in storage.
             #[inline]
             #[allow(unused_attributes)]
-            #[spirv(really_unsafe_ignore_bitcasts)]
+            #[cfg_attr(target_arch = "spirv", spirv(really_unsafe_ignore_bitcasts))]
```
eddyb, line 39: Can you move the `allow` into the `cfg_attr`? I think a single `cfg_attr` supports multiple attributes nowadays, but the formatting might be a bit weird.

eddyb: Also, this is a random place I ended up commenting, all of the changed attributes also have `allow`s.

VZout: done

crates/spirv-std/src/lib.rs
```diff
@@ -1,7 +1,10 @@
 #![no_std]
-#![feature(register_attr, repr_simd, core_intrinsics, lang_items)]
+#![cfg_attr(
+    target_arch = "spirv",
+    feature(register_attr, repr_simd, core_intrinsics, lang_items)
+)]
 #![cfg_attr(target_arch = "spirv", feature(asm))]
-#![register_attr(spirv)]
+#![cfg_attr(target_arch = "spirv", register_attr(spirv))]
```
eddyb, line 7: I think this can be a single `cfg_attr` now.

crates/spirv-std/src/textures.rs
```diff
@@ -1,22 +1,22 @@
 use glam::{Vec2, Vec3A, Vec4};
 
 #[allow(unused_attributes)]
-#[spirv(sampler)]
+#[cfg_attr(target_arch = "spirv", spirv(sampler))]
```
eddyb, line 4: This file has a `allow`s everywhere as well.

crates/spirv-std/src/lib.rs
VZout: Missed this comment yesterday :)

eddyb (approved):

PR #340, opened 2020-12-11
Compile examples and spirv-std on stable

VZout:
* Reverts 06de5ba6efc75963c9147e12d8acec04e997c8b0 in favor of `spirv-attrib`
* allows compiling spirv-std and shaders on the CPU with stable.
* allows compiling shaders on the GPU from a stable repo without having to nest workspaces. (Requires custom builder)

This would be a huge step forward for ark integration.


examples/shaders/shared/src/lib.rs
```diff
@@ -13,9 +19,9 @@ use spirv_std::glam::{vec3, Vec3};
 #[cfg(target_arch = "spirv")]
 use spirv_std::num_traits::Float;
 
+#[spirv(block)]
 #[derive(Copy, Clone)]
 #[allow(unused_attributes)]
-#[spirv(block)]
```
khyperia, line 18: any reason for this, or just formatting?

crates/spirv-attrib/src/lib.rs
```diff
@@ -0,0 +1,16 @@
...
+#[cfg(not(target_arch = "spirv"))]
+#[proc_macro_attribute]
+pub fn spirv(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut item_str = item.to_string();
+
+    while let Some(start) = item_str.find("#[spirv(") {
+        let end = item_str[start..item_str.len()].find(")]");
+        item_str.replace_range(start..start + end.unwrap() + 2, "");
+    }
+    TokenStream::from_str(item_str.as_str()).unwrap()
```
khyperia, line 15: the quadratic reparse time here makes me sad, but, I'll leave this up to @eddyb if they think this is OK since they know more about proc macros than me

khyperia (approved):

crates/spirv-attrib/src/lib.rs
eddyb, line 15: Can you implement this by looking for `[...]` groups instead (preceded by `#`), in the `TokenStream`? Using strings feels very error-prone.

Ideally you could/would use `syn` but that seems like it would be a lot more work.

crates/spirv-attrib/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+#![cfg_attr(target_arch = "spirv", feature(register_attr), register_attr(spirv))]
+
+use proc_macro::TokenStream;
+use std::str::FromStr;
+
+#[cfg(not(target_arch = "spirv"))]
```
eddyb, line 6: This won't do anything, I don't think, Cargo will only ever compile this for the host.

crates/spirv-attrib/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+#![cfg_attr(target_arch = "spirv", feature(register_attr), register_attr(spirv))]
```
eddyb, line 1: I don't think this does anything?

examples/shaders/shared/src/lib.rs
VZout: You can't put proc macros bellow derives.

crates/spirv-attrib/src/lib.rs
VZout: "quadratic reparse time" makes me sound really smart.

crates/spirv-attrib/src/lib.rs
VZout: Why not? It only registers the spriv attribute on spirv architecture.

crates/spirv-attrib/src/lib.rs
VZout: I think it did compile it unless I put the crate behind a [target.dependency] thing

crates/spirv-attrib/src/lib.rs
VZout: Oh your right 👍

crates/spirv-attrib/src/lib.rs
VZout: Yep your right doesn't get compiled.

PR #341, opened 2020-12-12
Mouse support for the wgpu runner, and a shader to demonstrate it.

eddyb:
The integration is similar to https://github.com/LykenSol/rust-gpu-shadertoys/pull/6 and https://github.com/LykenSol/rust-gpu-shadertoys/pull/12, but generalized to support all 3 mouse buttons (Left, Middle, Right). Only dragging is limited to a specific mouse button (Left).

In order to demonstrate the functionality, I've also written a new shader, using an SDF-like abstraction (AFAICT - at least it seems similar to SDF) to render basic shapes.
It also includes a colored background to interactively distort (for which tweaking/replacement is welcome - I didn't want to copy existing code so I had to make do with my own `sin`s).

In the screenshot below:
* the arrow indicates the drag trajectory (which is also what is distorting an otherwise more circular color pattern)
* the vaguely-mouse-looking circle is centered on the cursor
* the Middle and Right mouse buttons were recently pressed in quick succession, and so they are in the process of turning their color from red ("the button went down *just now*") to white ("the button is being held down")
  * the Middle button was pressed earlier, and thus it's closer to white than the Right button

Screenshot_20201212_015055

### DIFF
--- a/examples/shaders/mouse-shader/src/lib.rs
+++ b/examples/shaders/mouse-shader/src/lib.rs
@@ -1,0 +1,270 @@
+#![cfg_attr(target_arch = "spirv", no_std)]
+#![feature(lang_items)]
+#![feature(register_attr)]
+#![register_attr(spirv)]
+
+use core::f32::consts::PI;
+use shared::*;
+use spirv_std::glam::{const_vec4, vec2, vec3, Mat2, Vec2, Vec3, Vec4, Vec4Swizzles};
+use spirv_std::storage_class::{Input, Output, PushConstant};
+
+// Note: This cfg is incorrect on its surface, it really should be "are we compiling with std", but
+// we tie #[no_std] above to the same condition, so it's fine.
+#[cfg(target_arch = "spirv")]
+use spirv_std::num_traits::Float;
+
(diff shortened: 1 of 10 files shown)


examples/shaders/mouse-shader/src/lib.rs
```diff
@@ -0,0 +1,272 @@
+//! Ported to Rust from https://github.com/Tw1ddle/Sky-Shader/blob/master/src/shaders/glsl/sky.fragment
```
khyperia, line 1: Nit: remove

VZout (approved):

khyperia (approved):

examples/shaders/mouse-shader/src/lib.rs
eddyb: Oops, that's embarrassing :no_mouth:.

termhn (approved):
I'm *slightly* skeptical that we want to actually move in this direction with the example (runners), but the code looks great.

What you've written pretty much *is* SDFs btw ;D

eddyb:
@Mergifyio refresh

Issue #209, opened 2020-11-04
ICE in constant.rs when trying to compile liballoc


XAMPPRocky:
https://github.com/EmbarkStudios/rust-gpu/blob/afe3e1f83fe68b35d57efe23322f896aae6aa2ae/crates/rustc_codegen_spirv/src/codegen_cx/constant.rs#L505

<details><summary>Backtrace</summary>
<p>

```
thread 'rustc' panicked at 'called `Option::unwrap()` on a `None` value', crates/rustc_codegen_spirv/src/codegen_cx/constant.rs:505:79
stack backtrace:
   0:        0x10f175a9b - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h52fd820e22c41ab1
   1:        0x10f1e43a5 - core::fmt::write::h4c3abbf7ea400217
   2:        0x10f16a108 - std::io::Write::write_fmt::h995b6c278ed64243
   3:        0x10f175916 - std::sys_common::backtrace::print::h0df2d2a70cc86088
   4:        0x10f1a3cc4 - std::panicking::default_hook::{{closure}}::ha9c562449f7ea8fc
   5:        0x10f1a3adf - std::panicking::default_hook::h828dbd064cd66397
   6:        0x10770e0de - rustc_driver::report_ice::h4db77daa09b7e79b
   7:        0x10f717c4c - rustc_codegen_spirv::__rustc_codegen_backend::{{closure}}::hbf7712550e30b599
   8:        0x10f1a4151 - std::panicking::rust_panic_with_hook::hfc932780c9abefa9
   9:        0x10f16d2a9 - std::panicking::begin_panic_handler::{{closure}}::h03c86b72aa705909
  10:        0x10f16d238 - std::sys_common::backtrace::__rust_end_short_backtrace::hb2534239223eb56d
  11:        0x10f1a3d33 - _rust_begin_unwind
  12:        0x10f2072ff - core::panicking::panic_fmt::h48d4d1671d5d8620
  13:        0x10f207257 - core::panicking::panic::h9c825b9663dfa60c
  14:        0x10f643ea0 - rustc_codegen_spirv::codegen_cx::constant::<impl rustc_codegen_spirv::codegen_cx::CodegenCx>::create_const_alloc2::hfabdbc0fd9407b38
[...]

PR #342, opened 2020-12-14
fix: handle an ICE resulting from a dangling pointer

DeanBDean:
This fixes https://github.com/EmbarkStudios/rust-gpu/issues/209 . It follows [this logic](https://github.com/EmbarkStudios/rust-gpu/issues/209#issuecomment-722206191) from the comments in the issue. It also creates an ignored test for another issue

### DIFF
--- a/crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
@@ -6,12 +6,12 @@ use rspirv::spirv::Word;
 use rustc_codegen_ssa::mir::place::PlaceRef;
 use rustc_codegen_ssa::traits::{ConstMethods, MiscMethods, StaticMethods};
 use rustc_middle::bug;
-use rustc_middle::mir::interpret::{read_target_uint, Allocation, GlobalAlloc, Pointer};
+use rustc_middle::mir::interpret::{AllocId, Allocation, GlobalAlloc, Pointer, ScalarMaybeUninit};
 use rustc_middle::ty::layout::TyAndLayout;
 use rustc_mir::interpret::Scalar;
 use rustc_span::symbol::Symbol;
 use rustc_span::{Span, DUMMY_SP};
-use rustc_target::abi::{self, AddressSpace, HasDataLayout, LayoutOf, Primitive, Size};
+use rustc_target::abi::{self, AddressSpace, HasDataLayout, Integer, LayoutOf, Primitive, Size};
 
 impl<'tcx> CodegenCx<'tcx> {
     pub fn constant_u8(&self, span: Span, val: u8) -> SpirvValue {
@@ -215,7 +215,7 @@ impl<'tcx> ConstMethods<'tcx> for CodegenCx<'tcx> {
     ) -> Self::Value {
         match scalar {
             Scalar::Int(int) => {
-                assert_eq!(int.size(), layout.value.size(&self.tcx));
+                assert_eq!(int.size(), layout.value.size(self));
                 let data = int.to_bits(int.size()).unwrap();
 
                 match layout.value {
@@ -251,7 +251,19 @@ impl<'tcx> ConstMethods<'tcx> for CodegenCx<'tcx> {
                         res
                     }
                     Primitive::Pointer => {
-                        bug!("scalar_to_backend Primitive::Ptr is an invalid state")
+                        if data == 0 {
+                            self.constant_null(ty)
+                        } else {
+                            let result = self.undef(ty);
+                            // HACK(eddyb) we don't know whether this constant originated
+                            // in a system crate, so it's better to always zombie.
+                            self.zombie_even_in_user_code(
+                                result.def_cx(self),
+                                DUMMY_SP,
+                                "pointer has non-null integer address",
+                            );
+                            result
+                        }
                     }
                 }
             }
@@ -347,6 +359,17 @@ impl<'tcx> ConstMethods<'tcx> for CodegenCx<'tcx> {
 }
 
 impl<'tcx> CodegenCx<'tcx> {
+    // This function comes from https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_middle/ty/layout.rs.html,
+    // and is a lambda in the `layout_raw_uncached` function in Version 1.50.0-nightly (eb4fc71dc 2020-12-17)
+    pub fn primitive_to_scalar(&self, value: Primitive) -> abi::Scalar {
+        let bits = value.size(self.data_layout()).bits();
+        assert!(bits <= 128);
+        abi::Scalar {
+            value,
+            valid_range: 0..=(!0 >> (128 - bits)),
+        }
+    }
+
     pub fn create_const_alloc(&self, alloc: &Allocation, ty: Word) -> SpirvValue {
         // println!(
         //     "Creating const alloc of type {} with {} bytes",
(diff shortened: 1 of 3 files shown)


crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
```diff
@@ -481,17 +482,20 @@ impl<'tcx> CodegenCx<'tcx> {
...
+                        ptr.into(),
+                        &abi::Scalar {
+                            value: Primitive::Pointer,
+                            valid_range: 0..=!0,
+                        },
+                        ty,
+                    ),
+                    Scalar::Int(_) => self.constant_null(ty),
+                },
+                None => self.const_undef(ty),
```
eddyb, line 497: This should probably be `zombie`'d as per https://github.com/EmbarkStudios/rust-gpu/issues/209#issuecomment-742771895 - at least I don't think a plain undef value is correct, assuming SPIR-V used any of the existing meanings of "undef".

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
```diff
@@ -541,8 +546,24 @@ impl<'tcx> CodegenCx<'tcx> {
...
+                    inner_offset_size,
+                    (),
+                )))
+            }
+            None => {
+                if bytes.iter().any(|byte| *byte != 0) {
+                    None
+                } else {
+                    Some(Scalar::Int(ScalarInt::null(inner_offset_size)))
+                }
```
khyperia, line 565: Does `offset` need to be advanced in this branch?

crates/spirv-builder/src/test/basic.rs
```diff
@@ -309,3 +309,37 @@ pub fn main(i: Input<f32>, mut o: Output<f32>) {
...
+    let _pointer = POINTER;
+}"#);
+}
+
+#[test]
+fn allocate_vec_like_pointer() {
+    dis_fn(
+        r#"
+use core::ptr::Unique;
+const VEC_LIKE: (Unique<usize>, usize, usize) = (Unique::<usize>::dangling(), 0, 0);
```
khyperia, line 332: Could we also test the nullptr case?

crates/spirv-builder/src/test/basic.rs
```diff
@@ -309,3 +309,37 @@ pub fn main(i: Input<f32>, mut o: Output<f32>) {
...
+#[allow(unused_attributes)]
+#[spirv(fragment)]
+pub fn allocate_vec_like_pointer() {
+    let _vec_like = VEC_LIKE;
+}"#,
+        "allocate_vec_like_pointer",
+        r#"%1 = OpFunction %2 None %3
+%4 = OpLabel
+OpReturn
+OpFunctionEnd"#,
```
khyperia, line 343: There is nothing interesting going on in this function that we're disassembling here, should we use `val` instead of `dis_fn` here?

khyperia (changes requested):

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
```diff
@@ -530,8 +534,9 @@ impl<'tcx> CodegenCx<'tcx> {
         read_target_uint(self.data_layout().endian, bytes).unwrap()
     }
 
-    // Advances offset by ptr size
-    fn read_alloc_ptr<'a>(&self, alloc: &'a Allocation, offset: &mut Size) -> Pointer {
+    // Advances offset by ptr size, None indicates that the Scalar did not have relocations
+    // and that it was non zero
+    fn read_alloc_ptr<'a>(&self, alloc: &'a Allocation, offset: &mut Size) -> Option<Scalar> {
```
eddyb, line 539: Just so it doesn't get forgotten, we've discussed on Discord about replacing these two functions with `Allocation::read_scalar` from rustc itself.

eddyb (changes requested):

crates/spirv-builder/src/test/basic.rs
DeanBDean: Yes, I'll make a test case for nullptr

crates/spirv-builder/src/test/basic.rs
DeanBDean: The only reason I did this was for my own edification to understand the output better, I'll change it to val if it's not useful

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
```diff
@@ -374,23 +386,88 @@ impl<'tcx> CodegenCx<'tcx> {
...
-                let v = self.read_alloc_val(alloc, offset, (width / 8) as usize);
-                match width {
-                    32 => self.constant_f32(DUMMY_SP, f32::from_bits(v as u32)),
-                    64 => self.constant_f64(DUMMY_SP, f64::from_bits(v as u64)),
-                    other => self
+            SpirvType::Bool
+            | SpirvType::Integer(..)
+            | SpirvType::Float(_)
+            | SpirvType::Pointer { .. } => {
+                let (primitive, size) = match ty_concrete {
```
khyperia, line 393: why not `let size = ty_concrete.sizeof(self);`? would save quite a few lines of code and duplication

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
```diff
@@ -374,23 +386,88 @@ impl<'tcx> CodegenCx<'tcx> {
...
+                        .sizeof(self)
+                        .unwrap(),
+                    ),
+                    unsupported_spirv_type => bug!(
+                        "invalid spirv type internal to create_alloc_const2: {:?}",
+                        unsupported_spirv_type
+                    ),
+                };
+                let scalar = match alloc
+                    .read_scalar(self, Pointer::new(AllocId(0), *offset), size)
```
khyperia, line 451: maybe a comment would be good pointing out how `AllocId` isn't used and therefore doesn't need to be filled in, it's pretty confusing as-is.

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
```diff
@@ -374,23 +386,88 @@ impl<'tcx> CodegenCx<'tcx> {
...
+                {
+                    ScalarMaybeUninit::Scalar(scalar) => scalar,
+                    ScalarMaybeUninit::Uninit => self
                         .tcx
                         .sess
-                        .fatal(&format!("invalid float width {}", other)),
-                }
+                        .fatal("allocation pointed at uninitialized data"),
+                };
+                let scalar_unit = |value: Primitive| {
```
khyperia, line 460: why is this a lambda? why not straight code?

crates/spirv-builder/src/test/mod.rs
```diff
@@ -49,7 +49,7 @@ spirv-std = { path = "../../crates/spirv-std" }
 "#;
 
 static SRC_PREFIX: &str = r#"#![no_std]
-#![feature(register_attr, asm)]
+#![feature(register_attr, asm, dropck_eyepatch, ptr_internals)]
```
khyperia, line 52: why is `dropck_eyepatch` added?

khyperia (changes requested):

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
DeanBDean: Good call, the source was a lambda and I never cleaned it up. Will do

crates/spirv-builder/src/test/mod.rs
DeanBDean: When I initially made the test, dropck_eyepatch was required when instantiating the Unique dangling pointer at runtime. I think when I shifted to creating the pointer during the compile as a result of direction from Discord that it must have no longer been needed. I will remove it

khyperia (approved):
The holidays have completely flushed my cache of understanding of what the heck is going on here, but it loooks good, I think? I definitely want @eddyb to look at this too before merging, though.

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
```diff
@@ -251,7 +251,19 @@ impl<'tcx> ConstMethods<'tcx> for CodegenCx<'tcx> {
...
+                        if data == 0 {
+                            self.constant_null(ty)
+                        } else {
+                            let result = self.undef(ty);
+                            // HACK(eddyb) we don't know whether this constant originated
+                            // in a system crate, so it's better to always zombie.
+                            self.zombie_even_in_user_code(
+                                result.def_cx(self),
+                                DUMMY_SP,
+                                format!("pointer refers to non empty data: {}", data).as_str(),
```
eddyb, line 263: I'd replace this message with "unsupported integer to pointer cast" or similar.
Or if you want to keep the current style: "pointer has non-null integer integer address".

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
```diff
@@ -347,6 +359,17 @@ impl<'tcx> ConstMethods<'tcx> for CodegenCx<'tcx> {
 }
 
 impl<'tcx> CodegenCx<'tcx> {
+    // This function comes from https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_middle/ty/layout.rs.html,
+    // and is a lambda in the `layout_raw_uncached` function in Version 1.50.0-nightly (eb4fc71dc 2020-12-17)
+    pub fn create_abi_scalar_from_primitive(&self, value: Primitive) -> abi::Scalar {
```
eddyb, line 364: Can just be named `scalar_from_primitive` or `primitive_to_scalar`.

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
```diff
@@ -374,23 +397,65 @@ impl<'tcx> CodegenCx<'tcx> {
...
+                        unsupported_spirv_type
+                    ),
+                };
+                // alloc_id is not needed by read_scalar, so we just use 0. If the context
+                // refers to a pointer, read_scalar will find the the actual alloc_id. It
+                // only uses the input alloc_id in the case that the scalar is uninitilized
+                // as part of the error output
+                // tldr, the pointer here is only needed for the offset
+                let scalar = match alloc
+                    .read_scalar(self, Pointer::new(AllocId(0), *offset), size)
```
eddyb, line 443: If you pass `!0` instead of `0` in the `AllocId(...)`, it might be harder for this to go wrong (cc @oli-obk).

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
```diff
@@ -374,23 +397,65 @@ impl<'tcx> CodegenCx<'tcx> {
...
+                    .unwrap()
+                {
+                    ScalarMaybeUninit::Scalar(scalar) => scalar,
+                    ScalarMaybeUninit::Uninit => self
                         .tcx
                         .sess
-                        .fatal(&format!("invalid float width {}", other)),
-                }
+                        .fatal("allocation pointed at uninitialized data"),
```
eddyb, line 450: This is actually fine, you could use `self.undef(ty)` for it, but idk how you would test that codepath (`union`s?).

eddyb (changes requested):
This is more or less good to go, I only have a few nit comments. Sorry for the delay, feel free to ping me on Discord when something is waiting for me!

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
oli-obk: we use `0` in the compiler, too, so it seems fine. If we change anything here it'll be changing the datastructure in a way that needs code changes, so there won't be silent breakage or something.

DeanBDean: In lieu of the above, I just left it as 0

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
DeanBDean: I changed this and added a test to verify this branch

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
```diff
@@ -374,23 +397,62 @@ impl<'tcx> CodegenCx<'tcx> {
...
+                // as part of the error output
+                // tldr, the pointer here is only needed for the offset
+                let scalar = match alloc
+                    .read_scalar(self, Pointer::new(AllocId(0), *offset), size)
+                    .unwrap()
+                {
+                    ScalarMaybeUninit::Scalar(scalar) => scalar,
+                    ScalarMaybeUninit::Uninit => {
+                        *offset += size;
+                        return self.undef(ty);
```
eddyb, line 449: I think this would be cleaner if you moved the `scalar_to_backend` call inside this `match`, as then the `Uninit` case would be just `=> self.undef(ty)`.

eddyb (approved):

XAMPPRocky:
Thank you for your PR, and congrats on your first contribution! 🎉

PR #239, opened 2020-11-13
Refactor spirv-builder to build codegen and sysroot itself

XAMPPRocky:
This PR replaces building `rustc_codegen_spirv` and `spirv-unknown-unknown` with cargo in spirv-builder, with building each as a separate step. Currently this is actually slower as the tests rebuild those artefacts for each invocation, but I'm planning on finishing up the ability to use pre-built artefacts for tests before marking this PR as not a draft. 

Huge thanks to @diananites who added a lot of functionality to `cargo-sysroot` to make this possible. 🎉

PR #345, opened 2020-12-14
Builder now always builds with nightly

VZout:
This allows a lot more flexibility in using the builder in stable repositories.
Note that it doesn't install components.

### DIFF
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -149,6 +149,7 @@ fn invoke_rustc(builder: &SpirvBuilder) -> Result<PathBuf, SpirvBuilderError> {
     );
     let mut cargo = Command::new("cargo");
     cargo.args(&[
+        "+nightly-2020-12-11",
         "build",
         "--message-format=json-render-diagnostics",
         "-Z",


khyperia (changes requested):
If we do go this route, could we add a test to make sure `rust-toolchain` and this string match? I can absolutely see this accidentally getting out of date.

For example, extract this string to a const, and write a test to make sure `rust-toolchain` contains the string.

eddyb:
You can use `include_str!` + parsing the toml in a (`spirv-builder/build.rs`) build script, to compute the value, let Cargo record it, and then `env!(...)` to read it in `spirv-builder/src/lib.rs` (assuming we don't want to parse the TOML at runtime, but maybe a simple `toolchain = "` string search would work?).

But also we could bundle the `rust-toolchain` with the `spirv-builder` and drop it in the target directory, or in some temporary directory elsewhere, and use that directory as the working dir for the nested Cargo invocation.

VZout:
> You can use `include_str!` + parsing the toml in a (`spirv-builder/build.rs`) build script, to compute the value, let Cargo record it, and then `env!(...)` to read it in `spirv-builder/src/lib.rs` (assuming we don't want to parse the TOML at runtime, but maybe a simple `toolchain = "` string search would work?).
> 
> But also we could bundle the `rust-toolchain` with the `spirv-builder` and drop it in the target directory, or in some temporary directory elsewhere, and use that directory as the working dir for the nested Cargo invocation.

Your second suggestion doesn't work for this repo since if you call `cargo build` from root it will try to build the codegen crate on stable since its in the top level workspace.

VZout:
Depends on https://github.com/EmbarkStudios/rust-gpu/pull/239

PR #346, opened 2020-12-14
Better spirv-attrib

VZout:
Feedback related to https://github.com/EmbarkStudios/rust-gpu/pull/340

proc macro pseudo code:
```markdown
create empty vec of tokens

for input tokenstream
    if token is group
        create token list for inner tokenstream
        for token in group's tokenstream
            if token is bracket && starts with [spirv
                don't insert token, pop last token as well to rmeove #
            else
                insert token into inner token vec.
            insert new token created from the inner token vec
    else
        insert token into token vec.

return new tokenstream from vec of tokens.
```

How can I do this in a better way?

### DIFF
--- a/crates/spirv-attrib/src/lib.rs
+++ b/crates/spirv-attrib/src/lib.rs
@@ -5,34 +5,31 @@ use proc_macro::{Delimiter, Group, TokenStream, TokenTree};
 pub fn spirv(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let mut tokens = Vec::new();
     for tt in item {
-        match tt {
-            TokenTree::Group(group) => match group.delimiter() {
-                Delimiter::Parenthesis => {
-                    let mut sub_tokens = Vec::new();
-                    for tt in group.stream() {
-                        match tt {
-                            TokenTree::Group(group) => match group.delimiter() {
-                                Delimiter::Bracket => {
-                                    if group.stream().to_string().starts_with("spirv") {
-                                        sub_tokens.pop();
-                                    } else {
-                                        sub_tokens.push(TokenTree::from(group));
-                                    }
-                                }
-                                _ => sub_tokens.push(TokenTree::from(group)),
-                            },
-                            _ => sub_tokens.push(tt),
+        if let TokenTree::Group(group) = tt {
+            if let Delimiter::Parenthesis = group.delimiter() {
+                let mut sub_tokens = Vec::new();
+                for tt in group.stream() {
+                    if let TokenTree::Group(group) = tt {
+                        if let Delimiter::Bracket = group.delimiter() {
+                            if group.stream().to_string().starts_with("spirv") {
+                                sub_tokens.pop();
+                            } else {
+                                sub_tokens.push(TokenTree::from(group));
+                            }
                         }
+                    } else {
+                        sub_tokens.push(tt);
                     }
-
-                    tokens.push(TokenTree::from(Group::new(
-                        Delimiter::Parenthesis,
-                        TokenStream::from_iter(sub_tokens),
-                    )));
                 }
-                _ => tokens.push(TokenTree::from(group)),
-            },
-            _ => tokens.push(tt),
+                tokens.push(TokenTree::from(Group::new(
+                    Delimiter::Parenthesis,
+                    TokenStream::from_iter(sub_tokens),
+                )));
+            } else {
+                tokens.push(TokenTree::from(group));
+            }
+        } else {
+            tokens.push(tt);
         }
     }
 
--- a/crates/spirv-attrib/src/lib.rs
+++ b/crates/spirv-attrib/src/lib.rs
@@ -2,7 +2,7 @@ use proc_macro::{Delimiter, Group, TokenStream, TokenTree};
 
 #[proc_macro_attribute]
 pub fn spirv(_attr: TokenStream, item: TokenStream) -> TokenStream {
-    let mut tokens: Vec<TokenTree> = Vec::new();
+    let mut tokens = Vec::new();
     for tt in item {
         match tt {
             TokenTree::Group(group) if group.delimiter() == Delimiter::Parenthesis => {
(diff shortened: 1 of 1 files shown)


crates/spirv-attrib/src/lib.rs
```diff
@@ -1,16 +1,40 @@
...
+    for tt in item {
+        match tt {
+            TokenTree::Group(group) => match group.delimiter() {
+                Delimiter::Parenthesis => {
+                    let mut sub_tokens = Vec::new();
+                    for tt in group.stream() {
+                        match tt {
+                            TokenTree::Group(group) => match group.delimiter() {
+                                Delimiter::Bracket => {
+                                    if group.stream().to_string().starts_with("spirv") {
```
eddyb, line 16: You can use something like `matches!(group.stream.next(), TokenTree::Ident(ident) if ident == "spirv"))` to avoid stringifying the entire attribute.

crates/spirv-attrib/src/lib.rs
```diff
@@ -1,16 +1,40 @@
...
+                                        sub_tokens.pop();
+                                    } else {
+                                        sub_tokens.push(TokenTree::from(group));
+                                    }
+                                }
+                                _ => sub_tokens.push(TokenTree::from(group)),
+                            },
+                            _ => sub_tokens.push(tt),
```
eddyb, line 24: You only need the last one (and outside the `match`) if you make the one that skips use `continue;`.
To not consume `tt` you can use `match &tt` instead of `match tt`.

crates/spirv-attrib/src/lib.rs
```diff
@@ -1,16 +1,40 @@
...
+        match tt {
+            TokenTree::Group(group) => match group.delimiter() {
+                Delimiter::Parenthesis => {
+                    let mut sub_tokens = Vec::new();
+                    for tt in group.stream() {
+                        match tt {
+                            TokenTree::Group(group) => match group.delimiter() {
+                                Delimiter::Bracket => {
+                                    if group.stream().to_string().starts_with("spirv") {
+                                        sub_tokens.pop();
```
eddyb, line 17: You can check whether `sub_tokens.last()` is `TokenTree::Punct` of `'#'` at the same time you're checking whether the brackets have `spirv` inside, so that you don't handle anything that's not an attribute accidentally.

eddyb (changes requested):

crates/spirv-attrib/src/lib.rs
VZout: done

crates/spirv-attrib/src/lib.rs
VZout: done

crates/spirv-attrib/src/lib.rs
VZout: done

crates/spirv-attrib/src/lib.rs
```diff
@@ -1,16 +1,34 @@
...
 pub fn spirv(_attr: TokenStream, item: TokenStream) -> TokenStream {
-    let mut item_str = item.to_string();
-
-    while let Some(start) = item_str.find("#[spirv(") {
-        let end = item_str[start..item_str.len()].find(")]");
-        item_str.replace_range(start..start + end.unwrap() + 2, "");
+    let mut tokens = Vec::new();
+    for tt in item {
+        if let TokenTree::Group(group) = tt.clone() {
+            if let Delimiter::Parenthesis = group.delimiter() {
```
eddyb, line 10: You can use the same `match` + guard technique here as below.

crates/spirv-attrib/src/lib.rs
```diff
@@ -1,16 +1,34 @@
-#![cfg_attr(target_arch = "spirv", feature(register_attr), register_attr(spirv))]
+use core::iter::FromIterator;
+use core::iter::IntoIterator;
```
eddyb, line 2: Pretty sure these are both imported by default? Or at least `IntoIterator` should be.

crates/spirv-attrib/src/lib.rs
```diff
@@ -1,16 +1,34 @@
...
+                    Delimiter::Parenthesis,
+                    TokenStream::from_iter(sub_tokens),
+                )));
+                continue;
+            }
+        }
+        tokens.push(tt);
     }
-    TokenStream::from_str(item_str.as_str()).unwrap()
+    TokenStream::from_iter(tokens)
```
eddyb, line 33: Maybe this (and the other `from_iter` call) should/could be `tokens.into_iter().collect()`? I'm surprised `from_iter` can even take a `Vec`.

eddyb (changes requested):

crates/spirv-attrib/src/lib.rs
VZout: From isn't but yeah apparently Into is.

crates/spirv-attrib/src/lib.rs
VZout: `tokens.into_iter().collect()` Actually doesn't compile. it doesn't satisfy the FromIterator trait.

crates/spirv-attrib/src/lib.rs
```diff
@@ -1,16 +1,32 @@
...
-    while let Some(start) = item_str.find("#[spirv(") {
-        let end = item_str[start..item_str.len()].find(")]");
-        item_str.replace_range(start..start + end.unwrap() + 2, "");
+    let mut tokens = Vec::new();
+    for tt in item {
+        match tt.clone() {
+            TokenTree::Group(group) if group.delimiter() == Delimiter::Parenthesis => {
+                let mut sub_tokens = Vec::new();
+                for tt in group.stream() {
+                    match tt.clone() {
```
eddyb, line 12: You should be able to write `match &tt` here.

crates/spirv-attrib/src/lib.rs
```diff
@@ -1,16 +1,32 @@
...
 #[proc_macro_attribute]
 pub fn spirv(_attr: TokenStream, item: TokenStream) -> TokenStream {
-    let mut item_str = item.to_string();
-
-    while let Some(start) = item_str.find("#[spirv(") {
-        let end = item_str[start..item_str.len()].find(")]");
-        item_str.replace_range(start..start + end.unwrap() + 2, "");
+    let mut tokens = Vec::new();
+    for tt in item {
+        match tt.clone() {
```
eddyb, line 8: Oh also here, I missed this.

eddyb (approved):

PR #348, opened 2020-12-15
spriv-std 0.2.1

VZout:
* Reverts breaking change https://github.com/EmbarkStudios/rust-gpu/pull/327


khyperia:
The destination branch for this shouldn't be `main`, and should never be `main`, right?

Issue #349, opened 2020-12-16
Document MoltenVK version requirements.


eddyb:
As per https://github.com/KhronosGroup/SPIRV-Cross/pull/1450#issuecomment-746610707, one of the bugs that was hit with MoltenVK so far, should be fixed since MoltenVK v1.1.0 (tho we're still waiting on confirmation).

Also, looking at https://github.com/KhronosGroup/MoltenVK/releases/tag/v1.1.0, it starts off with:
> ### Release for Vulkan 1.1 Support and Vulkan SDK 1.2.154
> * Add support for Vulkan 1.1, including: 

So I'm pretty sure, given that we state Vulkan 1.1+ is required, that we should also say MoltenVK v1.1.0+ is required.

eddyb:
See the edits to https://github.com/KhronosGroup/SPIRV-Cross/pull/1450#issuecomment-746610707 - I sadly was wrong and we still need to deal with the bugs, and they'll only be fixed in a future version.

eddyb:
It ended up being reportedby @VZout as https://github.com/KhronosGroup/SPIRV-Cross/issues/1571, and fixed by https://github.com/KhronosGroup/SPIRV-Cross/pull/1576.
This is included in MoltenVK  v1.1.2 (released a couple days ago), so that's what we should recommend.

VZout:
For completeness: Also this one https://github.com/KhronosGroup/MoltenVK/pull/1204 is required for multiple entry points.

PR #352, opened 2020-12-17
docs: Clarify supported MoltenVK version for macOS

DeanBDean:
Resolves https://github.com/EmbarkStudios/rust-gpu/issues/349

### DIFF
--- a/docs/src/platform-support.md
+++ b/docs/src/platform-support.md
@@ -11,7 +11,7 @@ The `rust-gpu` project currently supports a limited number of platforms and grap
 |-------------------|---------|---------|-------
 | Windows | 10+ | Primary | |
 | Linux | Ubuntu 18.04+ | Primary | |
-| macOS | Catalina (10.15)+ | Secondary | Using [MoltenVK]
+| macOS | Catalina (10.15)+ | Secondary | Using [MoltenVK], requires v1.1.2+
 | Android | Tested 10-11 | Secondary | |
 
 [MoltenVK]: https://github.com/KhronosGroup/MoltenVK


eddyb:
Made this into a draft because it turns out I was wrong about that one bug at least - see the edits to https://github.com/KhronosGroup/SPIRV-Cross/pull/1450#issuecomment-746610707.

docs/src/platform-support.md
```diff
@@ -11,7 +11,7 @@ The `rust-gpu` project currently supports a limited number of platforms and grap
 |-------------------|---------|---------|-------
 | Windows | 10+ | Primary | |
 | Linux | Ubuntu 18.04+ | Primary | |
-| macOS | Catalina (10.15)+ | Secondary | Using [MoltenVK]
+| macOS | Catalina (10.15)+ | Secondary | Using [MoltenVK], requires v1.1.0+
```
eddyb, line 14:
```suggestion
| macOS | Catalina (10.15)+ | Secondary | Using [MoltenVK], requires v1.1.2+
```

as per https://github.com/EmbarkStudios/rust-gpu/issues/349#issuecomment-785804643

eddyb (changes requested):
A new release which contains a fix is now out.

eddyb (approved):

eddyb:
Waiting for confirmation from anyone else (cc @khyperia as well) about whether we want to do this (including potentially adding more explanation etc.)

khyperia:
LGTM as-is, I don't think we need more explanation (linking to bugs etc.) - like, if someone really cared, they'd look at the git blame, find this PR, get the links to the relevant issues, etc., but the message this says of "things may break if below `v1.1.2`" is both accurate and I think sufficient.

XAMPPRocky (approved):

VZout (approved):

PR #359, opened 2020-12-30
Add Const Generic Image Type

XAMPPRocky:
This PR refactors `Image` in the backend to not use attributes to determine things like dimensionality and depth, and to instead accept them as const generics as parameters. This has the advantage of only needing a single `Image` type as opposed previously to requiring a new type for each combination, and also allows us to have enums to represent all of the possibilities rather than using `u32`s directly.

### Additions

- `macros::Image` a new procedural macro for defining `Image` types.
- `image::SampleType` a trait that allows us to be generic of the underlying sampled type of the image.
- `image::ImageCoordinate` a trait built on top of `Vector` for parameterising image coordinate vectors based on the sampled type, dimensionality, and arrayed-ness of the Image.
- A new `spirv_types` crate for holding types that are useful in both `spirv-std` and `spirv-std-macros`.

### Re-works

- `crate/spirv-std-macros` has been moved into the `spirv-std` folder as `macros` beside the `shared` crate, this keeps the folder structure cleaner.

### DIFF
--- a/crates/rustc_codegen_spirv/src/abi.rs
+++ b/crates/rustc_codegen_spirv/src/abi.rs
@@ -10,7 +10,9 @@ use rustc_errors::ErrorReported;
 use rustc_middle::bug;
 use rustc_middle::ty::layout::{FnAbiExt, TyAndLayout};
 use rustc_middle::ty::subst::SubstsRef;
-use rustc_middle::ty::{GeneratorSubsts, PolyFnSig, Ty, TyKind, TypeAndMut};
+use rustc_middle::ty::{
+    Const, FloatTy, GeneratorSubsts, IntTy, ParamEnv, PolyFnSig, Ty, TyKind, TypeAndMut, UintTy,
+};
 use rustc_span::def_id::DefId;
 use rustc_span::Span;
 use rustc_target::abi::call::{CastTarget, FnAbi, PassMode, Reg, RegKind};
@@ -20,6 +22,9 @@ use rustc_target::abi::{
 use std::cell::RefCell;
 use std::collections::hash_map::Entry;
 use std::fmt;
+
+use num_traits::cast::FromPrimitive;
+use rspirv::spirv;
 
 /// If a struct contains a pointer to itself, even indirectly, then doing a naiive recursive walk
 /// of the fields will result in an infinite loop. Because pointers are the only thing that are
@@ -768,6 +773,94 @@ fn trans_intrinsic_type<'tcx>(
             }
             // Hardcode to float for now
             let sampled_type = SpirvType::Float(32).def(span, cx);
+
+            let ty = SpirvType::Image {
+                sampled_type,
+                dim,
+                depth,
+                arrayed,
+                multisampled,
+                sampled,
+                image_format,
+                access_qualifier,
+            };
+
+            Ok(ty.def(span, cx))
+        }
+        IntrinsicType::GenericImageType => {
+            // see SpirvType::sizeof
+            if ty.size != Size::from_bytes(4) {
+                cx.tcx
+                    .sess
+                    .err("#[spirv(generic_image)] type must have size 4");
+                return Err(ErrorReported);
+            }
+
+            fn type_from_variant_discriminant<'tcx, P: FromPrimitive>(
+                cx: &CodegenCx<'tcx>,
+                const_: &'tcx Const<'tcx>,
+            ) -> P {
+                let adt_def = const_.ty.ty_adt_def().unwrap();
+                assert!(adt_def.is_enum());
+                let destructured = cx.tcx.destructure_const(ParamEnv::reveal_all().and(const_));
+                let idx = destructured.variant.unwrap();
+                let value = const_.ty.discriminant_for_variant(cx.tcx, idx).unwrap().val as u64;
+                <_>::from_u64(value).unwrap()
+            }
+
+            let sampled_type = match substs.type_at(0).kind() {
+                TyKind::Int(int) => match int {
+                    IntTy::Isize => {
+                        SpirvType::Integer(cx.tcx.data_layout.pointer_size.bits() as u32, true)
+                            .def(span, cx)
+                    }
+                    IntTy::I8 => SpirvType::Integer(8, true).def(span, cx),
+                    IntTy::I16 => SpirvType::Integer(16, true).def(span, cx),
+                    IntTy::I32 => SpirvType::Integer(32, true).def(span, cx),
+                    IntTy::I64 => SpirvType::Integer(64, true).def(span, cx),
+                    IntTy::I128 => SpirvType::Integer(128, true).def(span, cx),
+                },
+                TyKind::Uint(uint) => match uint {
+                    UintTy::Usize => {
+                        SpirvType::Integer(cx.tcx.data_layout.pointer_size.bits() as u32, false)
+                            .def(span, cx)
+                    }
+                    UintTy::U8 => SpirvType::Integer(8, false).def(span, cx),
+                    UintTy::U16 => SpirvType::Integer(16, false).def(span, cx),
+                    UintTy::U32 => SpirvType::Integer(32, false).def(span, cx),
+                    UintTy::U64 => SpirvType::Integer(64, false).def(span, cx),
+                    UintTy::U128 => SpirvType::Integer(128, false).def(span, cx),
+                },
+                TyKind::Float(FloatTy::F32) => SpirvType::Float(32).def(span, cx),
+                TyKind::Float(FloatTy::F64) => SpirvType::Float(64).def(span, cx),
+                _ => {
+                    cx.tcx
+                        .sess
+                        .span_err(span, "Invalid sampled type to `Image`.");
+                    return Err(ErrorReported);
+                }
+            };
+
+            let dim: spirv::Dim = type_from_variant_discriminant(cx, substs.const_at(1));
+            let depth: u32 = type_from_variant_discriminant(cx, substs.const_at(2));
+            let arrayed: u32 = type_from_variant_discriminant(cx, substs.const_at(3));
+            let multisampled: u32 = type_from_variant_discriminant(cx, substs.const_at(4));
+            let sampled: u32 = type_from_variant_discriminant(cx, substs.const_at(5));
+            let image_format: spirv::ImageFormat =
+                type_from_variant_discriminant(cx, substs.const_at(6));
+
+            let access_qualifier = {
+                let option = cx
+                    .tcx
+                    .destructure_const(ParamEnv::reveal_all().and(substs.const_at(7)));
+
+                match option.variant.map(|i| i.as_u32()).unwrap_or(0) {
+                    0 => None,
+                    1 => Some(type_from_variant_discriminant(cx, option.fields[0])),
+                    _ => unreachable!(),
+                }
+            };
+
             let ty = SpirvType::Image {
                 sampled_type,
                 dim,
--- a/crates/rustc_codegen_spirv/src/symbols.rs
+++ b/crates/rustc_codegen_spirv/src/symbols.rs
@@ -329,7 +329,7 @@ impl Symbols {
             ("flat", SpirvAttribute::Flat),
             ("invariant", SpirvAttribute::Invariant),
             (
-                "generic_image",
+                "generic_image_type",
                 SpirvAttribute::IntrinsicType(IntrinsicType::GenericImageType),
             ),
             (
--- a/crates/rustc_codegen_spirv/src/symbols.rs
+++ b/crates/rustc_codegen_spirv/src/symbols.rs
@@ -329,10 +329,6 @@ impl Symbols {
             ("flat", SpirvAttribute::Flat),
             ("invariant", SpirvAttribute::Invariant),
             (
-                "generic_image_type",
-                SpirvAttribute::IntrinsicType(IntrinsicType::GenericImageType),
-            ),
-            (
                 "sampled_image",
                 SpirvAttribute::IntrinsicType(IntrinsicType::SampledImage),
             ),
--- a/crates/rustc_codegen_spirv/src/symbols.rs
+++ b/crates/rustc_codegen_spirv/src/symbols.rs
@@ -330,7 +330,7 @@ impl Symbols {
             ("invariant", SpirvAttribute::Invariant),
             (
                 "generic_image",
-                SpirvAttribute::IntrinsicType(IntrinsicType::SampledImage),
+                SpirvAttribute::IntrinsicType(IntrinsicType::GenericImageType),
             ),
             (
                 "sampled_image",
--- a/crates/rustc_codegen_spirv/src/symbols.rs
+++ b/crates/rustc_codegen_spirv/src/symbols.rs
@@ -24,6 +24,7 @@ pub struct Symbols {
     descriptor_set: Symbol,
     binding: Symbol,
     image_type: Symbol,
+    generic_image_type: Symbol,
     dim: Symbol,
     depth: Symbol,
     arrayed: Symbol,
@@ -366,6 +367,7 @@ impl Symbols {
             descriptor_set: Symbol::intern("descriptor_set"),
             binding: Symbol::intern("binding"),
             image_type: Symbol::intern("image_type"),
+            generic_image_type: Symbol::intern("generic_image_type"),
             dim: Symbol::intern("dim"),
             depth: Symbol::intern("depth"),
             arrayed: Symbol::intern("arrayed"),
@@ -430,7 +432,9 @@ pub(crate) fn parse_attrs_for_checking<'a>(
             .into_iter()
             .chain(args.into_iter().map(move |ref arg| {
                 let span = arg.span();
-                let parsed_attr = if arg.has_name(sym.image_type) {
+                let parsed_attr = if arg.has_name(sym.generic_image_type) {
+                    SpirvAttribute::IntrinsicType(IntrinsicType::GenericImageType)
+                } else if arg.has_name(sym.image_type) {
                     parse_image_type(sym, arg)?
                 } else if arg.has_name(sym.descriptor_set) {
                     SpirvAttribute::DescriptorSet(parse_attr_int_value(arg)?)
--- a/crates/rustc_codegen_spirv/src/symbols.rs
+++ b/crates/rustc_codegen_spirv/src/symbols.rs
@@ -24,7 +24,6 @@ pub struct Symbols {
     descriptor_set: Symbol,
     binding: Symbol,
     image_type: Symbol,
-    generic_image_type: Symbol,
     dim: Symbol,
     depth: Symbol,
     arrayed: Symbol,
@@ -319,12 +318,20 @@ impl Symbols {
                 SpirvAttribute::IntrinsicType(IntrinsicType::Sampler),
             ),
             (
+                "generic_image_type",
+                SpirvAttribute::IntrinsicType(IntrinsicType::GenericImageType),
+            ),
+            (
                 "acceleration_structure",
                 SpirvAttribute::IntrinsicType(IntrinsicType::AccelerationStructureKhr),
             ),
             ("block", SpirvAttribute::Block),
             ("flat", SpirvAttribute::Flat),
             ("invariant", SpirvAttribute::Invariant),
+            (
+                "generic_image",
+                SpirvAttribute::IntrinsicType(IntrinsicType::SampledImage),
+            ),
             (
                 "sampled_image",
                 SpirvAttribute::IntrinsicType(IntrinsicType::SampledImage),
@@ -367,7 +374,6 @@ impl Symbols {
             descriptor_set: Symbol::intern("descriptor_set"),
             binding: Symbol::intern("binding"),
             image_type: Symbol::intern("image_type"),
-            generic_image_type: Symbol::intern("generic_image_type"),
             dim: Symbol::intern("dim"),
             depth: Symbol::intern("depth"),
             arrayed: Symbol::intern("arrayed"),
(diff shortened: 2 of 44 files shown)


expenses:
Awesome! I'm glad validation catches things like sampling with the wrong coordinate type:
```rust
error: invalid data:0:0 - Expected Coordinate to have at least 3 components, but given only 1
  %85 = OpImageSampleImplicitLod %v4float %84 %77

  |
  = note: spirv-val failed
```

I think that type aliases would be really useful though, e.g. 
```rust
pub type ImageCube = Image<
    { Dimensionality::Cube },
    { ImageDepth::No },
    false,
    false,
    { Sampled::Yes },
    { ImageFormat::Unknown },
    { None },
>;
```
or possibly 
```rust
pub type ImageCube<
    const FORMAT: ImageFormat,
> = Image<
    { Dimensionality::Cube },
    { ImageDepth::No },
    false,
    false,
    { Sampled::Yes },
    { FORMAT },
    { None },
>;
```

if there's a performance gain with setting the image format.

XAMPPRocky:
> Awesome! I'm glad validation catches things like sampling with the wrong coordinate type:

Yeah, we should probably see if we can catch that at a higher level so that you see the `sample` call from the Rust code rather than the SPIR-V asm. 

> I think that type aliases would be really useful though

Sure, I don't really see the harm in that, though that's really just working around the current limitations of the lack of defaults for const generics. Once defaults are implemented we could set them on `Image` itself so that most users could just say `Image<{ Dimensionality::Cube }>` in most cases, and then if users need a more complex combination they can set their own type aliases.

crates/spirv-std/src/textures.rs
```diff
@@ -1,10 +1,81 @@
...
+    Rg8ui = 37,
+    R16ui = 38,
+    R8ui = 39,
+}
+
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub enum AccessQualifier {
+    ReadOnly = 0,
+    WriteOnly = 1,
+    ReadWrite = 2,
```
eddyb, line 78: Rust guarantees sequential values starting from `0`, not sure why they were explicitly provided here.

crates/spirv-std/src/textures.rs
```diff
@@ -52,58 +150,41 @@ impl Image2d {
...
+            { DEPTH },
+            { ARRAYED },
+            { MULTISAMPLED },
+            { SAMPLED },
+            { FORMAT },
+            { ACCESS_QUALIFIER },
+        >,
+    >
+{
+    pub fn sample(&self, coord: impl ScalarOrVector) -> Vec4 {
```
eddyb, line 187: Shouldn't the coordinate type depend on the dimensionality of the image?

You can replace `trait ScalarOrVector` with a `trait ImageCoord<const DIM: Dimensionality>` you can implement for specific dimensionalities, though it might make even more sense to extract a single coordinate type, instead.

Though with a `ImageCoord` trait, we can support both `Vec3` and `Vec3A`, by converting the former to the latter.

crates/spirv-std/src/textures.rs
```diff
@@ -18,18 +89,45 @@ pub struct Sampler {
...
+        const ACCESS_QUALIFIER: Option<AccessQualifier>,
+    >
+    Image<
+        { DIM },
+        { DEPTH },
+        { ARRAYED },
+        { MULTISAMPLED },
+        { SAMPLED },
+        { FORMAT },
+        { ACCESS_QUALIFIER },
```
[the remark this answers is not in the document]
XAMPPRocky: They're currently required for ADT types AFAICT. https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=47f4ed26b3f3965e164d075aede833fa

crates/spirv-std/src/textures.rs
XAMPPRocky: > Shouldn't the coordinate type depend on the dimensionality of the image?

Yes it should, just wasn't sure how to ensure that. That idea sounds great and I'll add it.

crates/spirv-std/src/textures.rs
XAMPPRocky: Well these are also the values of their SPIR-V assembly, and including them would allow us to re-order them and not worry about it affecting code, where as implicit sequential values relies on us, rustfmt, or a contributor remembering to not re-order them.

examples/shaders/sky-shader/src/lib.rs
```diff
@@ -159,14 +160,29 @@ pub fn fs(constants: &ShaderConstants, frag_coord: Vec2) -> Vec4 {
...
+    #[spirv(binding = 0)] img: UniformConstant<
+        Image<
+            { Dimensionality::TwoD },
+            { ImageDepth::No },
+            false,
+            false,
+            { Sampled::Yes },
+            { ImageFormat::Unknown },
+            { None },
+        >,
```
Jasper-Bekkers, line 174: I think we can't really have user facing code look like this, it's way to verbose so at a minimum we should have some typedefs for this.

Jasper-Bekkers (dismissed):

crates/spirv-std/src/textures.rs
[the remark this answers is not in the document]
eddyb: Name resolution doesn't care about types. These are generic parameter names, so they should work just fine unbraced.

rust-toolchain
```diff
@@ -5,5 +5,5 @@
 # to the user in the error, instead of "error: invalid channel name '[toolchain]'".
 
 [toolchain]
-channel = "nightly-2020-12-11"
+channel = "nightly-2020-12-28"
```
eddyb, line 8: I guess you can rebase on `main` to get rid of this.

crates/spirv-std/src/textures.rs
```diff
@@ -1,5 +1,37 @@
...
+            { Sampled::Unknown },
+            { ImageFormat::Unknown },
+            { None }
+            >;
+        )+
+    }
+}
+
+basic_image_type! {
+    /// A convenience type alias for a one dimensionsal image.
```
Jasper-Bekkers, line 21: Nit: small typo in dimensional (dimensionsal), happens a few more times below.

examples/shaders/sky-shader/src/lib.rs
Jasper-Bekkers: Thanks for adding the type aliasses @XAMPPRocky, we should use them here as well, code samples get copy & pasted far and wide.

XAMPPRocky: Yeah, I'll update it, I have to make the example work first. 😉

XAMPPRocky:
This is ready for review, one of the main things to consider before merging is that this will require nightly for `spirv-std` everywhere where as right now you can technically compile `spirv-std` on stable on host platforms.

khyperia:
> one of the main things to consider before merging is that this will require nightly for `spirv-std` everywhere

Pinging @VZout on this topic, since they did a bunch of work trying to get `spirv-std` building on stable, so I want them to sign off on this.

VZout:
> > one of the main things to consider before merging is that this will require nightly for `spirv-std` everywhere
> 
> Pinging @VZout on this topic, since they did a bunch of work trying to get `spirv-std` building on stable, so I want them to sign off on this.

I'm not the person to ok this. But this is small step back for ark. While it is a good change of course I personally think there is not enough reason to merge this. I prefer to leave the shaders intertwined with ark on stable. Especially if we decide to go with trait based rendering this becomes a big problem. 

Can we hold of on this decision once we have had the more technical meeting with Henrik on the rendering clean up and whether we continue with trait based stuff?

XAMPPRocky:
I've also updated this PR so this is no longer the case, it just means now that `Image` is only available when in a shader crates, which should be already be case.

khyperia:
> it just means now that `Image` is only available when in a shader crate, which should be already be case.

No, this was an explicit decision and intentional design made to make Image available on the CPU. This is because not having it available on the CPU significantly degrades editor experience out of the box, something that's important to us.

image

If you want to override that decision, that's up to you, I just want to make sure we're aware of the decisions we're making instead of flipping back and forth on some things repeatedly without being aware of history (something that's happened in the past).

XAMPPRocky:
> This is because not having it available on the CPU significantly degrades editor experience out of the box

This is a good point, it also means docs are wouldn't be available by default. You could enable both of those back by manually adding a feature flag for example and customising your rust-analyser config, but I don't think that's really ideal.

Fundamentally the stable restriction will require one or more trade-offs in the implementation. I've listed all possibilities that I could think of below and their tradeoffs below.


- Exposing `Image` with enum const generics in the API.
	- **Advantages**
		- Users can't create invalid image types in the type system.
		- We can specialise types in arguments based on these const generics (See `ImageCoord`)
	- **Disadvantages**
	    	- `spirv-std` no longer compiles on stable.
- Keeping `Image` with enum const generics but hidden behind `cfg()`.
	- **Advantages**
		- Same as all of the above.
		- Compiles on stable
	- **Disadvantages**
		- Editor and docs experience worse out of the gate.
- Using type level generics instead of const generics. (E.g. like `num`)
	- **Advantages**
		- Compiles on stable
		- Same benefits as const generics
	- **Disadvantages**
		- A lot more boilerplate to achieve the same goal on both the user and the backend.
- Using only primitives (`u8`, `bool`) const generics.
	- **Advantages**
		- Will be compilable on stable soon.
	- **Disadvantages**
		- Users can easily enter the wrong value
		- The generics become more opaque as to what they're doing.
			-  we can provide `consts` in a module to make it easier.
			- Still doesn't stop someone from writing or seeing `Image<1, 0, false, false, 0, 0, 0>`.
- Continuing with attributes.
	- **Advantages**
		- Compilable on stable.
	- **Disadvantages**
		- Requires a combinatorial explosion of the number types for covering each possible combination.

XAMPPRocky:
Actually there's one other possibility that just came to mind, which is using the `RUSTC_BOOTSTRAP` flag in Ark when building the shaders, so that it can use the nightly features while still being on stable.

**Advantages**
- Everything works.

**Disadvantage**
- Opens some ark to instability.

khyperia:
Removing myself from reviewers until const generics are stabilized and this can land, because I use https://github.com/pulls/review-requested a lot as my to-do list, and a glaring "review this thing!!" makes me distracted and less likely to review the things I'm supposed to be looking at :s

XAMPPRocky:
this will be able to land sooner since it's under a flag, but I still need to do a bit of polish work to get it ready for review so I'll move it back to the draft stage for now.

crates/spirv-std/shared/src/image_params.rs
```diff
@@ -0,0 +1,424 @@
...
+            "rgba16ui" => Self::Rgba16ui,
+            "rgba8ui" => Self::Rgba8ui,
+            "r32ui" => Self::R32ui,
+            "rgb10_a2ui" => Self::Rgb10A2ui,
+            "rg32ui" => Self::Rg32ui,
+            "rg16ui" => Self::Rg16ui,
+            "rg8ui" => Self::Rg8ui,
+            "r16ui" => Self::R16ui,
+            "r8ui" => Self::R8ui,
+            _ => return Err("Unknown specified image format. Use `unknown(<type>)` if this is intentional."),
```
XAMPPRocky, line 363:
```suggestion
            _ => return Err("Unknown specified image format. Use `type=<type>` instead if this is intentional."),
```

XAMPPRocky:
Actually, need to figure out some more stuff before it's ready for review.

eddyb:
The overall approach seems good to me, but I'm not so sure about the crate organization (and the new crate).

It might even make sense to generate some of the bigger `enum`s from e.g. [`spirv.core.grammar.json`](https://github.com/KhronosGroup/SPIRV-Headers/blob/master/include/spirv/unified1/spirv.core.grammar.json), instead of having them laid out by hand.

I'm deferring to @khyperia on the issue of the (`macros` + `shared`) crate organization, I don't want to block this (long-awaited) PR if she's fine with it.

XAMPPRocky:
> It might even make sense to generate some of the bigger enums from e.g. spirv.core.grammar.json, instead of having them laid out by hand.

I originally did it by hand because the definitions are already generated in the `spirv_headers` crate, but that's didn't used to work when I first tired to use it (back in January), but now it seems like it compiles just fine, so I can probably use them instead.

crates/rustc_codegen_spirv/src/abi.rs
```diff
@@ -768,6 +773,92 @@ fn trans_intrinsic_type<'tcx>(
...
+                    UintTy::U8 => SpirvType::Integer(8, false).def(span, cx),
+                    UintTy::U16 => SpirvType::Integer(16, false).def(span, cx),
+                    UintTy::U32 => SpirvType::Integer(32, false).def(span, cx),
+                    UintTy::U64 => SpirvType::Integer(64, false).def(span, cx),
+                    UintTy::U128 => SpirvType::Integer(128, false).def(span, cx),
+                },
+                TyKind::Float(FloatTy::F32) => SpirvType::Float(32).def(span, cx),
+                TyKind::Float(FloatTy::F64) => SpirvType::Float(64).def(span, cx),
+                _ => {
+                    cx.tcx.sess.err("Invalid sampled type to `Image`.");
```
khyperia, line 837: I think passing in the span would be good here?

crates/spirv-builder/src/lib.rs
```diff
@@ -204,7 +204,7 @@ fn invoke_rustc(builder: &SpirvBuilder, multimodule: bool) -> Result<PathBuf, Sp
         .unwrap_or_default();
 
     let rustflags = format!(
-        "-Z codegen-backend={} -Z symbol-mangling-version=v0{}",
+        "-Z codegen-backend={} {}",
```
khyperia, line 207: Wait, why was this done? I probably missed a discussion around this, but I couldn't find anything in the review comments here.

crates/spirv-std/shared/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "spirv-types"
+version = "0.4.0-alpha.3"
+authors = ["Erin Power <erin.power@embark-studios.com>"]
```
khyperia, line 4: Nit: `Embark <opensource@embark-studios.com>`. Also `version = "0.4.0-alpha.5"`, and `repository = "https://github.com/EmbarkStudios/rust-gpu"`

crates/spirv-std/src/image.rs
```diff
@@ -0,0 +1,713 @@
...
+    {
+        unsafe {
+            let mut result = Default::default();
+            asm!(
+                "%typeSampledImage = OpTypeSampledImage typeof*{1}",
+                "%image = OpLoad typeof*{1} {1}",
+                "%sampler = OpLoad typeof*{2} {2}",
+                "%coord = OpLoad typeof*{3} {3}",
+                "%sampledImage = OpSampledImage %typeSampledImage %image %sampler",
+                "%result = OpImageSampleImplicitLod typeof*{0} %sampledImage %coord",
```
khyperia, line 65: From the spec:

> OpImageSampleImplicitLod
> The MS operand of the underlying OpTypeImage must be 0.

Might be nice to encode that, as well as all the other restrictions here, but that can be done in a follow-up.

khyperia (commented):
Looks roughly good to me! The extra crate is a little annoying, but it works, so sure I guess. I think I might prefer nuking the old image system at the same time, having duplication for even a short amount of time is a big maintenance burden and causes confusion.

crates/spirv-builder/src/lib.rs
XAMPPRocky: So, it was required for a previous iteration of the PR, however my most recent changes removed the parts of `Image` that required the newer symbol mangling version, so this is no longer needed.

XAMPPRocky:
Alright, I've implemented all of the feedback.

> I think I might prefer nuking the old image system at the same time, having duplication for even a short amount of time is a big maintenance burden and causes confusion.

Well if we nuke the old system, we can't upgrade Ark until C-like enums are stabilised or we change the parts of Ark to not use reflection for entry points. So I do think it's worth keeping them both for a short time, I've also added a deprecated message if you have the const generics feature enabled, so if you can use it, it warns to move to the new type.

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -430,7 +432,9 @@ pub(crate) fn parse_attrs_for_checking<'a>(
             .into_iter()
             .chain(args.into_iter().map(move |ref arg| {
                 let span = arg.span();
-                let parsed_attr = if arg.has_name(sym.image_type) {
+                let parsed_attr = if arg.has_name(sym.generic_image_type) {
+                    SpirvAttribute::IntrinsicType(IntrinsicType::GenericImageType)
```
khyperia, line 436: Is there a reason this isn't in the `custom_attributes` array? Seems like that's the proper place for it.

crates/spirv-builder/src/lib.rs
```diff
@@ -204,7 +204,7 @@ fn invoke_rustc(builder: &SpirvBuilder, multimodule: bool) -> Result<PathBuf, Sp
         .unwrap_or_default();
 
     let rustflags = format!(
-        "-Z codegen-backend={} -Z symbol-mangling-version=v0{}",
+        "-Z codegen-backend={} -Zsymbol-mangling-version=v0 {}",
```
khyperia, line 207: Nit: This will cause a double space if `llvm_args` has a value, and a trailing space if `llvm_args` is empty. How it was before was correct.

tests/ui/image/sample_depth_reference/sample.rs
```diff
@@ -1,19 +1,20 @@
...
     #[spirv(descriptor_set = 3, binding = 3)] sampler: &Sampler,
     output: &mut f32,
 ) {
     let v2 = glam::Vec2::new(0.0, 1.0);
     let v3 = glam::Vec3A::new(0.0, 0.0, 1.0);
     *output = image.sample_depth_reference(*sampler, v2, 1.0);
     *output += image_array.sample_depth_reference(*sampler, v3, 1.0);
-    *output += cubemap.sample_depth_reference(*sampler, v3, 1.0);
+    let sample: f32 = cubemap.sample_depth_reference(*sampler, v3, 1.0);
+    *output += sample;
```
khyperia, line 19: Nit: Is there a reason for this change?

khyperia (approved):
LGTM pending nits.

Just realized, do we want to export common type aliases? Say, for all the types we already have explicitly defined, e.g. `pub type Image2D = Image!(2d, type=f32, sampled)` - might be nice for both examples for users of how to use the macro, plus saves users having to figure out e.g. what the heck `sampled` means and if they want it or not.

khyperia:
Oh yeah, also adding docs to https://embarkstudios.github.io/rust-gpu/book/attributes.html

XAMPPRocky:
> Oh yeah, also adding docs to the rust-gpu book on the macro syntax would be good, but that can be done in a follow-up if you'd like (file an issue if so).

The syntax is documented on the macro itself so when you visit the `Image!` API docs you'll see it there.

khyperia:
> The syntax is documented on the macro itself so when you visit the `Image!` API docs you'll see it there.

People look at the book instead of the API docs when trying to figure out language constructs (where language = DSL in this case), e.g. `u32` is documented both here https://doc.rust-lang.org/stable/std/primitive.u32.html and here https://doc.rust-lang.org/book/ch03-02-data-types.html . I think it'd make rust-gpu more accessible and easier to use if documentation was available in both places - something as core to graphics programming as images should be documented in the rust-gpu book, especially if it's not "normal" rust types.

crates/rustc_codegen_spirv/src/symbols.rs
XAMPPRocky: Well it's not how the `Image` was implemented, and this was just adapting from how that was implemented.

khyperia: That's because `image` has arguments, `generic_image` does not - `symbols.rs` is organized by syntax, not by semantics.

crates/spirv-std/src/lib.rs
```diff
@@ -68,22 +68,28 @@
...
+#[cfg(feature = "const-generics")]
+pub mod image;
 pub mod integer;
 pub mod memory;
 pub mod ray_tracing;
+mod sampler;
 pub mod scalar;
 pub(crate) mod sealed;
-mod textures;
+pub mod textures;
```
XAMPPRocky, line 88:
```suggestion
mod textures;
```

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -317,13 +317,21 @@ impl Symbols {
...
             (
                 "acceleration_structure",
                 SpirvAttribute::IntrinsicType(IntrinsicType::AccelerationStructureKhr),
             ),
             ("block", SpirvAttribute::Block),
             ("flat", SpirvAttribute::Flat),
             ("invariant", SpirvAttribute::Invariant),
+            (
+                "generic_image",
+                SpirvAttribute::IntrinsicType(IntrinsicType::SampledImage),
```
khyperia, line 333: huh?

XAMPPRocky, line 333:
```suggestion
                SpirvAttribute::IntrinsicType(IntrinsicType::GenericImageType),
```

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -317,13 +317,21 @@ impl Symbols {
...
+            ),
             (
                 "acceleration_structure",
                 SpirvAttribute::IntrinsicType(IntrinsicType::AccelerationStructureKhr),
             ),
             ("block", SpirvAttribute::Block),
             ("flat", SpirvAttribute::Flat),
             ("invariant", SpirvAttribute::Invariant),
+            (
+                "generic_image",
```
XAMPPRocky, line 332:
```suggestion
                "generic_image_type",
```

khyperia: wait, now there are two of them?

PR #366, opened 2021-01-07
Document Derivative

XAMPPRocky:


crates/spirv-std/src/derivative.rs
```diff
@@ -0,0 +1,98 @@
...
+        }
+    };
+}
+
+// "must be a scalar or vector of floating-point type. The component width must be 32 bits."
+deriv_impl!(f32);
+// TODO: Fix rustc to support these
+// deriv_impl!(glam::Vec2);
+// deriv_impl!(glam::Vec3);
+// deriv_impl!(glam::Vec4);
```
khyperia, line 98: This is offtopic on this PR, but IIRC @eddyb got these working via pointers and `typeof*` in their repo, might be nice to port those back here.

crates/spirv-std/src/derivative.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// the current fragment and its immediate neighbor(s).
+    fn ddy_fine(self) -> Self;
+    /// Returns the partial derivative of `Self` with respect to the window's
+    /// Y coordinate. Uses local differencing based on the value of `Self` for
+    /// the current fragment’s neighbors, and possibly, but not necessarily,
+    /// includes the value of `Self` for the current fragment. That is, over a
+    /// given area, the implementation can compute Y derivatives in fewer
+    /// unique locations than would be allowed by [`Derivative::ddy_fine`].
+    fn ddy_coarse(self) -> Self;
+    /// Returns the sum of [`Self::ddx`] and [`Self::ddy`] as a
```
khyperia, line 38: nit: this is `abs(ddx())+abs(ddy())`, not `ddx()+ddy()` (I would imagine most people reading these docs would already know that, but idk)

khyperia (approved):

crates/spirv-std/src/derivative.rs
XAMPPRocky: > (I would imagine most people reading these docs would already know that, but idk)

No you're right, it should mention that. I just glossed over the words *absolute values* in the spec. Best to make no assumptions about the readers.

eddyb (approved):

Issue #369, opened 2021-01-08
Unable to use image operands in asm!


msiglreith:
### Expected Behaviour
rust-gpu allows writing asm code including image operands

Example
```
%result = OpImageSampleExplicitLod typeof*{0} %sampledImage %coord Lod %lod"
```

This results in an error currently regarding too many operands as right now only the `Lod` image operand will be parsed. `rspirv` doesn't consume the following IdRefs.

Requires changes to rspirv: https://github.com/gfx-rs/rspirv/pull/178

msiglreith:
Fixed in https://github.com/EmbarkStudios/rust-gpu/pull/418

PR #372, opened 2021-01-12
Fix image operands

msiglreith:
Fixes https://github.com/EmbarkStudios/rust-gpu/issues/369

Updates `rspirv` to latest main which includes fix for handling of the image operands bits.
Image operands require the following IdRefs to be parsed as well depending on the numbers of bits in the bitmask.

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
+++ b/crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
@@ -953,7 +953,7 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
                 match parse_bitflags_operand(IMAGE_OPERANDS, word) {
                     Some(x) => {
                         inst.operands.push(dr::Operand::ImageOperands(x));
-                        while let Some(token) = tokens.next() {
+                        for token in tokens {
                             match self.parse_id_in(id_map, token) {
                                 Some(id) => {
                                     inst.operands.push(dr::Operand::IdRef(id));
--- a/crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
+++ b/crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
@@ -951,7 +951,23 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
 
             (OperandKind::ImageOperands, Some(word)) => {
                 match parse_bitflags_operand(IMAGE_OPERANDS, word) {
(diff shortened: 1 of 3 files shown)


crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
```diff
@@ -801,7 +801,17 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
 
             (OperandKind::ImageOperands, Some(word)) => {
                 match parse_bitflags_operand(IMAGE_OPERANDS, word) {
-                    Some(x) => inst.operands.push(dr::Operand::ImageOperands(x)),
+                    Some(x) => {
+                        inst.operands.push(dr::Operand::ImageOperands(x));
+                        for _ in 0..x.bits().count_ones() {
```
khyperia, line 806: Is this correct? For example, the `Grad` image operand is a single bit (0x4) but it takes two operands, and there are also a handful that have zero operands.

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
```diff
@@ -801,7 +801,17 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
...
+                            match tokens.next().and_then(|token| self.parse_id_in(id_map, token)) {
+                                Some(id) => {
+                                    inst.operands.push(dr::Operand::IdRef(id));
+                                }
+                                None => self.err("expected operand id after image operand"),
```
khyperia, line 811: We probably only want to only report this once, since there's no location attached.

khyperia (changes requested):
Thanks so much for this!

edit: haha wow, I'm apparently _really_ tired (just woke up) and thought the "add individual comment" button was the "start review" button. sorry for the spam.

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
msiglreith: good catch! I need to adjust `rspirv` as well.

msiglreith:
Should be ready now!

khyperia (changes requested):
I'm confused, it still seems to be counting the number of ones in the bitenum and parsing that number of arguments?

msiglreith:
oops sry missed that one

(I'm trying in the meanwhile to autogenerate more parts of the parsing in rspirv but will take a while)

msiglreith:
Closing in favor of #418

PR #376, opened 2021-01-14
 inline asm!: support writing _ in lieu of return types, for basic inference.

eddyb:
The bulk of this PR is the first commit, containing a basic type constraint system for SPIR-V instructions, which I've been working on as a prerequisite for #300 (see module-scope doc comment for its limitations).

What I've done with inline `asm!` is demonstrate the use of those constraints to alleviate the need to write some `typeof*{...}`, letting the user write just `_` instead, and having the type be inferred automatically.

However, `asm!` inference is merely a potential usecase I thought of while working on the type constraint system, and I don't have a strong opinion on keeping it (this is partly the reason why I haven't made it fully-featured yet).


khyperia (approved):
:tada:

PR #380, opened 2021-01-18
discard(), update rspirv, better capability computation

khyperia:
We could bikeshed on the names of `discard()` and `demote_to_helper_invocation()`, but I hope with the doc comments it's pretty clear to users regardless of the exact names.

Capability computation was apparently borked, as we were including e.g. OpCapability DerivativeControl in our shaders even without calling the derivative functions. We should probably rethink this area, the current system is still a "temporary" hack. I cleaned up the hack a little bit, but it's still pretty gross and weird.

### DIFF
--- a/crates/rustc_codegen_spirv/Cargo.toml
+++ b/crates/rustc_codegen_spirv/Cargo.toml
@@ -29,7 +29,7 @@ use-compiled-tools = ["spirv-tools/use-compiled-tools"]
 [dependencies]
 bimap = "0.6"
 indexmap = "1.6.0"
-rspirv = { git = "https://github.com/gfx-rs/rspirv.git", rev = "01ca0d2e5b667a0e4ff1bc1804511e38f9a08759" }
+rspirv = { git = "https://github.com/gfx-rs/rspirv.git", rev = "7111e6c5a8bb43563d280825fa65623032ee052d" }
 rustc-demangle = "0.1.18"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
--- a/crates/spirv-std/src/lib.rs
+++ b/crates/spirv-std/src/lib.rs
@@ -51,6 +51,26 @@ pub use glam;
 pub use num_traits;
 pub use textures::*;
 
+/// Calls the `OpDemoteToHelperInvocationEXT` instruction, which corresponds to discard() in HLSL
+pub fn demote_to_helper_invocation() {
+    #[cfg(target_arch = "spirv")]
+    unsafe {
+        asm!(
+            "OpExtension \"SPV_EXT_demote_to_helper_invocation\"",
+            "OpCapability DemoteToHelperInvocationEXT",
+            "OpDemoteToHelperInvocationEXT"
+        );
+    }
+}
+
+/// Calls the `OpKill` instruction, which corresponds to discard() in GLSL
+pub fn discard() {
+    #[cfg(target_arch = "spirv")]
+    unsafe {
+        asm!("OpKill", "%unused = OpLabel");
+    }
+}
+
 #[cfg(all(not(test), target_arch = "spirv"))]
 #[panic_handler]
 fn panic(_: &core::panic::PanicInfo<'_>) -> ! {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -235,9 +235,9 @@ checksum = "370f715b81112975b1b69db93e0b56ea4cd4e5002ac43b2da8474106a54096a1"
 dependencies = [
  "heck",
  "proc-macro-error",
- "proc-macro2 1.0.24",
- "quote 1.0.8",
- "syn 1.0.58",
+ "proc-macro2",
+ "quote",
+ "syn",
 ]
 
 [[package]]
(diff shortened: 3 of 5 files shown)


khyperia:
and to answer questions about "why not just default to demote to helper invocation always, not even support OpKill", I get this error message on my modern desktop GPU with up-to-date drivers:

> Shader requires VkPhysicalDeviceShaderDemoteToHelperInvocationFeaturesEXT::shaderDemoteToHelperInvocation but is not enabled on the device

crates/spirv-std/src/lib.rs
```diff
@@ -51,6 +51,26 @@ pub use glam;
...
+        asm!(
+            "OpExtension \"SPV_EXT_demote_to_helper_invocation\"",
+            "OpCapability DemoteToHelperInvocationEXT",
+            "OpDemoteToHelperInvocationEXT"
+        );
+    }
+}
+
+/// Calls the `OpKill` instruction, which corresponds to discard() in GLSL
+pub fn discard() {
```
XAMPPRocky, line 67: To be consistent with the other function, should we have this one called `kill` so that it matches the convention of name of instruction except `Op`.
```suggestion
pub fn kill() {
```

khyperia: I want to match user's expectations of having a function/intrinsic called `discard()`. Not having it seems problematic, and causes unnecessary friction - blindly copying the names of SPIR-V things is something we'd like to avoid (@Jasper-Bekkers believes this much stronger than I, haha). I chose to default the `discard()` implementation to the GLSL one instead of the HLSL one because it's supported by default instead of requiring an extension (see error message).

khyperia: Something we could do is have a
```rust
mod glsl { fn discard(); }
mod hlsl { fn discard(); }
fn discard() { glsl::discard(); }
```
not sure if that's better/worse

XAMPPRocky: > blindly copying the names of SPIR-V things is something we'd like to avoid

I mean this is ostensibly this is `spirv-std` so I think matching their naming makes more sense rather than having a mix of semantics from different specs and languages.

If discoverability is a concern, we can add a doc alias so that when users search for `discard` they find this function. E.g.

```rust
#[doc(alias = "discard")]
fn kill() {}
```

crates/rustc_codegen_spirv/Cargo.toml
```diff
@@ -29,7 +29,7 @@ use-compiled-tools = ["spirv-tools/use-compiled-tools"]
 [dependencies]
 bimap = "0.6"
 indexmap = "1.6.0"
-rspirv = { git = "https://github.com/gfx-rs/rspirv.git", rev = "01ca0d2e5b667a0e4ff1bc1804511e38f9a08759" }
+rspirv = { git = "https://github.com/gfx-rs/rspirv.git", rev = "7111e6c5a8bb43563d280825fa65623032ee052d" }
```
eddyb, line 32: Nice, looks like this resulted in some deduplication/version convergence in `Cargo.lock`.

crates/spirv-std/src/lib.rs
```diff
@@ -51,6 +51,26 @@ pub use glam;
...
+    #[cfg(target_arch = "spirv")]
+    unsafe {
+        asm!("OpKill", "%unused = OpLabel");
```
eddyb, line 70: Hah, I didn't think of this when we needed `discard()` for a shadertoy - I'm guessing this just bypasses `rspirv::dr`'s representation of blocks but still results in correct SPIR-V output? I wonder if it has a risk of breaking some of the `linker` algorithms, tho that won't happen if it gets serialized and reparsed.

crates/spirv-std/src/lib.rs
eddyb, line 67: Besides the naming discussion, there's two things I want to add here (tho neither should block the PR):
* would `fn discard() -> !` make sense? i.e. is execution guaranteed to end, *unlike* `demote_to_helper_invocation`?
* long-term we probably want to have some kind of token to avoid being able to call this outside of a fragment shader, right? (if `demote_to_helper_invocation` also needs that, it could be a `demote_to_helper` method on a `ShaderInvocation` token, which I find nice naming-wise, but that's just bikeshedding)

crates/spirv-std/src/lib.rs
```diff
@@ -51,6 +51,26 @@ pub use glam;
...
 pub use textures::*;
 
+/// Calls the `OpDemoteToHelperInvocationEXT` instruction, which corresponds to discard() in HLSL
+pub fn demote_to_helper_invocation() {
+    #[cfg(target_arch = "spirv")]
+    unsafe {
+        asm!(
+            "OpExtension \"SPV_EXT_demote_to_helper_invocation\"",
+            "OpCapability DemoteToHelperInvocationEXT",
+            "OpDemoteToHelperInvocationEXT"
```
eddyb, line 61: Oh, right, the other reason to have a token is because we could support it in the CPU runner if it just had a `&AtomicBool` (or just `&Cell<bool>`) indicating it's a helper invocation and the output shouldn't be written.

Far less sure about what to do with `OpKill` tho, short of a `panic!` with a special type that can be caught and acted upon. Anyway that's just more tangents.

eddyb (approved):
Left some comments for posterity (feel free to file some issues about some of these things if you want to) but I don't have anything against landing the PR as-is.

crates/spirv-std/src/lib.rs
khyperia: haha yeah it's extremely cursed. only reason it works is because we serialize to object files and re-parse in the linker.

PR #392, opened 2021-01-22
Add gpu_only proc macro

XAMPPRocky:
This PR adds a `#[gpu_only]` macro to `spirv-std-macros`, which essentially expands to the following. Which we were doing a lot in the code already, and will only add more in the future.

```rust
#[gpu_only]
fn foo() {
    // SPIR-V fn body.
}

// Generates

fn foo() {
     #[cfg(target_arch = "spriv")] {
        // SPIR-V fn body.
     }
     
     #[cfg(not(target_arch = "spriv"))] {
         unimplemented!()
     }
}

### DIFF
--- a/crates/spirv-std-macros/src/lib.rs
+++ b/crates/spirv-std-macros/src/lib.rs
@@ -29,3 +29,28 @@ pub fn spirv(_attr: TokenStream, item: TokenStream) -> TokenStream {
     }
     tokens.into_iter().collect()
 }
+
+#[proc_macro_attribute]
+pub fn gpu_only(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let syn::ItemFn {
+        attrs,
+        vis,
+        sig,
+        block,
+    } = syn::parse_macro_input!(item as syn::ItemFn);
+
+    let fn_name = sig.ident.clone();
+
+    let output = quote::quote! {
+        // Don't warn on unused arguments on the CPU side.
+        #[cfg_attr(not(target_arch = "spirv"), allow(unused_variables))]
+        #(#attrs)* #vis #sig {
+            #[cfg(target_arch="spirv")] { #block }
+            #[cfg(not(target_arch="spirv"))] {
+                unimplemented!(concat!("`", stringify!(#fn_name), "` is only available on SPIR-V platforms."))
+            }
+        }
+    };
+
+    output.into()
+}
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: crates/spirv-std/src/textures.rs [hunk 1]
```diff
@@ -66,13 +66,8 @@ pub struct Image2dArray {
...
 impl Image2dArray {
+    #[spirv_std_macros::gpu_only]
     pub fn sample(&self, sampler: Sampler, coord: Vec3A) -> Vec4 {
-        #[cfg(not(target_arch = "spirv"))]
-        {
-            let _ = sampler;
-            let _ = coord;
-            panic!("Image sampling not supported on CPU");
-        }
         #[cfg(target_arch = "spirv")]
```
khyperia, line 71: did you mean to delete this?

crates/spirv-std/src/lib.rs
```diff
@@ -52,8 +52,8 @@ pub use num_traits;
 pub use textures::*;
 
 /// Calls the `OpDemoteToHelperInvocationEXT` instruction, which corresponds to discard() in HLSL
+#[spirv_std_macros::gpu_only]
 pub fn demote_to_helper_invocation() {
-    #[cfg(target_arch = "spirv")]
```
khyperia, line 56: just calling out the change of semantics here for others, but I probably don't care either way.

khyperia (changes requested):

khyperia:
how does this handle unused argument warnings on the CPU?

crates/spirv-std-macros/src/lib.rs
```diff
@@ -29,3 +29,28 @@ pub fn spirv(_attr: TokenStream, item: TokenStream) -> TokenStream {
...
+        vis,
+        sig,
+        block,
+    } = syn::parse_macro_input!(item as syn::ItemFn);
+
+    let fn_name = sig.ident.clone();
+
+    let output = quote::quote! {
+        // Don't warn on unused arguments on the CPU side.
+        #[cfg_attr(not(target_arch = "spirv"), allow(unused_variables))]
```
XAMPPRocky, line 46: > how does this handle unused argument warnings on the CPU?

It adds `allow(unused_variables)` when compiled for CPU targets.

khyperia (approved):

PR #394, opened 2021-01-25
Add initial arch module

XAMPPRocky:
This adds the initial arch module which acts a low-level abstraction over SPIR-V, which currently only consists of a single function. We can add more as needed, but I wanted to add it now, since the it also adds some useful marker traits for writing generic functions.

### DIFF
--- a/crates/spirv-std/src/vector.rs
+++ b/crates/spirv-std/src/vector.rs
@@ -1,0 +1,16 @@
+/// Abstract trait representing a SPIR-V vector type.
+pub trait Vector<T: crate::scalar::Scalar>: crate::sealed::Sealed {}
+
+impl Vector<bool> for glam::BVec2 {}
+impl Vector<bool> for glam::BVec3 {}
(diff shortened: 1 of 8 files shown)


crates/spirv-std/src/vector.rs
```diff
@@ -0,0 +1,7 @@
+/// Abstract trait representing a SPIR-V vector type.
+pub trait Vector<T: crate::scalar::Scalar>: crate::sealed::Sealed {}
+
+impl Vector<f32> for glam::Vec2 {}
+impl Vector<f32> for glam::Vec3 {}
+impl Vector<f32> for glam::Vec3A {}
+impl Vector<f32> for glam::Vec4 {}
```
khyperia, line 7: I think this should implement it for integer vectors as well?

khyperia (changes requested):

khyperia (approved):
ah, didn't realize we didn't have the new glam yet, thanks!

PR #396, opened 2021-01-26
deduplicate OpVariable

khyperia:
I believe this may help getting things working on Android.

On an internal issue, @VZout said:

> My minimal test case right now is just 2 entry points with vertex index to reproduce the pipeline creation failure

Here's my theory. If you have two shaders set up like this:

```rust
#[spirv(vertex)]
pub fn vs_1(#[spirv(vertex_index)] vert_id: Input<i32>) { }
#[spirv(vertex)]
pub fn vs_2(#[spirv(vertex_index)] vert_id: Input<i32>) { }
```

then right now, rust-gpu will generate _two_ `OpVariable`s: one per entry point. Each OpVariable will get used in their respective vertex shader, and all is well and good... until we arrive at buggy android shader compilers. My guess is that it emits some automatic intrinsic code, or something, idk, when it encounters a vertex index OpVariable. Then, it hits *another* vertex index OpVariable, and says "wait we already did that, oops, crash", even though they're only used in separate shaders, never at the same time.

My theory is that emitting a _single_ OpVariable for _both_ shaders will make it happy. But, I don't have @VZout's testing code, so I'm putting this up and seeing if it works~

However! An open question is what to deduplicate _on_. If the descriptor set, binding, location, type, or any other decoration differs, then obviously we shouldn't deduplicate it. But... what if it has a different `OpName`? For some use cases, we should keep both around! They're different names, different things, shows up differently in OpenGL when searching for uniforms by name! But for some use cases, we should deduplicate them - for example, the following would generate _two_ vertex index OpVariables, which isn't great!

```rust
#[spirv(vertex)]
pub fn vs_1(#[spirv(vertex_index)] one: Input<i32>) { }
#[spirv(vertex)]
pub fn vs_2(#[spirv(vertex_index)] two: Input<i32>) { }
```

Right now, this code _does_ consider name to be part of the unique key of an OpVariable - so it will _not_ deduplicate the above, and keep two around.

---

Amusingly, I got this idea from a completely unrelated issue with a community member having issues with I think spirv-cross, which dumped out two identical variables in the source (because there's two OpVariables), which got sad. Context in the [arkadia discord here](https://discord.com/channels/750717012564770887/750717499737243679/788459465950625832).

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/duplicates.rs
+++ b/crates/rustc_codegen_spirv/src/linker/duplicates.rs
@@ -2,6 +2,7 @@ use crate::decorations::{CustomDecoration, ZombieDecoration};
 use rspirv::binary::Assemble;
 use rspirv::dr::{Instruction, Module, Operand};
 use rspirv::spirv::{Op, Word};
+use rustc_middle::bug;
 use std::collections::{hash_map, HashMap, HashSet};
 
 pub fn remove_duplicate_extensions(module: &mut Module) {
@@ -93,11 +94,25 @@ fn gather_annotations(annotations: &[Instruction]) -> HashMap<Word, Vec<u32>> {
         .collect()
 }
 
-fn make_type_key(
+fn gather_names(debugs: &[Instruction]) -> HashMap<Word, String> {
+    debugs
+        .iter()
+        .filter(|inst| inst.class.opcode == Op::Name)
+        .map(|inst| {
+            (
+                inst.operands[0].unwrap_id_ref(),
+                inst.operands[1].unwrap_literal_string().to_owned(),
+            )
+        })
+        .collect()
+}
+
+fn make_dedupe_key(
     inst: &Instruction,
     unresolved_forward_pointers: &HashSet<Word>,
     zombies: &HashSet<Word>,
     annotations: &HashMap<Word, Vec<u32>>,
+    names: &HashMap<Word, String>,
 ) -> Vec<u32> {
     let mut data = vec![];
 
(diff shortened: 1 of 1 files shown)


crates/rustc_codegen_spirv/src/linker/duplicates.rs
```diff
@@ -134,6 +149,22 @@ fn make_type_key(
...
+            if let Some(name) = names.get(&id) {
+                // Jump through some hoops to shove a String into a Vec<u32>.
+                for chunk in name.as_bytes().chunks(4) {
+                    let slice = match *chunk {
+                        [a] => [a, 0, 0, 0],
+                        [a, b] => [a, b, 0, 0],
+                        [a, b, c] => [a, b, c, 0],
+                        [a, b, c, d] => [a, b, c, d],
+                        _ => bug!(),
+                    };
```
eddyb, line 163: Might be simpler to start with a `[0; 4]` and write the bytes into it?

VZout:
Why did it merge.... @eddyb didn't aprove it.

XAMPPRocky:
@VZout We only have 1 required reviewer, so if you request more than that, whoever approves first will cause it to be merged.

eddyb:
> Why did it merge.... @eddyb didn't aprove it.

FWIW I was fine with merging it, and the only code change I might've requested was https://github.com/EmbarkStudios/rust-gpu/pull/396#discussion_r564546076 - but I expected more discussion might be needed on whether we *should* do this in the first place.

khyperia:
> @VZout We only have 1 required reviewer, so if you request more than that, whoever approves first will cause it to be merged.

Do note that we also have a requirement of 0 _requested_ reviewers (and 0 requested changes), which means that 1 person approving will _not_ cause it to be merged. But, @eddyb commented (like a reasonable person), and github says that clears the review request, so 0 requested reviews left, and merge.

> FWIW I was fine with merging it, and the only code change I might've requested was [#396 (comment)](https://github.com/EmbarkStudios/rust-gpu/pull/396#discussion_r564546076) - but I expected more discussion might be needed on whether we _should_ do this in the first place.

Yeah, was planning on doing that, as well as more discussion around OpName handling. Oh well I guess, can do that in a follow-up.

PR #397, opened 2021-01-28
Use rspirv::dr::Function's def_id convenience method.

eddyb:
I was checking if such a method exist and noticed code using `func.def.as_ref().unwrap().result_id` instead of `func.def_id()`.


khyperia (approved):
neat! I was the one that added `def_id` in https://github.com/gfx-rs/rspirv/pull/169, and I never went back to update anything written before that

PR #410, opened 2021-02-09
Add inline assembly docs

### LINKED ISSUE #390: Document specifics of inline assembly, as supported by rustc_codegen_spirv.
There's a few helpful features we have for `asm!`, like using `%name` to refer to abstract IDs, `typeof{...}` for getting types of inputs/outputs, and #376 added basic result type inference.

We probably want to document this somewhere, though it's somewhat "internal" in that ideally only `spirv-std` would need to use it.

XAMPPRocky:
### [Rendered](https://github.com/EmbarkStudios/rust-gpu/blob/inline-asm-docs/docs/src/inline-asm.md)

resolves #390

### DIFF
--- a/docs/src/inline-asm.md
+++ b/docs/src/inline-asm.md
@@ -51,11 +51,11 @@ pub unsafe fn vector_extract_dynamic<T: Scalar, V: Vector<T>>(vector: V, index: 
 
     asm! {
         "%vector = OpLoad _ {vector}",
-            "%element = OpVectorExtractDynamic _ %vector {index}",
-            "OpStore {element} %element",
-            vector = in(reg) &vector,
-            index = in(reg) index,
-            element = in(reg) &mut result
+        "%element = OpVectorExtractDynamic _ %vector {index}",
+        "OpStore {element} %element",
+        vector = in(reg) &vector,
+        index = in(reg) index,
+        element = in(reg) &mut result
     }
 
     result
--- a/docs/src/inline-asm.md
+++ b/docs/src/inline-asm.md
@@ -6,10 +6,66 @@ refer to the [SPIR-V specification][spec].
 
 [spec]: https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html
 
+### Basic syntax & usage.
+You can write inline assembly using the new `asm!` macro available with the
+`asm` feature on nightly. Refer to the [Rust unstable book][asm-docs] for more
+information on how to use the macro.
+
+[asm-docs]: https://doc.rust-lang.org/unstable-book/library-features/asm.html
+
+Non-ID arguments are written as-is, e.g.
+
+```rust
+asm! {
+    "OpCapability DerivativeControl"
+}
+```
+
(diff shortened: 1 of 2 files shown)


docs/src/inline-asm.md
```diff
@@ -0,0 +1,15 @@
...
+below. For more information on specific instruction behaviour and syntax, please
+refer to the [SPIR-V specification][spec].
+
+[spec]: https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html
+
+### Additional syntax
+
+| Syntax | Description |
+| ------ | ----------- |
+| `%name` | Used to refer to an abstract ID of a type. |
```
khyperia, line 13: `%name` refers to any abstract ID, not just types. Might be good to point out that we do the same thing as `spirv-as` in that every unique `%name` generates a new, unrelated ID, even if an integer is written.

khyperia (commented):
Do we want some examples about how `asm!` works? Also, this is missing a lot of syntax about how non-ID arguments (bitflags especially) are specified, do we want to add that too? Also things like syntax for instructions that return IDs, how to use `in` and `out` parameters, how to get around via pointers `asm!` not supporting some types, etc.

XAMPPRocky:
> Do we want some examples about how asm! works? 

I can probably include one or two basic examples and link to the inline asm RFC and docs.

> Also, this is missing a lot of syntax about how non-ID arguments (bitflags especially) are specified, do we want to add that too?

Sure, though you'll have to show me how they work, as I didn't know we were treating those specially. I know SPIR-V has bitflags like `OpImage*lod blah blah Grad | Lod` is that what you mean?

> Also things like syntax for instructions that return IDs, how to use in and out parameters, how to get around via pointers asm! not supporting some types, etc.

What are the parts around returning IDs, I can add a bit around using pointers.

khyperia:
> Sure, though you'll have to show me how they work, as I didn't know we were treating those specially. I know SPIR-V has bitflags like `OpImage*lod blah blah Grad | Lod` is that what you mean?

Yeah, the `Grad|Lod` stuff - specifically things like requiring no spaces. However, just basic notes around "non-ID arguments are written like `OpCapability DerivativeControl`, you slap the word in there and that's that" would be useful I think (which is related to providing examples showing the basics). If someone's already worked with SPIR-V and have viewed a lot of `spirv-dis` output, they might be familiar with a lot of the syntax, but I guess I want to explicitly point that out.

> What are the parts around returning IDs, I can add a bit around using pointers.

Sorry, writing too quickly, I mean stuff like `%x = Inst %t %y %z` - that `=` syntax isn't anywhere close to the spec (it's spec'ed as `Inst %t %x %y %z`). I intentionally tried to match `spirv-as` here, but it still is a very unique design decision that's not obvious on its own. (The formal thing is that any `Result <id>` is written on the LHS of an equals, not in the argument list as written in the spec).

Also, just had a thought, it might be nice to link to the `asm!` rustc docs as well to give docs on how the syntax of that works (it's definitely not obvious).

khyperia (approved):

PR #411, opened 2021-02-09
Add vector_insert_dynamic

XAMPPRocky:
Adds two new simple instructions to the arch module that map to `OpVectorInsertDynamic` and `OpCopyObject` instructions.

### DIFF
--- a/crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
+++ b/crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
@@ -371,8 +371,10 @@ pub fn instruction_signatures(op: Op) -> Option<&'static [InstSig<'static>]> {
         // 3.37.12. Composite Instructions
         Op::VectorExtractDynamic => sig! { (Vector(T), _) -> T },
         Op::VectorInsertDynamic => sig! {
-            // FIXME(eddyb) missing equality constraint between input and output vectors.
-            // (Vector(T), T, _) -> Vector(T)
+            // FIXME(eddyb) was `(Vector(T), T, _) -> Vector(T)` but that was
+            // missing an equality constraint between input and output vectors;
+            // we should use `(Vector(T, N), T, _) -> Vector(T, N)`, or constrain
+            // input and output vectors to have the same type some other way.
             (T, _, _) -> T
         },
         Op::VectorShuffle => sig! { (Vector(T), Vector(T)) -> Vector(T) },
--- a/crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
+++ b/crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
@@ -22,7 +22,7 @@
 use rspirv::spirv::Op;
 
 /// Pattern for a SPIR-V type, dynamic representation (see module-level docs).
-#[derive(PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq)]
 pub enum TyPat<'a> {
     /// Unconstrained type.
     Any,
@@ -97,7 +97,7 @@ impl TyPat<'_> {
 ///
 /// "Type lists" are used for `OpTypeStruct` fields, `OpTypeFunction` parameters,
 /// and operand types of instructions signatures.
-#[derive(PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq)]
 pub enum TyListPat<'a> {
     /// Unconstrained type list (any length, and the types can freely vary).
     Any,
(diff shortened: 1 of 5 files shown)


crates/spirv-std/src/arch.rs
```diff
@@ -27,3 +27,47 @@ pub unsafe fn vector_extract_dynamic<T: Scalar, V: Vector<T>>(vector: V, index:
...
+}
+
+/// Make a copy of `object`.
+/// 
+/// # Safety
+/// `T` must be a composite type (vector, array, or matrix).
+#[spirv_std_macros::gpu_only]
+#[doc(alias = "OpCopyObject")]
+#[inline]
+pub unsafe fn copy_object<T>(object: T) -> T {
```
khyperia, line 63: I believe `OpCopyObject` is an extremely esoteric bizarre instruction only used in either compilers that have very strange hacks in how they treat SSA form (and is functionally useless), and mmmaybe strange decoration edge cases, and there's no use in exposing it as a rust function. However, there's no _harm_ in exposing it I guess.

khyperia (approved):

crates/spirv-std/src/arch.rs
XAMPPRocky: Removed it for now.

crates/spirv-std/src/arch.rs
```diff
@@ -27,3 +27,28 @@ pub unsafe fn vector_extract_dynamic<T: Scalar, V: Vector<T>>(vector: V, index:
...
+/// number of components in `vector`.
+#[spirv_std_macros::gpu_only]
+#[doc(alias = "OpVectorInsertDynamic")]
+#[inline]
+pub unsafe fn vector_insert_dynamic<T: Scalar, V: Vector<T>>(vector: V, index: usize, element: T) -> V {
+    let mut result = V::default();
+
+    asm! {
+        "%vector = OpLoad _ {vector}",
+        "%new_vector = OpVectorInsertDynamic _ %vector {element} {index}",
```
XAMPPRocky, line 45: @khyperia || @eddyb Would you happen to know why inference would fail here?

eddyb: Given this, I'm guessing the `FIXME`?
https://github.com/EmbarkStudios/rust-gpu/blob/2f9d4d8e0d9dce5d06154929025e2c1fc0182a71/crates/rustc_codegen_spirv/src/spirv_type_constraints.rs#L373-L376

It requires adding another type pattern variable and a type pattern like `Either` but which is "both must match at once" - a bit like Rust's `x @ Foo(_)` - I almost did it, IIRC for @msiglreith, but then got swamped further by #300 (the implementation of which also interacts with this but not in any difficult way).

XAMPPRocky: Should I try to add that, or wait till #414 is merged?

eddyb: > Should I try to add that

Feel free to! It should be easy for me to rebase on such a change.

crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
```diff
@@ -372,7 +372,8 @@ pub fn instruction_signatures(op: Op) -> Option<&'static [InstSig<'static>]> {
         Op::VectorExtractDynamic => sig! { (Vector(T), _) -> T },
         Op::VectorInsertDynamic => sig! {
             // FIXME(eddyb) missing equality constraint between input and output vectors.
-            (Vector(T), T, _) -> Vector(T)
+            // (Vector(T), T, _) -> Vector(T)
+            (T, _, _) -> T
```
eddyb, line 376: LGTM, mind changing the comment to something like this? just so it's super clear what's going on:
```rust
            // FIXME(eddyb) was `(Vector(T), T, _) -> Vector(T)` but that was
            // missing an equality constraint between input and output vectors;
            // we should use `(Vector(T, N), T, _) -> Vector(T, N)`, or constrain
            // input and output vectors to have the same type some other way.
            (T, _, _) -> T
```

eddyb (changes requested):

crates/spirv-std/src/arch.rs
```diff
@@ -27,3 +27,33 @@ pub unsafe fn vector_extract_dynamic<T: Scalar, V: Vector<T>>(vector: V, index:
...
+    vector: V,
+    index: usize,
+    element: T,
+) -> V {
+    let mut result = V::default();
+
+    asm! {
+        "%vector = OpLoad _ {vector}",
+        "%element = OpLoad _ {element}",
+        "%new_vector = OpVectorInsertDynamic _ %vector %element {index}",
```
eddyb, line 50: cc @msiglreith this works now :tada:

eddyb (approved):

PR #418, opened 2021-02-11
Inline asm types

Jasper-Bekkers:
This most likely supersedes #372.

Various SPIRV instructions have operands that can require more operands when they're specified. Image instructions are one example, `OpDecorate` is another.

Todo:
 - [x] Add tests


eddyb (approved):

crates/rustc_codegen_spirv/src/linker/capability_computation.rs
```diff
@@ -104,28 +72,36 @@ fn operand_required_extensions(op: &Operand) -> &'static [&'static str] {
...
                 .chain(additional_extensions(module, inst))
         })
         .copied()
         .collect();
+
+    set.extend(
+        module
+            .all_inst_iter()
+            .flat_map(|inst| inst.operands.iter().flat_map(|op| op.required_extensions())),
+    );
```
khyperia, line 103: Why did you split this out to its own thing? It's less efficient this way

crates/rustc_codegen_spirv/src/linker/capability_computation.rs
```diff
@@ -104,28 +72,36 @@ fn operand_required_extensions(op: &Operand) -> &'static [&'static str] {
 // It says no. We strip it. Things explode.
 // So, this function is to encode any special version-specific rules that aren't in rspirv.
 fn additional_extensions(module: &Module, inst: &Instruction) -> &'static [&'static str] {
-    if module.header.as_ref().unwrap().version() < (1, 5)
-        && inst.class.opcode == Op::Capability
-        && inst.operands[0].unwrap_capability() == Capability::VulkanMemoryModel
-    {
-        &["SPV_KHR_vulkan_memory_model"]
+    let version = module.header.as_ref().unwrap().version();
```
khyperia, line 75: Probably good to move this let inside the if statement, as this is really hot code

crates/rustc_codegen_spirv/src/linker/capability_computation.rs
```diff
@@ -104,28 +72,36 @@ fn operand_required_extensions(op: &Operand) -> &'static [&'static str] {
...
+    if inst.class.opcode == Op::Capability {
+        match inst.operands[0].unwrap_capability() {
+            Capability::VulkanMemoryModel if version < (1, 5) => &["SPV_KHR_vulkan_memory_model"],
+            Capability::RuntimeDescriptorArray => &["SPV_EXT_descriptor_indexing"],
```
khyperia, line 79: I believe this is incorrect, it should be `if version < (1, 5)`

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
```diff
@@ -436,8 +492,8 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
         }
         if tokens.next().is_some() {
             self.tcx.sess.err(&format!(
-                "too many operands to instruction: {}",
-                instruction.class.opname
+                "too many operands to instruction: {} {:?}",
+                instruction.class.opname, instruction
```
khyperia, line 496: The debug impl includes the opname, probably best to print it only once (but also the debug impl is pretty gross, do we really want to include it in an error message?)

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
```diff
@@ -256,10 +256,47 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
...
+            Op::Extension => {
+                self.emit_global()
+                    .extension(inst.operands[0].unwrap_literal_string());
+                return;
+            }
+            Op::Capability => {
+                self.emit_global()
+                    .capability(inst.operands[0].unwrap_capability());
+                return;
+            }
```
khyperia, line 292: Why are these needed?

khyperia (approved):

khyperia (changes requested):
hit the wrong button, sorry, need some morning tea :(

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
eddyb: My guess was that `emit_global` helped with avoiding inserting these into the current block in a function, but maybe `rspirv` handles that already?

Or, wait, should `OpExtension`/`OpCapability` never be manually specified in `asm!`, and always deduced from other instructions?

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
eddyb: TIL `opname`, is that like `format!("{:?}", instruction.class.opcode)`?

crates/rustc_codegen_spirv/src/linker/capability_computation.rs
Jasper-Bekkers: Iterator sadness - `required_extensions` became something that returns a Vec<> instead of a &'static slice now and I could get the chain calls to match up anymore.

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
Jasper-Bekkers: They weren't supported before and were needed in my test sample, I can move them to a different PR if you feel the need.

@eddyb `OpCapability` and `OpExtension` cant always be deduced from the SPIR-V opcode unfortunately. In my testing I needed `SPV_EXT_descriptor_indexing` and `RuntimeDescriptorArray` to be able to index into an `OpTypeRuntimeArray` of images, which otherwise just isn't allowed.

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
Jasper-Bekkers: This is debug left-over that I can remove.

khyperia (approved):
LGTM pending tests if you'd like to write those! :tada:

Jasper-Bekkers:
Adding a test for `OpImageSampleProjExplicitLod` also to show how to use BitEnums.

PR #419, opened 2021-02-11
Don't auto-merge PRs with the `manual-merge` label set.

eddyb:
This change has been made by @eddyb from https://mergify.io config editor.


eddyb:
Oh this is weird, not really what I wanted.

PR #424, opened 2021-02-14
Fix wgpu-runner validation errors on minimize

Hentropy:
Vulkan swap chains must have non-zero dimensions, this adds a check for that during swap chain resizing and does not create one if it would be invalid.

### DIFF
--- a/examples/runners/wgpu/src/graphics.rs
+++ b/examples/runners/wgpu/src/graphics.rs
@@ -149,11 +149,16 @@ async fn run(
                 event: WindowEvent::Resized(size),
                 ..
             } => {
-                // Recreate the swap chain with the new size
-                sc_desc.width = size.width;
-                sc_desc.height = size.height;
-                if let Some(surface) = &surface {
-                    swap_chain = Some(device.create_swap_chain(surface, &sc_desc));
+                if size.width != 0 && size.height != 0 {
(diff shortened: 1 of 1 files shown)


khyperia:
I can't reproduce any validation errors, is this a platform-specific issue? (what platform are you on?)

Hentropy:
Oh, I'm on Windows. The swap chain error:
```
[5.019229 ERROR]()(no module): 
VALIDATION [VUID-VkSwapchainCreateInfoKHR-imageExtent-01689 (320081257)] : Validation Error: [ VUID-VkSwapchainCreateInfoKHR-imageExtent-01689 ] Object 0: handle = 0x1945c0b4eb8, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x13140d69 | vkCreateSwapchainKHR(): pCreateInfo->imageExtent = (0, 0) which is illegal. The Vulkan spec states: imageExtent members width and height must both be non-zero (https://vulkan.lunarg.com/doc/view/1.2.154.1/windows/1.2-extensions/vkspec.html#VUID-VkSwapchainCreateInfoKHR-imageExtent-01689)
object info: (type: DEVICE, hndl: 1736711032504)
```
There are similar ones for framebuffer creation and setting the viewport as well.

VZout (approved):

PR #432, opened 2021-02-16
Update version to match 0.2.1

XAMPPRocky:
Currently the version on crates.io is marked 0.2.1 which is what is used by ark, but the source code repo still has 0.2.0 in it's version. This updates it to match, and syncs the version numbers of all the crates to be the same.


VZout (approved):

khyperia:
I thought the current version has a breaking change against 0.2.0 and that's why we have 0.2.1 on a separate branch? (which reverts the breaking change)

XAMPPRocky:
@khyperia Updated to `0.3.0-alpha.0`

khyperia (approved):

PR #441, opened 2021-02-22
Add any and all arch instructions

XAMPPRocky:
This also adds a const generic (`N`) to vector for representing the length of a vector, this allows us to enforced requirements like "The Result type must have the same number of components as the operand", and allows us to create the correct types in SPIR-V generically.

### DIFF
--- a/crates/spirv-builder/src/test/basic.rs
+++ b/crates/spirv-builder/src/test/basic.rs
@@ -564,6 +564,31 @@ OpFunctionEnd",
 }
 
 #[test]
+fn any() {
+    val(r#"
+
+#[allow(unused_attributes)]
+#[spirv(fragment)]
+pub fn main() {
+    let vector = glam::BVec2::new(true, false);
+    assert!(arch::any(vector));
+}
+"#);
+}
+
+#[test]
+fn all() {
+    val(r#"
+#[allow(unused_attributes)]
+#[spirv(fragment)]
+pub fn main() {
+    let vector = glam::BVec2::new(true, true);
+    assert!(arch::all(vector));
+}
+"#);
+}
+
+#[test]
 fn image_read() {
     val(r#"
 #[allow(unused_attributes)]
--- a/crates/spirv-std/src/arch.rs
+++ b/crates/spirv-std/src/arch.rs
@@ -5,6 +5,72 @@
 //! no additional safety checks beyond type-checking.
 use crate::{scalar::Scalar, vector::Vector};
 
+/// Result is true if any component of `vector` is true, otherwise result is
+/// false.
+#[spirv_std_macros::gpu_only]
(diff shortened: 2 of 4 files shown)


crates/spirv-std/src/arch.rs
```diff
@@ -5,6 +5,72 @@
...
+pub fn any<V: Vector<bool, N>, const N: usize>(vector: V) -> bool {
+    let mut result = false;
+
+    unsafe {
+        asm! {
+            // Types & Constants
+            "%bool = OpTypeBool",
+            "%uchar = OpTypeInt 8 0",
+            "%uchar_0 = OpConstant %uchar 0",
+            "%uchar_1 = OpConstant %uchar 1",
```
eddyb, line 22: Why `uchar` and not `u8` or `byte`?

crates/spirv-std/src/arch.rs
```diff
@@ -5,6 +5,72 @@
...
+            "%bool_vec_type = OpTypeVector %bool {len}",
+            "%false_vec = OpConstantNull %glam_vec_type",
+            // Code
+            "%vector = OpLoad %glam_vec_type {vector}",
+            "%bool_vec = OpINotEqual %bool_vec_type %vector %false_vec",
+            "%result = OpAny %bool %bool_vec",
+            "%boolean = OpSelect %uchar %result %uchar_1 %uchar_0",
+            "OpStore {result} %boolean",
+            vector = in(reg) &vector,
+            len = const N,
```
eddyb, line 33: Love that parameterizing inline `asm!` by `const` generics is a thing now! (I wonder how it's implemented but I'm getting distracted...)

eddyb (changes requested):
LGTM, comments are mostly nits.

crates/spirv-builder/src/test/basic.rs
```diff
@@ -454,33 +454,6 @@ pub fn create_uninit_and_write() {
...
-#[spirv(fragment)]
-pub fn main() {
-    let vector = glam::Vec2::new(1.0, 2.0);
-    let expected = glam::Vec2::new(1.0, 3.0);
-    let new_vector = unsafe { spirv_std::arch::vector_insert_dynamic(vector, 1, 3.0) };
-    assert!(new_vector == expected);
-}
-"#);
-}
-
```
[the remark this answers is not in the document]
XAMPPRocky: Shouldn't be, bad merge on my part.

crates/spirv-std/src/arch.rs
XAMPPRocky: I believe it came from some spirv code from `spirv-dis` but I'll change it. I think u8 makes more sense.

eddyb (approved):

Issue #416, opened 2021-02-11
Storage class inference interface


khyperia:
Storage class inference is now a thing as of #300/#414. This details the design for the interface system taking advantage of inference, pointing out all the rules of how storage classes are specified in entry points.

There are a handful of ways of specifying a storage class:

1) direct attribute: `fn main(#[spirv(input)] x: &f32)`. This uses the current list of storage classes.
2) via builtin: `fn main(#[spirv(position)] x: &Vector3)` (inferred as input). This uses the current list of builtins (a table is needed to map builtin to storage class, I think some are input, some are output, and there may be others)
3) via image: `fn main(x: &Image2d)` (inferred as uniform_constant). I thiiink images are always uniform_constant and not uniform, but I'm not sure.
4) unspecified future inference rules that we haven't discovered yet (the spec is light on what goes in what storage class), suggestions are welcome

If exactly one of these rules matches, then use the storage class it specifies.

If more than one of these rules matches, and they all compute the same storage class, emit a warning (e.g. `#[spirv(position, input)]`, warn on `input` and say it's redundant). If more than one of these rules matches, and they compute different storage classes, emit an **error**.

If none of these rules matches, then we have an open design question. The options here are a trade-off between catching user errors that may be difficult to diagnose/guiding users with explicit syntax suggestions, and not annoying people with overly explicit syntax that takes a while to type out and read.

1) `fn main(x: &f32)` -> is this an error, or does it default to `input`? (or something else)
2) `fn main(x: &Struct)` -> is this an error, does it default to `input`, or does it default to `uniform`? (or something else)
3) `fn main(x: &mut f32)` -> is this an error, or does it default to `output`? (or something else)

For 2, I don't know if it's valid to have a struct (or any other non-scalar) be an `input`/`output` variable, more research is needed.

---

An alternative is to keep the current system of `Input<T>` and friends. We would *remove* the `.load()` and `.store()` methods entirely, and implement `Deref`/`DerefMut` (when applicable) for them. I much prefer the readability, recognizability, and usability of using plain references, but I understand others don't feel the same way~

eddyb:
> via builtin: `fn main(#[spirv(position)] x: &Vector3)` (inferred as input). This uses the current list of builtins (a table is needed to map builtin to storage class, I think some are input, some are output, and there may be others)

[...]

PR #443, opened 2021-02-22
Replace spirv_std::storage_class::X<T> with T/&mut T and #[spirv(x)] &T/&mut T.

eddyb:
~~This is my pre-#416 branch, I started this as part of #414, but only later realized I don't need it.~~

~~It's easier than I expected because, once again, I was able to mostly reuse existing logic (attribute parsing in this case).~~

As it's currently implemented, rules are:
* only `&T` and `&mut T` are allowed as entry parameter types, with explicit storage class attributes
  * should probably be `&mut MaybeUninitialized<T>` instead of `&mut T`, but ideally we can use `-> T` instead for `Output`
  * it considers `&T` immutable but it should also check that `T: Freeze`
* `T` defaults to `Input`, `&mut T` defaults to `Output`, and those storage classes cannot be explicitly specified
* all other storage classes use a `#[spirv(...)]` attribute on the parameter, e.g. this line in `sky-shader`:
  ```rust
      #[spirv(push_constant)] constants: &ShaderConstants,
  ```
* `UniformConstant`, and `PushContant` storage classes only allow `&T` (i.e. not `&mut T`) types
* `Private`, `Function` and `Generic` storage classes cannot be specified at all
  * should probably restrict this further based on shader stage or kernel vs shader, but we can do that later

~~This works and passes tests, but I'm opening it as a draft because it should wait on a decision regarding #416, especially around the `Output` storage class and whether we should use a `->` style, support bundling them in `struct`s etc.~~

**EDIT**: updated to reflect PR changes, i.e. `Input` not taking references anymore.


khyperia (approved):
Looks hecking great! @hrydgard and I had a conversation in discord, path forward seems to be to make inputs be by-value, and then in a follow-up, do the annoyingly complex work of making outputs be returned via tuple or struct or whatever.

Hentropy:
I mentioned this on the MCP as well, but representing PhysicalStorageBuffer as &/&mut would be really neat so switching to by-value for Input/Output would be wonderful.

eddyb:
`mut my_output: f32` just means `my_output: f32` with a `let mut my_output = my_output;` in the body.

*Even if* Rust-GPU wanted to do something funky, we don't *really* have control over what things like that mean.
What happens today is the SPIR-V entry-point calls the *regular* Rust `#[spirv(entry)] fn foo(...) {...}` with the right arguments.

eddyb:
@khyperia FWIW this is ready for a re-review, but I can't figure out what's wrong with the CI.
I don't want to tell it to retry because it might auto-merge because of the previous review.

XAMPPRocky:
I'm going to re-run and temporarily disable auto-merge

XAMPPRocky:
Yeah, it was an issue with GitHub Actions, the PR passes.

crates/rustc_codegen_spirv/src/attr.rs
```diff
@@ -298,6 +297,31 @@ impl CheckSpirvAttrVisitor<'_> {
                                 span,
                                 "attribute is only valid on a parameter of an entry-point function",
                             );
+                        } else {
+                            // FIXME(eddyb) should we just remove all 5 of these storage class
+                            // attributes, instead of disallowing them here?
```
XAMPPRocky, line 302: Would there ever be a need to use these attributes in `spirv-std`? If not, I agree that we should remove them.

XAMPPRocky (approved):

khyperia (approved):

Issue #389, opened 2021-01-21
MCP: Bindless support


Jasper-Bekkers:
Let's start off with the simplest hlsl shader I can more or less come up with to do bindless support, which I would like to port to rust-gpu.

Notice that instead of ByteArrayBuffer here, for the full feature one should be able to have (RW)Texture2D and other types as well.

```
ByteAddressBuffer g_byteAddressBuffer[] : register(t0, space3);
RWByteAddressBuffer g_rwByteAddressBuffer[] : register(u0, space4);

[numthreads(64, 1, 1)]
void main(int threadId: SV_DispatchThreadID)
{
    g_rwByteAddressBuffer[0].Store(threadId, g_byteAddressBuffer[0].Load(threadId));
}
```

Emitted SPIR-V from DXC:

```
; SPIR-V
; Version: 1.0
; Generator: Google spiregg; 0
; Bound: 30
; Schema: 0
               OpCapability Shader
               OpCapability RuntimeDescriptorArray
               OpExtension "SPV_EXT_descriptor_indexing"
               OpMemoryModel Logical GLSL450
               OpEntryPoint GLCompute %main "main" %gl_GlobalInvocationID
               OpExecutionMode %main LocalSize 64 1 1
               OpSource HLSL 660
               OpName %type_ByteAddressBuffer "type.ByteAddressBuffer"
               OpName %g_byteAddressBuffer "g_byteAddressBuffer"
               OpName %type_RWByteAddressBuffer "type.RWByteAddressBuffer"
               OpName %g_rwByteAddressBuffer "g_rwByteAddressBuffer"
               OpName %main "main"
               OpDecorate %gl_GlobalInvocationID BuiltIn GlobalInvocationId
               OpDecorate %g_byteAddressBuffer DescriptorSet 3
               OpDecorate %g_byteAddressBuffer Binding 0
               OpDecorate %g_rwByteAddressBuffer DescriptorSet 4
               OpDecorate %g_rwByteAddressBuffer Binding 0
               OpDecorate %_runtimearr_uint ArrayStride 4
               OpMemberDecorate %type_ByteAddressBuffer 0 Offset 0
               OpMemberDecorate %type_ByteAddressBuffer 0 NonWritable
               OpDecorate %type_ByteAddressBuffer BufferBlock
               OpMemberDecorate %type_RWByteAddressBuffer 0 Offset 0
               OpDecorate %type_RWByteAddressBuffer BufferBlock
[...]

PR #450, opened 2021-02-25
Initial work for bindless buffers & textures in rust-gpu

Jasper-Bekkers:
While a lot of this stuff is still in flux (naming wise, most code is in a terrible location, etc). I wanted to share the work so-far to gather some feedback.

There are several reasons for adding bindless support to `rust-gpu`;
 - Much lower CPU cost for updating bindings (bindings are created only at resource creation time, instead of on-demand like in a more traditional renderer); see for example this post https://ourmachinery.com/post/moving-the-machinery-to-bindless/
 - More flexibility when using resource on the GPU, meaning one can create nested data structures which was previously impossible to do.

Relevant MCP: #389

Example of how to use:

```rust
#[derive(Copy, Clone)]
pub struct Bindings {
    pub buffer: Buffer,
}

#[derive(Copy, Clone)]
#[spirv(block)]
pub struct ShaderConstants {
    pub bindings: SimpleBuffer<Bindings>,
    pub user_data0: u32,
    pub user_data1: u32,
    pub user_data2: u32,
}

struct Data {
    a: f32,
    b: u32,
    c: f32,
    d: u32,
}

#[spirv(gl_compute)]
pub fn main(constants: PushConstant<ShaderConstants>) {
    let data = Data {
        a: 1.0,
        b: 0,
        c: 6.0,
        d: 1782,
    };

    let bindings = constants.bindings.load();

    unsafe {
        bindings.buffer.store(0, data)
    };
}
```

Draft because; lots of work still left to do.

- [x] `store` method on `Buffer`
- [ ] `StorageTexture`
- [x] `StructuredBuffer<T>` equivalent that can function as an array instead of a "bag of bytes" like `Buffer`
- [ ] Remaining functions on `Texture`
- [x] Update `Texture` to use the generics that recently got added.
- [ ] Other `Texture1d` and `Texture3d`
- [ ] Test for the implementations of each, instead of just passing spirv-val and eye-balling the SPIR-V
- [ ] Add NonUniformResourceIndex equivalent


crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1974,6 +1975,128 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
             // needing to materialize `&core::panic::Location` or `format_args!`.
             self.abort();
             self.undef(result_type)
+        } else if self.internal_buffer_load_id.borrow().contains(&llfn_def) {
+            // simple structs are returned as values, complex ones are returned as out parameters
```
Jasper-Bekkers, line 1979: Is there a good way to detect this other then going, "well I see we have one extra out-parameter now"?

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1974,6 +1975,128 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
             // needing to materialize `&core::panic::Location` or `format_args!`.
             self.abort();
             self.undef(result_type)
+        } else if self.internal_buffer_load_id.borrow().contains(&llfn_def) {
```
Jasper-Bekkers, line 1978: Clearly the code below should be factored out into a separate function - but for now this was easier during development.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1974,6 +1975,128 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
+
+                        let constructed_adt = self
+                            .emit()
+                            .composite_construct(adt, None, composite_components)
+                            .unwrap()
+                            .with_type(adt);
+
+                        return self.store(constructed_adt, args[0], Align::from_bytes(0).unwrap());
+                    }
+                    _ => {}
```
Jasper-Bekkers, line 2095: There are quite a few more cases to handle here: scalars / vectors, also nested types etc.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -1974,6 +1975,128 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
+                }
+            }
+
+            bug!("Unhandled case for `internal_buffer_load` intrinsic");
```
Jasper-Bekkers, line 2099: Do we have a better way to do this?

crates/spirv-std/src/bindless.rs
```diff
@@ -0,0 +1,220 @@
...
+        // jb-todo: figure out why this assert breaks with complaints about pointers
+        // assert!(self.0.tag() == RenderResourceTag::Buffer);
+
+        unsafe { internal_buffer_load(self.0.index(), dword_aligned_byte_offset) }
+    }
+}
+
+#[derive(Copy, Clone)]
+#[repr(transparent)]
+pub struct Texture2d(RenderResourceHandle);
```
Jasper-Bekkers, line 127: Texture vs. Image ... Meh, I've always felt we should've called our types Textures since that's what everybody (except Vulkan / SPIR-V) calls them, so it's what I've called them here.

crates/spirv-std/src/bindless.rs
```diff
@@ -0,0 +1,220 @@
...
+/// And to sub-allocate descriptors from those tables. It must use `RenderResourceHandle` to
+/// refer to slots within this table, and it's then expected that these RenderResourceHandle's
+/// are freely copied to the GPU to refer to resources there.
+///
+/// | Buffer Type      | Set |
+/// |------------------|-----|
+/// | Buffers          | 0   |
+/// | Textures         | 1   |
+/// | Storage textures | 2   |
+/// | Tlas             | 3   |
```
Jasper-Bekkers, line 22: These will be hard requirements on a CPU runtime implementing a Vulkan renderer. Samplers aren't yet in this list because they're quite controversial here and a bit of a trade-off.

On low-end devices only 4 descriptor sets are guaranteed to exist for us to use. I've chosen here to dedicate the 4 slots like this and later on require only the use of immutable samplers. This is a bit of a contentious topic of course since low-end GPUs with limited DescriptorSets are ... rare, and having actual Samplers might be more useful.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
Jasper-Bekkers: `extern "unadjusted"` resolved this.

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
Jasper-Bekkers: Done.

crates/rustc_codegen_spirv/src/builder/load_store.rs
```diff
@@ -0,0 +1,506 @@
...
+                    uint_values_and_offsets,
+                );
+            }
+            _ => {
+                let mut err = self.tcx.sess.struct_err(
+                    "Type unsupported for `codegen_internal_buffer_store` return / args:",
+                );
+                err.note(&format!("type `{:?}`", val));
+                err.emit();
+            }
```
khyperia, line 109: We should support more types here - e.g. sized arrays. Listing every type and being explicit about what we don't support is probably best for maintainability when adding new types.

crates/rustc_codegen_spirv/src/builder/load_store.rs
```diff
@@ -0,0 +1,506 @@
...
+                            &field_names[element_idx]
+                        } else {
+                            "<unknown>"
+                        };
+
+                        self.fatal(&format!(
+                            "Trying to store to unaligned field: `{}::{}`. Field must be aligned to multiple of 4 bytes, but has offset {}",
+                            adt_name,
+                            field_name,
+                            offset.bytes() as u32));
```
khyperia, line 49: Is there a reason this is fatal and not an error? Seems better to make it an error.

crates/rustc_codegen_spirv/src/builder/load_store.rs
```diff
@@ -0,0 +1,506 @@
...
+                .access_chain(uniform_uint_ptr, None, sets.buffers, indices)
+                .unwrap()
+                .with_type(uniform_uint_ptr);
+
+            self.store(uint_value, access_chain, Align::from_bytes(0).unwrap());
+        }
+
+        self.emit_global()
+            .type_void()
+            .with_type(SpirvType::Void.def(rustc_span::DUMMY_SP, self))
```
khyperia, line 234: This... doesn't do what I think you think it does. I'm really confused if this passes spirv-val, it shouldn't, so I'm worried about test coverage if tests pass.

crates/rustc_codegen_spirv/src/builder/load_store.rs
```diff
@@ -0,0 +1,506 @@
...
+                base_offset_var,
+                element_offset_literal,
+                sets,
+            ),
+            _ => {
+                self.fatal(&format!(
+                    "Unhandled case for `internal_buffer_load` return / args: {:?}",
+                    data
+                ));
+            }
```
khyperia, line 503: Same here, should support more types.

crates/rustc_codegen_spirv/src/spirv_type.rs
```diff
@@ -217,8 +217,8 @@ impl SpirvType {
                     cx.zombie_even_in_user_code(
                         result,
                         def_span,
-                        "function pointer types are not allowed",
-                    )
+                        "function pointer types are not allowed (def)",
```
khyperia, line 220: forgotten println debugging change?

crates/spirv-std/src/bindless.rs
```diff
@@ -0,0 +1,273 @@
...
+#[derive(Copy, Clone)]
+#[repr(transparent)]
+pub struct Buffer(RenderResourceHandle);
+
+#[allow(unused_attributes)]
+#[spirv(internal_buffer_load)]
+#[spirv_std_macros::gpu_only]
+pub extern "unadjusted" fn internal_buffer_load<T>(_buffer: u32, _offset: u32) -> T {
+    unimplemented!()
+} // actually implemented in the compiler
```
khyperia, line 113: `#[allow(unused_attributes)]` shouldn't be needed anymore, if it is, then the hook into our macro system isn't working.

Also, I'm surprised this compiles on CPU if `extern "unadjusted"` requires `#[feature(abi_unadjusted)]`, what am I missing that makes this work?

khyperia (changes requested):
Didn't look too hard at the load/store algorithms, I'd like tests to pass before really getting into reviewing stuff, so I'm not accidentally reviewing things that are causing failures and wasting everyone's time

crates/rustc_codegen_spirv/src/builder/load_store.rs
Jasper-Bekkers: Switched it

crates/rustc_codegen_spirv/src/spirv_type.rs
Jasper-Bekkers: Reverted.

crates/spirv-std/src/bindless.rs
Jasper-Bekkers: Fixed.

crates/rustc_codegen_spirv/src/builder/load_store.rs
Jasper-Bekkers: I've added error message for un-supported types, and added support for fixed size arrays.

msiglreith:
Tried to get this running locally, some parts I noticed on the way:
- `RuntimeDescriptorArray` capability (and the corresponding extension on lower spv versions) isn't added, probably because I only use the buffer load/store parts in my shader?
- The storage buffers are decorated with `Uniform`/`BufferBlock` which is missing after spv1.3 and got replaced by `StorageBuffer`/`Block`, resulting in a validation error
- The `load`/`store` implementations generate warnings due to unnecessary `unsafe` blocks

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -104,6 +105,152 @@ impl<'tcx> CodegenCx<'tcx> {
...
+            def_id: None,
+            size: Some(Size::from_bytes(4)),
+            align: Align::from_bytes(4).unwrap(),
+            field_types: vec![runtime_array_uint],
+            field_offsets: vec![],
+            field_names: None,
+        }
+        .def(rustc_span::DUMMY_SP, self);
+
+        let runtime_array_struct = SpirvType::RuntimeArray {
```
msiglreith, line 150: My nvidia driver (latest beta 458.17) doesn't seem to be to happy about this one. Using a `[struct; 1]` array instead works, which glslangValidator generates as well

crates/rustc_codegen_spirv/src/builder/load_store.rs
Jasper-Bekkers: Added support for sized arrays, and list all of the types individually to provide better error messages.

crates/rustc_codegen_spirv/src/builder/load_store.rs
Jasper-Bekkers: Yeah it's quite an odd line, and I'm not sure if it's correct, so I could use some input here.

Jasper-Bekkers:
@msiglreith Thanks for testing SPIR-V 1.5, so far I've only tested 1.3 myself, pushed some fixes to make it work for 1.5 as well now. However,

> The descriptor sets are not added to the entry point definitions on newer spv versions

This one is a bit more tricky, it probably requires a compile-time flag or something to that effect to opt-in replace / set the bindings in the entry point.

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
Jasper-Bekkers: I'm not quite sure what this comment refers to and it's not caused any trouble for me so far (I'm on  460.89, but I don't think that should make a huge difference).

msiglreith: I tried with the Game Ready Driver 466.11 again and seen same result:

Using a runtime array type results in `0.0` values when loading values through the buffer interface.
spirv-dis:
```
       %uint = OpTypeInt 32 0
%_runtimearr_uint = OpTypeRuntimeArray %uint
 %_struct_20 = OpTypeStruct %_runtimearr_uint
%_runtimearr__struct_20 = OpTypeRuntimeArray %_struct_20
%_ptr_StorageBuffer__runtimearr__struct_20 = OpTypePointer StorageBuffer %_runtimearr__struct_20
         %11 = OpVariable %_ptr_StorageBuffer__runtimearr__struct_20 StorageBuffer
```

Using the mentioned fixed size array correctly loads the values from my shader.
spirv-dis:
```
%uint = OpTypeInt 32 0
%_runtimearr_uint = OpTypeRuntimeArray %uint
 %_struct_20 = OpTypeStruct %_runtimearr_uint
     %uint_1 = OpConstant %uint 1
%_arr__struct_20_uint_1 = OpTypeArray %_struct_20 %uint_1
%_ptr_StorageBuffer__arr__struct_20_uint_1 = OpTypePointer StorageBuffer %_arr__struct_20_uint_1
         %11 = OpVariable %_ptr_StorageBuffer__arr__struct_20_uint_1 StorageBuffer
```

<details><summary>Shader Code</summary>

```rust
#[repr(C)]
#[derive(Copy, Clone)]
pub struct LocalsPbr {
    world_to_view: f32x4x4,
    view_to_clip: f32x4x4,
}

#[spirv(vertex)]
pub fn mesh_vs(
    v_position_obj: f32x3,
    v_normal_obj: f32x3,
    v_texcoord: f32x2,
    v_tangent_obj: f32x4,
    #[spirv(position)] a_position: &mut f32x4,
    a_normal_world: &mut f32x3,
    a_texcoord: &mut f32x2,
    a_tangent_world: &mut f32x4,
    a_position_world: &mut f32x3,
) {
    *a_normal_world = v_normal_obj;
    *a_texcoord = v_texcoord;
    *a_tangent_world = v_tangent_obj;

    let pos_obj = v_position_obj;
    let pos_world = vec4(pos_obj.x, pos_obj.y, pos_obj.z, 1.0);
    *a_position_world = vec3(pos_world.x, pos_world.y, pos_world.z);

    let u_locals = spirv_std::bindless::internal_buffer_load::<LocalsPbr>(0, 0);

    let pos_view = pos_world * u_locals.world_to_view;
    let pos_clip = pos_view * u_locals.view_to_clip;

    *a_position = pos_clip;
}
```
</detail>

msiglreith: Ok, so it seems to work when building it as `spirv-unknown-spv1.2` target but not with the `spirv-unknown-spv1.3` or higher target.

crates/rustc_codegen_spirv/src/builder/load_store.rs
Jasper-Bekkers: I've pushed a change that addresses this - could you double check that it matches what we've discussed in our call?

khyperia: Yep, LGTM!

khyperia (approved):

PR #453, opened 2021-02-26
Safeness bug in StorageImage2d::write

Jasper-Bekkers:
It's pretty easy to see why this is unsafe, if multiple threads write to the same coordinate race-conditions happen (for example calling this with Coord(0, 0) in a pixel shader that covers more then one pixel would trample data).

Ultimately this should be addressed through something like #216 and some higher level abstractions on top of our buffer types, but since we don't have those for now, marking this as unsafe seems to be the only thing we can do for now.

### DIFF
--- a/crates/spirv-builder/src/test/basic.rs
+++ b/crates/spirv-builder/src/test/basic.rs
@@ -613,7 +613,9 @@ fn image_write() {
 #[spirv(fragment)]
 pub fn main(input: Input<glam::Vec2>, image: UniformConstant<StorageImage2d>) {
     let texels = *input;
-    image.write(glam::UVec2::new(0, 1), texels);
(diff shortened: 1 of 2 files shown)


eddyb (approved):
Seems straight-forward enough, cc @khyperia for visibility.

**EDIT**: oh can't land anyway, failed CI

XAMPPRocky:
LGTM, just need to update the test to use `unsafe`.

https://github.com/EmbarkStudios/rust-gpu/blob/98fb0e0a9425d1b3b2a517ca4e3f6418bb36654a/crates/spirv-builder/src/test/basic.rs#L616

khyperia (approved):
LGTM once the test is fixed

PR #455, opened 2021-03-01
Fix storage image method restrictions

XAMPPRocky:
This updates the const restrictions on `StorageImage2d::{write, read}` to more properly match the SPIR-V spec.

### DIFF
--- a/crates/spirv-std/src/textures.rs
+++ b/crates/spirv-std/src/textures.rs
@@ -64,11 +64,10 @@ pub struct StorageImage2d {
 impl StorageImage2d {
     /// Read a texel from an image without a sampler.
     #[spirv_std_macros::gpu_only]
-    pub fn read<I, V, V2, const N: usize>(&self, coordinate: V) -> V2
+    pub fn read<V, V2, const N: usize>(&self, coordinate: V) -> V2
     where
-        I: Integer,
-        V: Vector<I, N>,
-        V2: Vector<f32, 2>,
+        V: Vector<f32, 2>,
+        V2: Vector<f32, N>,
     {
         let mut result = V2::default();
 
@@ -89,11 +88,10 @@ impl StorageImage2d {
 
     /// Write a texel to an image without a sampler.
     #[spirv_std_macros::gpu_only]
-    pub fn write<I, V, V2, const N: usize, const N2: usize>(&self, coordinate: V, texels: V2)
+    pub fn write<V, V2, const N: usize>(&self, coordinate: V, texels: V2)
     where
-        I: Integer,
-        V: Vector<I, N>,
-        V2: Vector<f32, N2>,
+        V: Vector<f32, 2>,
+        V2: Vector<f32, N>,
     {
         unsafe {
             asm! {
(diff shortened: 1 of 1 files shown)


crates/spirv-std/src/textures.rs
```diff
@@ -89,11 +89,11 @@ impl StorageImage2d {
 
     /// Write a texel to an image without a sampler.
     #[spirv_std_macros::gpu_only]
-    pub fn write<I, V, V2, const N: usize>(&self, coordinate: V, texels: V2)
+    pub fn write<I, V, V2, const N: usize, const N2: usize>(&self, coordinate: V, texels: V2)
```
khyperia, line 92: Shouldn't `coordinate` be `Vec2` here?

crates/spirv-std/src/textures.rs
```diff
@@ -68,7 +68,7 @@ impl StorageImage2d {
     where
         I: Integer,
         V: Vector<I, N>,
-        V2: Vector<f32, N>,
+        V2: Vector<f32, 2>,
```
khyperia, line 71: Looks like you accidentally made the result be a 2-component vector instead of the coordinate.

khyperia (changes requested):

crates/spirv-std/src/textures.rs
XAMPPRocky: Well doesn't it say ["See the client API specification for handling of coordinates outside the image."](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpImageRead) for the coordinates?  So shouldn't coordinates be `N` to allow for that and the return vector should be `2` because it's a 2D unarrayed (is that how you say that?) image, so it returns Vector of 2?

khyperia: "coordinates outside the image" means if the coordinates are out-of-bounds, not if they're the wrong dimension. A 2d image will always take a Vec2 coordinate. Also, the size of the vector stored in the image has no correlation to the dimension of the image, e.g. a 2d image can store a 4-component pixel, it doesn't always have to store a Vec2.

crates/spirv-std/src/textures.rs
```diff
@@ -64,10 +64,9 @@ pub struct StorageImage2d {
 impl StorageImage2d {
     /// Read a texel from an image without a sampler.
     #[spirv_std_macros::gpu_only]
-    pub fn read<I, V, V2, const N: usize>(&self, coordinate: V) -> V2
+    pub fn read<V, V2, const N: usize>(&self, coordinate: V) -> V2
     where
-        I: Integer,
-        V: Vector<I, N>,
+        V: Vector<f32, 2>,
```
khyperia, line 69: `coordinate` isn't a `f32` vector, it is an integer vector. Sorry for my typo saying `Vec2` earlier, I meant to point out that it's a 2-component vector and wrote `Vec2` too quickly, I should have written `IVec2/UVec2`.

> Coordinate is an integer scalar or vector containing [...]

Might be best to add tests for this to make sure it compiles correctly.

khyperia (changes requested):

crates/spirv-std/src/textures.rs
XAMPPRocky: Oh nope yeah that's my bad, I got confused reading the spec, and misread the result type restriction as the coordinate restriction, reverting.

khyperia (approved):

PR #457, opened 2021-03-01
Update rust and add split-debuginfo option

XAMPPRocky:
This updates us to the latest Rust currently possible, as well adding adding the `split-debuginfo` option to the `Cargo.toml` to speed up builds on macOS (no-ops for other platforms).


khyperia (approved):

eddyb:
> (no-ops for other platforms)

Curious why that is - I recall building LLVM with split DWARF on linux being a massive boost.

XAMPPRocky:
@eddyb It's a no-op in that I believe `unpacked` is the default on Linux and Windows, and it's `packed` on macOS. So this only affects macOS builds.

eddyb:
Looking at both Cargo docs and implementation (the former of which are wrong! filed https://github.com/rust-lang/cargo/issues/9218), and rustc docs (https://doc.rust-lang.org/nightly/rustc/codegen-options/index.html#split-debuginfo), it seems that the status is this:
* default on ELF platforms (i.e. most UNIX systems, including Linux) is `off`
* default on windows and macOS is `packed`
* *only* macOS supports changing this option via stable `rustc`, so Cargo ignores it outside of macOS

So I'd say the option in `Cargo.toml` is noop *only for now*, and in a few releases we might see benefits on e.g. Linux as well.

There's also, separately, talk of making `unpacked` the default on some platforms, but it's unclear when/if that will happen.

XAMPPRocky:
Closing for now, as my revert was merged, so I'm going to make PR that updates to the latest nightly.

PR #458, opened 2021-03-01
Add entry point attribute configuration.

XAMPPRocky:
Adds an nested `entry_point` attribute to all the execution models attributes. This allows you to override the `OpEntryPoint` for a given entry point with your own name.

### Example

```rust
#[allow(unused_attributes)]
#[spirv(fragment(entry_point_name="hello_world"))]
pub fn main() { }
```

### DIFF
--- a/crates/spirv-builder/src/test/basic.rs
+++ b/crates/spirv-builder/src/test/basic.rs
@@ -30,12 +30,21 @@ pub fn main() {
 
 #[test]
 fn custom_entry_point() {
-    val(r#"
+    dis_globals(r#"
 #[allow(unused_attributes)]
-#[spirv(fragment(entry_point="hello_world"))]
-pub fn main() {
-}
-"#);
+#[spirv(fragment(entry_point_name="hello_world"))]
+pub fn main() { }
+"#,
+r#"OpCapability Shader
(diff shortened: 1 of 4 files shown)


crates/spirv-builder/src/test/basic.rs
```diff
@@ -29,6 +29,16 @@ pub fn main() {
...
 }
 
+#[test]
+fn custom_entry_point() {
+    val(r#"
+#[allow(unused_attributes)]
+#[spirv(fragment(entry_point="hello_world"))]
+pub fn main() {
+}
+"#);
```
eddyb, line 39: Maybe this should be a disasm test?

eddyb:
Implementation LGTM, but I'm not sure about the name of the field. I feel like it should maybe have the word `name` in it?
There's precedent in `#[linkage_name = "..."]` and `#[export_name = "..."]` in Rust, too.

Unless it's `#[spirv(fragment = "foo)]` but that feels too magical.

eddyb (changes requested):
You need to remove `#[allow(unused_attributes)]` from the PR description and the test, and rebase past #461.

eddyb (approved):

PR #463, opened 2021-03-02
Update example to use modern rust-gpu

XAMPPRocky:
### [Rendered](https://github.com/EmbarkStudios/rust-gpu/blob/XAMPPRocky-patch-1/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,19 +21,9 @@ However, many things aren't implemented yet: for example, loops and switches are
 ```rust
 use spirv_std::{glam::{Vec3, Vec4, vec2, vec3}, storage_class::{Input, Output}};
 
-#[derive(Copy, Clone)]
-#[repr(C)]
-#[spirv(block)]
-pub struct ShaderConstants {
-    pub width: u32,
-    pub height: u32,
-}
-
-#[allow(unused_attributes)]
 #[spirv(fragment)]
 pub fn main(
     #[spirv(frag_coord)] in_frag_coord: Input<Vec4>,
-    constants: PushConstant<ShaderConstants>,
     mut output: Output<Vec4>,
 ) {
     let frag_coord = vec2(in_frag_coord.x, in_frag_coord.y);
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,21 +19,39 @@ However, many things aren't implemented yet: for example, loops and switches are
...
-    
+use spirv_std::{glam::{Vec3, Vec4, vec2, vec3}, storage_class::{Input, Output}};
+
+#[derive(Copy, Clone)]
+#[repr(C)]
+#[spirv(block)]
+pub struct ShaderConstants {
+    pub width: u32,
+    pub height: u32,
+}
```
Jasper-Bekkers, line 30: Not sure if we should include this in the readme, we removed it last time to keep the sample small.

README.md
```diff
@@ -19,21 +19,39 @@ However, many things aren't implemented yet: for example, loops and switches are
...
+
+#[allow(unused_attributes)]
```
Jasper-Bekkers, line 32: We shouldn't need this anymore since 6571f45a

README.md
```diff
@@ -19,21 +19,39 @@ However, many things aren't implemented yet: for example, loops and switches are
...
+    let eye_pos = vec3(0.0, 0.0997, 0.2);
+    let sun_pos = vec3(0.0, 75.0, -1000.0);
+    let dir = get_ray_dir(uv, eye_pos, sun_pos);
+
     // evaluate Preetham sky model
     let color = sky(dir, sun_pos);
 
-    output.store(color.extend(0.0))
+    // Tonemapping
+    let color = color.max(Vec3::splat(0.0)).min(Vec3::splat(1024.0));
```
Jasper-Bekkers, line 52: Can we use `clamp` here and/or remove this clamp, it seems incorrect?

Jasper-Bekkers (changes requested):

README.md
XAMPPRocky: I've removed it for now, but this is copy-pasted from the `sky-shader`.

Jasper-Bekkers (approved):

PR #464, opened 2021-03-03
Refactor validation tests to use compiletest

### LINKED ISSUE #429: Use compiletest-rs for tests
# Proposal

Currently our tests are using our custom setup inside spirv-builder. This has a few disadvantages, namely that each test runs in serial because it's creating and destroying a single scratchpad project in order to create tests and re-use dependencies. However this process is pretty slow and it also currently requires us to build a lot of the infrastructure for tests ourselves, and while that might still be the case for some tests, for a lot of tests, we could use another testing framework.

One of the obvious alternatives at the moment, is the [compiletest](https://github.com/laumann/compiletest-rs) framework that used by rustc itself. It has numerous advantages for us.

- Uses rustc directly. Allowing us to compile spirv-std and rustc_codegen_spirv once and simply link them to rustc ourselves for the tests. This will allow the tests to run in parallel rather serially after setup.
- Has a rich testing framework built with a wide variety of features and is [pretty well documented](https://rustc-dev-guide.rust-lang.org/tests/adding.html#adding-new-tests).
  - Per-test test and rustc configuration.
  - UI tests and stdout & stderr normalisation, allowing us to test error messages.
  - "Revisions" allowing the same file to be used for multiple tests.
  - A variety of different testing modes that relevant for us, such as compile-pass, compile-fail, codegen, etc.
- More familiar for rustc developers and contributors, this would also allow tests to be much easier to pull downstream from rustc to ensure it also works on rust-gpu or upstream a test to rustc from rust-gpu.

It also has some tradeoffs.

- Tests are divided up across files and directories by convention. This isn't an advantage or disadvantage but simply a change in layout. However having tests be in separate complete Rust files does have the advantage of being able to be syntax highlighted and formatted by `rustfmt`.
- We obviously lose some amount of flexibility using a third party testing framework, though hopefully any missing features for us might also be useful for rustc and other codegen projects.

XAMPPRocky:
This PR moves nearly all validation tests out of `spirv-builder` and moves them into `tests/` which is a crate that uses completest to compile and build each each test. I've also added the `cargo testsuite` command which is a short alias to run the tests with `--release` mode, as it is substantially quicker (~40s in debug vs ~7s in release). You do need to run these tests in a separate command, as including it in the same workspace was leading to issues with conflicting files from cargo. Overall this implementation speeds up testing by quite a bit on my local machine.

Disassembly tests are still to be done, as I haven't figured how to get those working with SPIR-V yet, but that can be done in a follow-up PR.

closes #429


tests/ui/push_constant_vulkan.rs
```diff
@@ -0,0 +1,26 @@
+// Test that using push constants work.
+// NOTE(eddyb) we specifically run Vulkan validation here, as the default
+// validation rules are more lax and don't require a `Block` decoration
+// (`#[spirv(block)]` here) on `struct ShaderConstants`.
+
+// build-pass
+// compile-flags: -Ctarget-feature=+vulkan
```
khyperia, line 7: I don't think this runs vulkan validation, it only enables the vulkan memory model.

tests/build.rs
```diff
@@ -0,0 +1 @@
+fn main() {}
```
khyperia, line 1: Why is this needed?

Cargo.toml
```diff
@@ -1,4 +1,6 @@
+
```
khyperia, line 1: nit: extra line

tests/ui/push_constant_vulkan.rs
XAMPPRocky: Some we add a target feature to run Vulkan validation?

tests/build.rs
XAMPPRocky: Needed in an older version, but I forgot to delete it.

tests/ui/allocate_const_scalar_pointer.rs
```diff
@@ -0,0 +1,17 @@
+// Doesn't work, only worked before because I think it got optimized away before
+// hitting the backend.
```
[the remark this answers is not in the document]
XAMPPRocky: Hmm yeah, I can replicate it failing, when building as a shader, but it is passing in the testsuite. I'll have to investigate more.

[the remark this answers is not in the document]
XAMPPRocky: I figured out the issue. It was just getting optimised away again. I've updated the code to use the pointer and now it fails.

XAMPPRocky:
@khyperia This should be ready for a second round of review. I have fixed the issue with the test passing, when it should have failed (it was being optimised away). I've also added documentation, and a new `target_env` target feature flag that allows us to specify the target_env we're compiling for, and used to add Vulkan validation to the `push_constant_vulkan.rs` test.

All of the new target feature flag code should be in the second commit if you want to review that separately from the completest changes.

khyperia:
> I have fixed the issue with the test passing, when it should have failed (it was being optimised away)

Why was it being optimized away? Changing the test to make it work as expected seems like a workaround to the actual problem, which is that our test setup changed and caused a test to pass that shouldn't have. Changing the test instead of the test setup to make it fail again seems wrong, and may miss something critical that could have changed.

XAMPPRocky:
> Why was it being optimized away?

I figured it out. It was because of a different level of `debuginfo`, using cargo we were giving cargo's default (`2` for debug) and this PR was using rustc's default (`0`). I've updated and removed the changes now.

tests/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[[bin]]
+name = "compiletests"
+path = "./main.rs"
+
+[dependencies]
+compiletest = { version = "0.6.0", package = "compiletest_rs" }
+rustc_codegen_spirv = { path = "../crates/rustc_codegen_spirv" }
+
+# TODO: Needed for handling SPIR-V extension across platforms. Remove once
```
DeanBDean, line 17: Might be good to add an issue for this, just for visibility?

tests/ui/lang/core/ptr/allocate_const_scalar.rs
```diff
@@ -0,0 +1,18 @@
+// Doesn't work, only worked before because I think it got optimized away before
+// hitting the backend.
+
+// Needed because compiletest currently emits `\r\n` on Windows
+// normalize-stderr-test "\r\n" -> "\n"
```
eddyb, line 5: Hmm, how sure are we that this isn't caused by SPIRV-Tools? The Rust side should not be using `\r` at all.

tests/ui/arch/vector_insert_dynamic.rs
```diff
@@ -0,0 +1,12 @@
...
+// build-pass
+
+use spirv_std::arch;
+
+#[spirv(fragment)]
+pub fn main() {
+    let vector = glam::Vec2::new(1.0, 2.0);
+    let expected = glam::Vec2::new(1.0, 3.0);
+    let new_vector = unsafe { arch::vector_insert_dynamic(vector, 1, 3.0) };
+    assert!(new_vector == expected);
```
eddyb, line 11: This seems to be missing the `UVec` case from the original.
**EDIT**: ugh there are actually two "originals".

tests/ui/arch/vector_extract_dynamic.rs
```diff
@@ -0,0 +1,11 @@
+// Test `OpVectorExtractDynamic`
+// build-pass
+
+use spirv_std::arch;
+
+#[spirv(fragment)]
+pub fn main() {
+    let vector = glam::Vec2::new(1.0, 2.0);
+    let element = unsafe { arch::vector_extract_dynamic(vector, 1) };
+    assert!(2.0 == element);
```
eddyb, line 10: This seems to be missing the `UVec` case from the original.
**EDIT**: ugh there are actually two "originals".

eddyb (changes requested):

Cargo.toml
```diff
@@ -1,4 +1,5 @@
 [workspace]
+exclude = ["tests/", "target/"]
```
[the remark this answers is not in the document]
XAMPPRocky: Nope.

tests/ui/lang/core/ptr/allocate_const_scalar.rs
XAMPPRocky: That is a possibility, but I would have assumed `compiletest` would normalise the all of the `stderr` output.

tests/main.rs
```diff
@@ -0,0 +1,319 @@
...
+            .stderr(std::process::Stdio::inherit())
+            .stdout(std::process::Stdio::inherit())
+            .status()
+            .and_then(map_status_to_result)
+            .unwrap();
+
+        std::fs::write(TEST_DEPS_TOML_PATH, CARGO_TOML.as_bytes()).unwrap();
+        std::fs::write(
+            PathBuf::from(TEST_DEPS_PATH).join("src/lib.rs"),
+            TEST_DEPS_SRC,
```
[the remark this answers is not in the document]
XAMPPRocky: Nope, to successfully build it needs the crate attributes and an entrypoint.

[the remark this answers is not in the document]
eddyb: Is this because it's a `dylib`? Feels like it should work with an empty `lib.rs` if it were a `lib` crate.

tests/src/main.rs
```diff
@@ -0,0 +1,295 @@
...
+        &core,
+        &spirv_std,
+        &glam,
+        &spirv_std_macros,
+    ]
+    .iter()
+    .any(|o| o.is_none())
+    {
+        clean_project(manifest_dir);
+        build_spirv_std(manifest_dir, codegen_backend_path)
```
eddyb, line 160: Love it :shipit:.

eddyb (approved):

tests/ui/lang/core/ptr/allocate_const_scalar.rs
XAMPPRocky, line 5:
```suggestion
```

PR #465, opened 2021-03-03
Revert const generics

XAMPPRocky:
If we want to have ark use whatever commit we tag as `0.3` we need to revert the const generic part of `Vector` for now, as while it's stable for us, it won't be stable for Ark until the end of the month.

### DIFF
--- a/crates/spirv-std/src/arch.rs
+++ b/crates/spirv-std/src/arch.rs
@@ -10,6 +10,8 @@ use crate::{scalar::Scalar, vector::Vector};
 #[spirv_std_macros::gpu_only]
 #[doc(alias = "OpAny")]
 #[inline]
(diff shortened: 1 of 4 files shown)


crates/spirv-std/src/arch.rs
```diff
@@ -5,71 +5,72 @@
...
-            "OpStore {element} %boolean",
-            vector = in(reg) &vector,
-            len = const N,
-            element = in(reg) &mut result
-        }
-    }
-
-    result
-}
+// Uncomment after 25.03.2021 (Rust 1.51)
```
eddyb, line 8: You can use `#[cfg(any())]` instead of commenting.

eddyb (approved):

PR #475, opened 2021-03-06
Use Embark standard lints v0.2

repi:
Switches to use our new standard template for defining shared Rust and Clippy lints for our Embark projects, https://github.com/EmbarkStudios/rust-ecosystem/issues/59.

Did a couple of lint fixes from this but didn't want to make the change too intrusive so added some allow exceptions for now that can be revisited.


XAMPPRocky (approved):

msiglreith:
This re-added `crates/spirv-tools-sys` submodules which should be removed imo

XAMPPRocky:
@msiglreith Nice catch, don't know how I missed that, I've now removed them.

repi:
Oops, think I had it locally and even though I tried to remove it still had it come with. Submodules are a mysterious thing.

Thanks for removing it!

PR #476, opened 2021-03-08
Decouple glam from spirv-std, add const-generics feature flag.

XAMPPRocky:
This PR removes `glam` as a dependency from `spirv-std`. Since we still use it for our tests, we currently have to patch it with a forked version of `glam` that contains the trait implementations we need until `glam` can depend on these changes directly. It also exports `derivative::derive_impl` as a hidden export so that `glam` can use it in its implementation.

This PR also introduces a new `const-generics` feature flag, and reintroduces some of the const generics in order to be able enforce the restrictions on these types through the `Vector` trait. Having these under a feature flag, will allow us to iterate and add condt generic code without affecting Ark until they are released on stable at the end of the month, and then we can remove it for primitives.

[glam branch diff](https://github.com/bitshifter/glam-rs/compare/master...EmbarkStudios:spirv-std-impl)


crates/spirv-std/src/arch.rs
```diff
@@ -83,7 +82,11 @@ pub fn all<V: Vector<bool, N>, const N: usize>(vector: V) -> bool {
 #[spirv_std_macros::gpu_only]
 #[doc(alias = "OpVectorExtractDynamic")]
 #[inline]
-pub unsafe fn vector_extract_dynamic<T: Scalar, V: Vector<T>>(vector: V, index: usize) -> T {
+#[cfg(feature = "const-generics")]
+pub unsafe fn vector_extract_dynamic<T: Scalar, const N: usize>(
+    vector: impl Vector<f32, N>,
```
khyperia, line 87: I don't think this is correct?

> Vector must have a type OpTypeVector whose Component Type is Result Type.

examples/shaders/shared/Cargo.toml
```diff
@@ -9,3 +9,5 @@ publish = false
 [dependencies]
 spirv-std-macros = { path = "../../../crates/spirv-std-macros" }
 spirv-std = { path = "../../../crates/spirv-std" }
+glam = { version = "0.13", default-features = false, features = ["libm", "scalar-math"] }
+
```
khyperia, line 13: nit: extra line

crates/spirv-std/src/derivative.rs
```diff
@@ -74,25 +77,23 @@ macro_rules! deriv_fn {
...
         }
     };
 }
 
 // "must be a scalar or vector of floating-point type. The component width must be 32 bits."
 deriv_impl!(f32);
-deriv_impl!(glam::Vec2);
-deriv_impl!(glam::Vec3);
-deriv_impl!(glam::Vec3A);
-deriv_impl!(glam::Vec4);
```
khyperia, line 98: I would prefer to keep these in `spirv-std`, as they're not general-purpose functions that can be used by any consumer of glam.

crates/spirv-std/src/textures.rs
```diff
@@ -61,36 +61,36 @@ pub struct StorageImage2d {
 impl StorageImage2d {
     /// Read a texel from an image without a sampler.
     #[spirv_std_macros::gpu_only]
-    pub fn read<I, V, V2>(&self, coordinate: V) -> V2
+    #[cfg(feature = "const-generics")]
+    pub unsafe fn read<I, V, const N: usize>(&self, coordinate: impl Vector<I, 2>) -> V
```
khyperia, line 65: I'm assuming there was some discussion somewhere on why `read` should be unsafe - could you add context of why it's unsafe in doc comments? (rather than just `write` being unsafe, which IIRC is what we agreed on last week)

khyperia (dismissed):

crates/spirv-std/src/textures.rs
XAMPPRocky: Hmm, no I don't think we did that, and looking at this now I'm not sure why I changed that, removing.
```suggestion
    pub fn read<I, V, const N: usize>(&self, coordinate: impl Vector<I, 2>) -> V
```

crates/spirv-std/src/derivative.rs
XAMPPRocky: Sure, they are hidden from the public API, so they wouldn't have a stability guarantee, but I'll just copy a version of the macro into `glam` to add the trait impls. This will also remove the changes around macros that were needed to handle correctly calling `derive_impl!` from `glam`.

examples/shaders/shared/Cargo.toml
XAMPPRocky:
```suggestion
```

XAMPPRocky:
This should be ready for review. The main change is that I've added a new proc-macro called `vectorized` where it automatically creates a vector version of the function based on the arguments of the original function.

crates/spirv-std/src/vector.rs
```diff
@@ -1,16 +1,3 @@
...
-impl Vector<f32> for glam::Vec3A {}
-impl Vector<f32> for glam::Vec4 {}
-impl Vector<u32> for glam::UVec2 {}
-impl Vector<u32> for glam::UVec3 {}
-impl Vector<u32> for glam::UVec4 {}
-impl Vector<i32> for glam::IVec2 {}
-impl Vector<i32> for glam::IVec3 {}
-impl Vector<i32> for glam::IVec4 {}
+#[cfg(feature = "const-generics")]
+pub trait Vector<T: crate::scalar::Scalar, const N: usize>: Default {}
```
eddyb, line 3: Is this used to be able to assume the type is a SPIR-V `OpTypeVector`, in e.g. inline `asm!`? If so, it should be an `unsafe trait`, IMO.

crates/spirv-std/src/float.rs
```diff
@@ -0,0 +1,12 @@
+/// Abstract trait representing a SPIR-V floating point type.
+pub trait Float: num_traits::Float + crate::scalar::Scalar + Default {
```
eddyb, line 2: Shouldn't this be sealed and/or `unsafe`? (Same concern as with `Vector`)

eddyb (commented):
LGTM, modulo the concerns about whether the traits should be `unsafe trait`.

crates/spirv-std/src/float.rs
XAMPPRocky: Sealed is implied by `Scalar`, since implementing `Scalar` requires `Sealed`, so does implementing `Float`.

XAMPPRocky:
@khyperia This should be ready for a final review.

crates/spirv-builder/src/test/basic.rs
```diff
@@ -461,8 +461,11 @@ fn vector_extract_dynamic() {
 #[spirv(fragment)]
 pub fn main() {
     let vector = glam::Vec2::new(1.0, 2.0);
-    let element = unsafe { spirv_std::arch::vector_extract_dynamic(vector, 1) };
+    let element: f32 = unsafe { spirv_std::arch::vector_extract_dynamic(vector, 1) };
```
khyperia, line 464: Is this change needed now?

crates/spirv-builder/src/test/basic.rs
```diff
@@ -568,7 +575,7 @@ fn image_read() {
     val(r#"
 #[spirv(fragment)]
 pub fn main(image: UniformConstant<StorageImage2d>, mut output: Output<glam::Vec2>) {
-    let coords = image.read(glam::IVec2::new(0, 1));
+    let coords = unsafe { image.read(glam::IVec2::new(0, 1)) };
```
khyperia, line 578: Is this change needed now?

khyperia (commented):
LGTM pending some outdated stuff

XAMPPRocky:
I need one of you to give me the green checkmark to merge. 😄

crates/spirv-builder/src/test/basic.rs
```diff
@@ -463,6 +463,9 @@ pub fn main() {
     let vector = glam::Vec2::new(1.0, 2.0);
     let element = unsafe { spirv_std::arch::vector_extract_dynamic(vector, 1) };
     assert!(2.0 == element);
+    let uvector = glam::UVec2::new(1, 2);
+    let element: u32 = unsafe { spirv_std::arch::vector_extract_dynamic(uvector, 1) };
```
khyperia, line 467: sorry, meant to include this in my comment too but didn't want to spam too many comments
```suggestion
    let element = unsafe { spirv_std::arch::vector_extract_dynamic(uvector, 1) };
```

crates/spirv-builder/src/test/basic.rs
```diff
@@ -568,7 +575,7 @@ fn image_read() {
     val(r#"
 #[spirv(fragment)]
 pub fn main(image: UniformConstant<StorageImage2d>, mut output: Output<glam::Vec2>) {
-    let coords = image.read(glam::IVec2::new(0, 1));
+    let coords =  image.read(glam::IVec2::new(0, 1));
```
khyperia, line 578: nit: formatting
```suggestion
    let coords = image.read(glam::IVec2::new(0, 1));
```

khyperia (approved):
approved pending two more things, sorry!

khyperia:
oh. silly me, didn't see the automerge enable, that's why I didn't approve in my previous comment :P

crates/spirv-builder/src/test/basic.rs
XAMPPRocky: Spam me with review comments 😄

XAMPPRocky:
I'll make a followup PR for those

PR #480, opened 2021-03-08
feat: add image fetch

DeanBDean:
Adds [OpImageFetch](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpImageFetch)

### DIFF
--- a/crates/spirv-std/src/textures.rs
+++ b/crates/spirv-std/src/textures.rs
@@ -42,6 +42,28 @@ impl Image2d {
             result
         }
     }
+    /// Fetch a single texel with a sampler set at compile time
+    #[spirv_std_macros::gpu_only]
+    pub fn fetch<I, V>(&self, coordinate: V) -> Vec4
+    where
+        I: Integer,
+        V: Vector<I>,
+    {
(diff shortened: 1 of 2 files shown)


XAMPPRocky (approved):

crates/spirv-std/src/textures.rs
```diff
@@ -42,6 +42,28 @@ impl Image2d {
             result
         }
     }
+    /// Fetch a single texel with a sampler set at compile time
+    #[spirv_std_macros::gpu_only]
+    pub fn fetch<I, V>(&self, coordinate: V) -> Vec4
+    where
+        I: Integer,
+        V: Vector<I>,
```
XAMPPRocky, line 50: I think it's better the coordinate uses impl trait, this way you don't need to specify it if you need to set the integer type.

```suggestion
    pub fn fetch<I>(&self, coordinate: impl Vector<I>) -> Vec4
    where
        I: Integer,
```

DeanBDean: Makes sense, I have updated to this

PR #482, opened 2021-03-09
Use Embark standard lints v0.3

repi:
Fixes up our code to support the additional lints in the draft v0.3 of our standard set: https://github.com/EmbarkStudios/rust-ecosystem/issues/60. This is not final yet, but getting close there. Probably be quite a while until the next version after this, now we have most of the main lints we've been testing with.

The main change was with the [`match_same_arms`](https://rust-lang.github.io/rust-clippy/master/index.html#match_same_arms) lint, many match cases worked nicely and got merged together which do find easier to read, but some would require larger ordering changes & merges which looked like it could reduce readability in the complex backend code so left it with `allow´ in those few cases.

Leaving this as draft until  https://github.com/EmbarkStudios/rust-ecosystem/issues/60 is merged and v0.3 finalized. If you have any issues with any of these changes/lints, please do comment here and/or in the PR if we should not enable one of the lints.

### DIFF
--- a/crates/spirv-std/src/lib.rs
+++ b/crates/spirv-std/src/lib.rs
@@ -4,7 +4,7 @@
     feature(asm, register_attr, repr_simd, core_intrinsics, lang_items),
     register_attr(spirv)
 )]
-// BEGIN - Embark standard lints v0.2.
+// BEGIN - Embark standard lints v0.3
 // do not change or add/remove here, but one can add exceptions after this section
 // for more info see: <https://github.com/EmbarkStudios/rust-ecosystem/issues/59>
 #![deny(unsafe_code)]
@@ -23,13 +23,16 @@
     clippy::if_let_mutex,
     clippy::imprecise_flops,
     clippy::inefficient_to_string,
+    clippy::large_types_passed_by_value,
     clippy::let_unit_value,
     clippy::linkedlist,
     clippy::lossy_float_literal,
     clippy::macro_use_imports,
+    clippy::map_err_ignore,
     clippy::map_flatten,
     clippy::map_unwrap_or,
     clippy::match_on_vec_items,
+    clippy::match_same_arms,
     clippy::match_wildcard_for_single_variants,
     clippy::mem_forget,
     clippy::mismatched_target_os,
@@ -39,9 +42,12 @@
     clippy::pub_enum_variant_names,
     clippy::ref_option_ref,
     clippy::rest_pat_in_fully_bound_structs,
+    clippy::string_add_assign,
+    clippy::string_add,
     clippy::string_to_string,
     clippy::suboptimal_flops,
     clippy::todo,
+    clippy::unimplemented,
     clippy::unnested_or_patterns,
     clippy::unused_self,
     clippy::verbose_file_reads,
@@ -49,9 +55,15 @@
     nonstandard_style,
     rust_2018_idioms
 )]
-// END - Embark standard lints v0.2
+// END - Embark standard lints v0.3
 // crate-specific exceptions:
-#![allow(unsafe_code)] // still quite a bit needed
+#![allow(
+    // Needed for `asm!`.
+    unsafe_code,
+    // We deblierately provide an unimplemented version of our API on CPU
+    // platforms so that code completion still works.
+    clippy::unimplemented,
+)]
 
 #[cfg(not(target_arch = "spirv"))]
 #[macro_use]


XAMPPRocky (approved):
LGTM. There seems to be some trailing whitespace failing CI.

repi:
Fixed CI

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
```diff
@@ -60,7 +59,9 @@ impl<'tcx> CodegenCx<'tcx> {
             SpirvType::Integer(32, true) => self
                 .builder
                 .def_constant(SpirvConst::U32(ty, val as i64 as i32 as u32)),
-            SpirvType::Integer(64, true) => self.builder.def_constant(SpirvConst::U64(ty, val)),
+            SpirvType::Integer(64, false) | SpirvType::Integer(64, true) => {
```
eddyb, line 62: This can be collapsed further into `SpirvType::Integer(64, _)` - note that the `32`-bit case can also be collapsed, if you replace `val as i64 as i32 as u32` with `val as u32` (they're equivalent, no order of truncations is different from any other, since you're chopping off bits - {sign,zero}-extension can only happen when you're widening the type).

crates/rustc_codegen_spirv/src/spirv_type.rs
```diff
@@ -305,50 +305,43 @@ impl SpirvType {
...
-            Self::Opaque { .. } => Size::ZERO,
             Self::Vector { element, count } => {
                 cx.lookup_type(element).sizeof(cx)? * count.next_power_of_two() as u64
             }
             Self::Array { element, count } => {
                 cx.lookup_type(element).sizeof(cx)? * cx.builder.lookup_const_u64(count).unwrap()
             }
             Self::RuntimeArray { .. } => return None,
-            Self::Pointer { .. } => cx.tcx.data_layout.pointer_size,
-            Self::Function { .. } => cx.tcx.data_layout.pointer_size,
```
eddyb, line 322: This one is wrong I'm pretty sure, a SPIR-V function is not a function pointer, and should `return None` here - assuming that still can compile examples and tests, ofc.

crates/rustc_codegen_spirv/src/lib.rs
```diff
@@ -52,24 +55,27 @@
...
-// END - Embark standard lints v0.2
+// END - Embark standard lints v0.3
 // crate-specific exceptions:
 #![allow(
     unsafe_code,                // still quite a bit of unsafe
     clippy::map_unwrap_or,      // TODO: test enabling
     clippy::match_on_vec_items, // TODO: test enabling
     clippy::enum_glob_use,
-    clippy::todo,               // still lots to implement :)
+    clippy::todo                // still lots to implement :)
```
eddyb, line 78: Huh, we use `todo!()`? I remember trying to use `unimplemented!()` and getting an error from a lint.

crates/rustc_codegen_spirv/src/codegen_cx/type_.rs
```diff
@@ -156,7 +156,6 @@ impl<'tcx> BaseTypeMethods<'tcx> for CodegenCx<'tcx> {
     fn type_kind(&self, ty: Self::Type) -> TypeKind {
         match self.lookup_type(ty) {
             SpirvType::Void => TypeKind::Void,
-            SpirvType::Bool => TypeKind::Integer, // thanks llvm
             SpirvType::Integer(_, _) => TypeKind::Integer,
```
eddyb, line 159: This looks like a lint bug, it should suggest collapsing these two cases, but somehow it finds the other ones down below instead?

crates/rustc_codegen_spirv/src/lib.rs
repi: yup plenty of `todo!` in the crate

crates/rustc_codegen_spirv/src/codegen_cx/type_.rs
repi: It did suggest collapsing them, that's why I had to move it down to the integer arm, but formatted it a bit differently to keep the comment.

Would indeed be good if the lint took comments into consideration, as we talked about on discord. but this is the best I could get it to do without that.

any ideas on how to improve it here in the code?

crates/rustc_codegen_spirv/src/spirv_type.rs
repi: nice that the lint caught it, it does read better and easier to spot bugs when same arms are collapsed!

can you do an issue or fix it in a separate PR? not sure if good idea to fix and change behavior here in the lint PR

crates/rustc_codegen_spirv/src/codegen_cx/constant.rs
repi: sounds like a good idea

XAMPPRocky:
This PR now needs a rebase to be merged,

repi:
@XAMPPRocky can you take it over?

PR #485, opened 2021-03-10
builder: constant-fold `bx.mul(a, b)` in order to support `ptr::{read,write}`.

### LINKED ISSUE #462: `OpCopyMemorySized without OpCapability Addresses` in libcore on newer nightlies.
Nightly version: `2021-01-30`

Command: `cargo build`

- [libcore source](https://github.com/rust-lang/rust/blob/4f20caa6258d4c74ce6b316fd347e3efe81cf557/library/core/src/ptr/mod.rs#L911)

### Rust

```rust
use core::mem::MaybeUninit;
const MAYBEI32: MaybeUninit<&i32> = MaybeUninit::<&i32>::uninit();

pub fn create_uninit_and_write() {
    unsafe {
        asm! {
            "OpCapability Addresses"
        }
    }
    let mut maybei32 = MAYBEI32;
    unsafe { maybei32.as_mut_ptr().write(&0); }
    let _maybei32 = unsafe { maybei32.assume_init() };
}

#[spirv(fragment)]
pub fn main() {}
```

### Error

```
error: OpCopyMemorySized without OpCapability Addresses
   --> /Users/.rustup/toolchains/nightly-2021-01-30-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:894:9
    |
894 |         copy_nonoverlapping(&src as *const T, dst, 1);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `test-project`
```

eddyb:
`rustc_codegen_ssa` uses `bx.mul(count, size_of_pointee)` to implement e.g. `ptr::copy`, and functions like `ptr::read`, `ptr::write`, and some abstractions, rely on calling `ptr::copy(src, dst, 1)`, instead of having a separate intrinsic for the `count == 1` case.

Under `rustc_codegen_llvm`, this obviously works for arbitrary `count` values, but also it gets constant-folded, so by the time the `memcpy` call is created, the number of bytes to copy is a constant. I've replicated the constant-folding behavior in `rustc_codegen_spirv`, which gets rid of these kinds of errors (i.e. in the cases `OpCopyMemorySized` isn't *actually* needed):
```
error: OpCopyMemorySized without OpCapability Addresses
   --> core/src/ptr/mod.rs:894:9
    |
894 |         copy_nonoverlapping(&src as *const T, dst, 1);
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Sadly, this doesn't work through the `copy_from`/`copy_to` pointer methods (see comments on tests for a more in-depth explanation and potential solutions), but it looks like a lot of low-level library code avoids using the methods itself.

Fixes #462 by un-`#[ignore]`-ing the `create_uninitialized_memory` test.

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -1,5 +1,5 @@
 use super::Builder;
-use crate::builder_spirv::{BuilderCursor, SpirvValue, SpirvValueExt, SpirvValueKind};
+use crate::builder_spirv::{BuilderCursor, SpirvConst, SpirvValue, SpirvValueExt, SpirvValueKind};
 use crate::spirv_type::SpirvType;
 use rspirv::dr::{InsertPoint, Instruction, Operand};
 use rspirv::spirv::{Capability, MemoryModel, MemorySemantics, Op, Scope, StorageClass, Word};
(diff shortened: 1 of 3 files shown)


crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -21,9 +21,43 @@ use std::iter::empty;
...
+                            // as well, with the same `int(a, b) => ...` syntax.
+                            $((
+                                SpirvConst::U32(lhs_ty, $fold_int_lhs),
+                                SpirvConst::U32(rhs_ty, $fold_int_rhs)
+                            ) => {
+                                assert_eq!(lhs_ty, rhs_ty);
+                                let bits = self.lookup_type(lhs_ty).sizeof(self).unwrap().bits();
+                                if bits <= 32 {
+                                    let mask = !0u32 >> (32 - bits);
+                                    Some(SpirvConst::U32(lhs_ty, $fold_int & mask))
```
eddyb, line 47: Oops, I should use the `Size::truncate` helper instead of manually doing this.

XAMPPRocky (approved):
LGTM, but I'm still not too familiar with the `builder` part of the codebase so I'm going to wait for @khyperia to give a review before merging.

khyperia (approved):
neat! probably want to enable constant folds for other ops Eventually(tm) but I guess who cares right now.

PR #490, opened 2021-03-11
Bump rust-toolchain to nightly-2021-03-11.

eddyb:
Opened this as a draft, as this specific bump might be a bad idea because [this nightly lacks `rls`](https://rust-lang.github.io/rustup-components-history/) (which I still rely on).

I mostly did this to be able to test https://github.com/rust-lang/rust/pull/83019, and the fixes here can be reused when updating to an even newer nightly.
(and yes, with https://github.com/rust-lang/rust/pull/83019 we can finally use `for i in 0..n` loops in shaders!)

cc @XAMPPRocky for visibility


eddyb:
(Closed in favor of #493)

PR #493, opened 2021-03-15
Unlock `for i in 0..n` loops! (by bumping rust-toolchain to nightly-2021-03-17)

eddyb:
Most of the changes are uninteresting (and in their own commit), and were already part of #490.

What's new is @m-ou-se landed https://github.com/rust-lang/rust/pull/83022 in nightly, decoupling `mem::replace` from `mem::swap`.
So while `mem::swap` still doesn't work for us (as https://github.com/rust-lang/rust/pull/83019 wasn't approved), most things only use `mem::replace`, and that's in fact the case for `Range`'s `Iterator` impl, which was the last thing that had been blocking `for i in 0..n`.

And after this nightly bump, we now have e.g. this loop in `mouse-shader`:
```rust
    for i in 0..3 {
        painter.fill(
            mouse_button(i),
            // ...
        );
    }
```

However, we/I haven't had a chance to test more complicated patterns of iterators, and it would be great if everyone started experimenting with all kinds of iterators (either base ones or combinators etc.), to get a feel for what works and what doesn't.

### DIFF
--- a/crates/rustc_codegen_spirv/src/attr.rs
+++ b/crates/rustc_codegen_spirv/src/attr.rs
@@ -43,9 +43,10 @@ struct CheckSpirvAttrVisitor<'tcx> {
 }
 
 impl CheckSpirvAttrVisitor<'_> {
-    fn check_spirv_attributes(&self, hir_id: HirId, attrs: &[Attribute], target: Target) {
+    fn check_spirv_attributes(&self, hir_id: HirId, target: Target) {
         let parse_attrs = |attrs| crate::symbols::parse_attrs_for_checking(&self.sym, attrs);
 
+        let attrs = self.tcx.hir().attrs(hir_id);
         for (attr, parse_attr_result) in parse_attrs(attrs) {
             // Make sure to mark the whole `#[spirv(...)]` attribute as used,
             // to avoid warnings about unused attributes.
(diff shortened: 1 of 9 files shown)


termhn (approved):

XAMPPRocky (approved):

XAMPPRocky:
Hmm, seems we've ran into an ICE on ubuntu.


```
thread 'rustc' panicked at 'found unstable fingerprints for extern_mod_stmt_cnum(rustc_codegen_spirv[c17d]::decorations::_::_serde)', /rustc/d6eaea1c8860adb5302d2fbaad409e36585ab217/compiler/rustc_query_system/src/query/plumbing.rs:593:5

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust/issues/new?labels=C-bug%2C+I-ICE%2C+T-compiler&template=ice.md

note: rustc 1.52.0-nightly (d6eaea1c8 2021-03-14) running on x86_64-unknown-linux-gnu

note: compiler flags: -C prefer-dynamic -C embed-bitcode=no -C debuginfo=2 -C incremental --crate-type dylib

note: some of the compiler flags provided by cargo are hidden

query stack during panic:
#0 [extern_mod_stmt_cnum] computing crate imported by `decorations::_::_serde`
#1 [check_mod_unstable_api_usage] checking for unstable API usage in module `decorations`
end of query stack
error: could not compile `rustc_codegen_spirv`

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed

```

eddyb:
FWIW I can now reproduce with this:
```sh
rm -rf target
.github/workflows/test.sh
```

Will try to bisect it, but I'm worried it's a latent Cargo bug that `rustc` started detecting recently.

eddyb:
I'm sure now that this is fallout from https://github.com/rust-lang/rust/pull/83007, and specifically https://github.com/rust-lang/rust/issues/83126.

Latest progress on this was today (2h ago), so I suppose we'll just have to wait https://github.com/rust-lang/rust/pull/83128#issuecomment-799416952.

khyperia (approved):

PR #494, opened 2021-03-15
Use release when testing

XAMPPRocky:


eddyb:
Looks like this takes 4 minutes longer, so we'd have to do enough things differently for this to be worth it.
Maybe instead of `--release` we can try adding this to `Cargo.toml` (not permanently, just to check what it does):
```toml
[profile.dev.build-override]
opt-level = 3
```

eddyb:
Amazing, that's *even slower*. So we can't use a blanket approach here, it's probably better to just wait for the `compiletest` PR to land (FWIW I'm looking at https://github.com/EmbarkStudios/rust-gpu/actions/workflows/ci.yaml to see the times).

Issue #486, opened 2021-03-10
MCP: Require local size parameters in gl_compute attribute


Hentropy:
# Current Syntax
Currently, specifying the local size of a compute shader looks like this:
```rust
#[spirv(gl_compute(local_size_x = 3, local_size_y = 54, local_size_z = 23))]
pub fn main_cs() {}
```
If any (or all) of the local_size_* parameters are not specified they default to 1 which is inefficient.  There isn't really a sensible default here, it's very hardware specific.  In HLSL specifying local size is a hard requirement.

# Proposal
Specifying the dimensions should be a requirement, and the syntax should be cleaned up to improve usability:
```rust
// LocalSize 32, 23, 67
#[spirv(gl_compute(32, 23, 67))]
pub fn main_cs() {}

// Error!
#[spirv(gl_compute)]
pub fn main_cs() {}
```

Unspecified Y or Z dimensions could be assumed to be 1 for the common case of one or two dimensioned computations:
```rust
// LocalSize 32, 1, 1
#[spirv(gl_compute(32))]
pub fn main_cs() {}

// LocalSize 1, 45, 1
#[spirv(gl_compute(1, 45))]
pub fn main_cs() {}
```

XAMPPRocky:
> Specifying the dimensions should be a requirement, and the syntax should be cleaned up to improve usability

Thank you for your issue! I agree that it would be better for this to be hard requirement that causes a compile error, however I don't really agree the that the syntax proposed is "cleaner", or improves usability to me, because I think those numbers look very magic. There' no real indication as to what these numbers should be doing or what they represent.

If I was to pick new syntax, I think my preference would be for the following, it's just as clear as the current version, but is a lot shorter.

```rust
#[spirv(gl_compute(local_size(x = 3, y = 54, z = 23)))]
pub fn main_cs() {}
```

Hentropy:
The local_size bit is redundant since there is only one kind of dimensionality, and there isn't anything else the numbers could be referring to.

An alternative:
```rust
// LocalSize 1, 45, 1
#[spirv(gl_compute(y = 45))]
pub fn main_cs() {}
```
This extends well to LocalSizeId as well, if/when support for that is added:
```rust
// LocalSizeId 
#[spirv(gl_compute(y_id = /* syntax tbd */))]
pub fn main_cs() {}
```

The only other ExecutionModes that apply to gl_compute entry points are the SPV_KHR_float_controls ones.

[...]

PR #495, opened 2021-03-16
Require local size x dimension and remove gl_

Hentropy:
Implements #486

- Removes the `gl_` prefix from the compute shader attribute.
- Shortens the thread dimension declaration to `threads(x, y, z)`.
- Requires the x dimension to be specified.
- Trailing ones may be elided for the y or z dimensions.

```rust
// old syntax: #[spirv(gl_compute(local_size_x = 128, local_size_y = 32, local_size_z = 64))]
#[spirv(compute(threads(128, 32, 64)))]

// equivalent to: #[spirv(compute(threads(32, 1, 1)))]
#[spirv(compute(threads(32)))]

// equivalent to: #[spirv(compute(threads(64, 32, 1)))]
#[spirv(compute(threads(64, 32)))]
```

### DIFF
--- a/crates/rustc_codegen_spirv/src/symbols.rs
+++ b/crates/rustc_codegen_spirv/src/symbols.rs
@@ -198,7 +198,7 @@ const EXECUTION_MODELS: &[(&str, ExecutionModel)] = {
         ("tessellation_evaluation", TessellationEvaluation),
         ("geometry", Geometry),
         ("fragment", Fragment),
-        ("gl_compute", GLCompute),
+        ("compute", GLCompute),
         ("kernel", Kernel),
         ("task_nv", TaskNV),
         ("mesh_nv", MeshNV),
@@ -218,6 +218,7 @@ enum ExecutionModeExtraDim {
     X,
     Y,
     Z,
+    Tuple,
 }
 
 const EXECUTION_MODES: &[(&str, ExecutionMode, ExecutionModeExtraDim)] = {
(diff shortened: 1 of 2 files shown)


crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -690,6 +689,32 @@ fn parse_attr_int_value(arg: &NestedMetaItem) -> Result<u32, ParseAttrError> {
...
+                    }) if x <= u32::MAX as u128 => local_size[idx] = x as u32,
+                    _ => return Err((lit.span(), "must be a u32 literal".to_string())),
+                }
+            }
+            Ok(local_size)
+        }
+        _ => Err((
+            arg.span,
+            "attribute must have 1 to 3 parameters, trailing ones may be elided".to_string(),
+        )),
```
XAMPPRocky, line 714: I would add a separate match arm for the `is_empty` case, and have different error messages for having no parameters and too many parameters.

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -715,30 +740,21 @@ fn parse_entry_attrs(
...
-                                }
-                                Z => {
-                                    local_size[2] = val;
-                                }
-                                _ => unreachable!(),
+                                local_size.replace(parse_local_size_attr(attr)?);
+                            } else {
+                                return Err((
+                                    attr_name.span,
+                                    String::from("threads may only be specified once"),
```
XAMPPRocky, line 756:
```suggestion
                                    String::from("`#[spirv(compute(threads))]` may only be specified once"),
```

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -838,10 +854,13 @@ fn parse_entry_attrs(
...
-            let local_size = local_size.unwrap_or([1, 1, 1]);
-            entry
-                .execution_modes
-                .push((LocalSize, ExecutionModeExtra::new(local_size)));
+            if let Some(local_size) = local_size {
+                entry
+                    .execution_modes
+                    .push((LocalSize, ExecutionModeExtra::new(local_size)));
+            } else {
+                return Err((arg.span(), String::from("threads must be specified")));
```
XAMPPRocky, line 862:
```suggestion
                return Err((arg.span(), String::from("The `threads` argument must be specified when using `#[spirv(compute)]`")));
```

XAMPPRocky (commented):
LGTM, the new syntax seems fine to me. There could be a few improvements to the error messages and handling.

Hentropy:
@XAMPPRocky Done 🙂

XAMPPRocky (approved):

khyperia (approved):

Issue #383, opened 2021-01-18
SPIR-V Support


XAMPPRocky:
This is a tracking issue for tracking what rust-gpu can produce in terms of SPIR-V assembly.

#### Emoji Legend
- 🛑 — Missing
- ⚠️ — Incomplete
- ✅ — Complete

## Execution Models ⚠️

<details><summary>Details</summary>
<p>

- [x] Vertex
- [ ] Tessellation Control
- [ ] Tessellation Evaluation
- [x] Geometry
- [x] Fragment
- [ ] GLCompute
- [x] Kernel (Compute)
- [ ] Task
- [ ] Mesh
- [ ] Ray Generation
- [ ] Ray Intersection
- [ ] Ray Any-hit
- [ ] Ray Closest-hit
- [ ] Ray Miss
- [ ] Ray Callable

</p>
</details>

## Addressing Model ⚠️

<details><summary>Details</summary>
<p>

- [x] Logical
- [ ] Physical32
- [ ] Physical64
- [ ] PhysicalStorageBuffer64
- [ ] PhysicalStorageBuffer64EXT

</p>
</details>

## Memory Model ⚠️

<details><summary>Details</summary>
<p>

- [x] Simple
- [x] GLSL450
- [ ] OpenCL
- [x] Vulkan

</p>
</details>

## Execution Modes 🛑

<details><summary>Details</summary>

### All models 🛑
<details><summary>Details</summary>
<p>

- [ ] Initializer
- [ ] Finalizer
- [ ] SubgroupSize
- [ ] SubgroupsPerWorkgroup
- [ ] SubgroupsPerWorkgroupId
- [ ] PostDepthCoverage
- [ ] DenormPreserve
- [ ] DenormFlushToZero
- [ ] SignedZeroInfNanPreserve
- [ ] RoundingModeRTE
- [ ] RoundingModeRTZ

</p>
</details>

### Kernel 🛑
<details><summary>Details</summary>
<p>

- [ ] LocalSizeHint
- [ ] LocalSizeHintId
- [ ] VecTypeHint
- [ ] ContractionOff
- [ ] MaxWorkgroupSizeINTEL
- [ ] MaxWorkDimINTEL
- [ ] NoGlobalOffsetINTEL
- [ ] NumSIMDWorkitemsINTEL

</p>
</details>

### Kernel & GLCompute 🛑
<details><summary>Details</summary>
<p>

- [ ] LocalSize
- [ ] LocalSizeId
[...]

Issue #364, opened 2021-01-05
Supporting Image Operands


XAMPPRocky:
A lot of image operations (mainly those around sampling, or getting/setting texels from images) support [additional optional operands](https://www.khronos.org/registry/spir-v/specs/1.0/SPIRV.html#Image_Operands). At the SPIR-V level this comes in the form of an operand at the end of each instruction which is a 8 bit integer that acts as a bit mask for indicating the how many operands follow and what they are used for. Additionally certain operands can only be used with certain image parameters, (such as the `Bias` operand only being available for single sampled 1D, 2D, 3D, and Cube image.)

The question becomes how to best support using these operands at the Rust level and validate these constraints.

### Possibilities

- Have a different function for each operand.
   - **Advantage** Simplest to implement.
   - **Drawback** Requires dozens if not hundreds of functions to cover every possible combination for each instruction, leading to a lot of repetition.
- Use const generics on the functions for each instruction.
  - **Advantage** Potential to avoid a lot of repetition
  - **Drawback** Unclear how generic it could be at this point. 

One example of where it's unclear with constant generics is the `Grad` operand. Which allows you to pass 2 vectors or scalars of the same size as the `Dim`, so right now I'm unsure how we'd allow that to  generically while also validating the length.

Jasper-Bekkers:
For reference; GLSL has the following texture sampling functions

```glsl
ivec textureSize(gsampler sampler​, int lod​);
int textureQueryLevels(gsampler sampler​);
gvec texture(gsampler sampler​, vec texCoord​[, float bias​]);
gvec textureOffset(gsampler sampler​, vec texCoord​, ivec offset​[, float bias​]);
gvec textureProj(gsampler sampler​, vec projTexCoord​[, float bias​]);
gvec textureLod(gsampler sampler​, vec texCoord​, float lod​);
gvec textureGrad(gsampler sampler​, vec texCoord​, gradvec dTdx​, gradvec dTdy​);
gvec4 textureGather(gsampler sampler​​, vec texCoord​​, int comp​);
gvec4 textureGatherOffset(gsampler sampler​​, vec texCoord​​, ivec2 offset​, int comp​);
gvec4 textureGatherOffsets(gsampler sampler​​, vec texCoord​​, ivec2 offsets​[4], int comp​);
gvec textureProjLodOffset(gsampler sampler​, vec projTexCoord​, float lod​, ivec offset​);
```

[...]

PR #498, opened 2021-03-16
feat: add OpImageSampleExplicitLod to spirv-std

DeanBDean:
https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpImageSampleExplicitLod

Adds feature from [this tracking issue](https://github.com/EmbarkStudios/rust-gpu/issues/383)

### DIFF
--- a/crates/spirv-std/src/textures.rs
+++ b/crates/spirv-std/src/textures.rs
@@ -46,6 +46,65 @@ impl Image2d {
             result
         }
     }
+    #[spirv_std_macros::gpu_only]
+    #[cfg(feature = "const-generics")]
+    /// Sample the image at a coordinate by a lod
+    pub fn sample_by_lod<V: Vector<f32, 4>>(
+        &self,
+        sampler: Sampler,
+        coordinate: impl Vector<f32, 2>,
+        lod: f32,
+    ) -> V {
+        let mut result = Default::default();
+        unsafe {
+            asm!(
+                "%image = OpLoad _ {this}",
+                "%sampler = OpLoad _ {sampler}",
+                "%coordinate = OpLoad _ {coordinate}",
+                "%lod = OpLoad _ {lod}",
+                "%sampledImage = OpSampledImage _ %image %sampler",
+                "%result = OpImageSampleExplicitLod _ %sampledImage %coordinate Lod %lod",
+                "OpStore {result} %result",
+                result = in(reg) &mut result,
+                this = in(reg) self,
+                sampler = in(reg) &sampler,
+                coordinate = in(reg) &coordinate,
+                lod = in(reg) &lod
+            );
+        }
+        result
+    }
+    #[spirv_std_macros::gpu_only]
+    #[cfg(feature = "const-generics")]
+    /// Sample the image based on a gradient formed by (dx, dy). Specifically, ([du/dx, dv/dx], [du/dy, dv/dy])
+    pub fn sample_by_gradient<V: Vector<f32, 4>>(
+        &self,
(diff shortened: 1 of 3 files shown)


crates/spirv-std/src/textures.rs
```diff
@@ -172,6 +172,48 @@ impl Image2dArray {
             result
         }
     }
+    #[spirv_std_macros::gpu_only]
+    pub fn sample_by_lod(&self, sampler: Sampler, coordinate: Vec3A, lod: f32) -> Vec4 {
```
XAMPPRocky, line 176: This needs to be rebased onto the latest main and use the traits instead of glam types.

DeanBDean:
So I don't love the code duplication here, but I'm not sure on the best way to get out of it. Since importing the inputs in the `asm` macro is a TokenStream, I believe it would require a macro or proc macro to handle properly. I didn't want to make one just for one use case because it seems like it might tie in to this issue https://github.com/EmbarkStudios/rust-gpu/issues/364

crates/spirv-std/src/textures.rs
```diff
@@ -46,6 +46,65 @@ impl Image2d {
             result
         }
     }
+    #[spirv_std_macros::gpu_only]
+    #[cfg(feature = "const-generics")]
+    /// Sample the image at a coordinate by a lod
+    pub fn sample_by_lod<V: Vector<f32, 4>>(
+        &self,
+        sampler: Sampler,
+        coordinate: impl Vector<f32, 3>,
```
khyperia, line 55: Shouldn't this be a 2d vector?

crates/spirv-std/src/textures.rs
```diff
@@ -46,6 +46,65 @@ impl Image2d {
...
+        }
+        result
+    }
+    #[spirv_std_macros::gpu_only]
+    #[cfg(feature = "const-generics")]
+    /// Sample the image based on a gradient formed by (dx, dy). Specifically, ([du/dx, dv/dx], [du/dy, dv/dy])
+    pub fn sample_by_gradient<V: Vector<f32, 4>>(
+        &self,
+        sampler: Sampler,
+        coordinate: impl Vector<f32, 3>,
```
khyperia, line 83: same here

crates/spirv-std/src/textures.rs
DeanBDean: Yes, you are right on both counts

crates/spirv-std/src/textures.rs
```diff
@@ -46,6 +46,65 @@ impl Image2d {
...
+                "%coordinate = OpLoad _ {coordinate}",
+                "%lod = OpLoad _ {lod}",
+                "%sampledImage = OpSampledImage _ %image %sampler",
+                "%result = OpImageSampleExplicitLod _ %sampledImage %coordinate Lod %lod",
+                "OpStore {result} %result",
+                result = in(reg) &mut result,
+                this = in(reg) self,
+                sampler = in(reg) &sampler,
+                coordinate = in(reg) &coordinate,
+                lod = in(reg) &lod
```
khyperia, line 72: `lod` technically doesn't have to go through a pointer, but, shrug, whatever, consistency is neat too

khyperia (approved):

crates/spirv-std/src/textures.rs
DeanBDean: I honestly wasn't sure, so I did it out of ignorance moreso than consistency, but I'll take it

XAMPPRocky:
Thank you for your PR!

PR #499, opened 2021-03-16
Update compute shader attribute docs

### LINKED ISSUE #478: Document numthreads(x, y, z) attribute
Compute shaders need to be able to specify how many threads they use as an entry point attribute. Right now compute shaders are 1x1x1 which wastes a significant part of the machine's performance.

Hentropy:
Adds documentation for the new compute shader attribute syntax.  Closes #478


XAMPPRocky (approved):
Thank you for your PR!

PR #505, opened 2021-03-19
Use installed SPIR-V tools for compiletest

XAMPPRocky:
This PR changes the default to require installed tools and to disable compiling examples for tests, both of which help reduce compile times. We don't currently have any tests for these examples but they are always compiled when changes in the compiler happen which makes tests run longer.

### DIFF
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -9,6 +9,21 @@ fi
 
 os=$1
 
+function cargo_build() {
+    echo ::group::"$1 build"
+    cargo build \
+        --manifest-path "$1/Cargo.toml" \
+        --no-default-features \
+        --features "$FEAT"
+    echo ::endgroup::
+}
+
+function cargo_build_no_features() {
+    echo ::group::"$1 build"
+    cargo build --manifest-path "$1/Cargo.toml"
+    echo ::endgroup::
+}
+
 function cargo_test() {
     echo ::group::"$1 build"
     cargo test \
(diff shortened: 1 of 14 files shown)


XAMPPRocky:
Hmm, so it seems spirv-tools' error message changes based on whether we use installed or compiled. Do we want to require installed tools when using compiletest?

XAMPPRocky:
This PR roughly halves our CI time just comparing two runs.

Screenshot 2021-03-19 at 09 49 01

khyperia:
I thought we discussed this a while back and user machines and the default should always use compiled tools, and CI should always use installed tools. If CI is using compiled tools, that's a bug and needs to be fixed, but the default should be compiled tools.

khyperia:
Also, it looks like the examples aren't being built in CI anymore - looking at the logs for ubuntu, it says

```
examples/runners/ash build
      Finished test [unoptimized + debuginfo] target(s) in 0.09s
```

which doesn't seem right...

XAMPPRocky:
> hmm, I thought we discussed this a while back and user machines and the default should always use compiled tools, and CI should always use installed tools? If CI is using compiled tools, that's a bug and needs to be fixed, but I think the default should be compiled tools.

Well the problem with that setup is as I mentioned above the output for SPIR-V tool error output differs between compiled builds and installed builds, so we have to pick one for both.

XAMPPRocky:
Updated the tests to build the examples. 😅

khyperia:
> Well the problem with that setup is as I mentioned above the output for SPIR-V tool error output differs between compiled builds and installed builds, so we have to pick one for both.

Compiled tools are preferred IMO, because they're consistent, well-bundled, and less prone to unexpected system issues.

On a higher level though, could you give an example of the output difference? Specifically, is it issues like version mismatches giving completely different lints, output formatting differences, etc.? Could we fix the differences?

XAMPPRocky:
It's literally this.

```
-error: invalid binary:0:0 - No OpEntryPoint instruction was found. This is only allowed if the Linkage capability is being used.
341
+error: internal error:0:0 - No OpEntryPoint instruction was found. This is only allowed if the Linkage capability is being used.
```

khyperia:
Hmm, that seems unintentional and fixable (at the very least, doing a text postprocessing step). We probably shouldn't be spitting out what is presumably supposed to be a filename+line number in the error anyway.

XAMPPRocky:
> Hmm, that seems unintentional and fixable (at the very least, doing a text postprocessing step). We probably shouldn't be spitting out what is presumably supposed to be a filename+line number in the error anyway.

Yeah right now on our side, we could add `// normalize-stderr-test "invalid binary" -> "internal error"` to the test and it would work, I'd like something where you don't have to add that to every test about errors.

khyperia:
> Yeah right now on our side, we could add `// normalize-stderr-test "invalid binary" -> "internal error"` to the test and it would work, I'd like something where you don't have to add that to every test about errors.

I was more thinking fixing it on the side of https://github.com/EmbarkStudios/spirv-tools-rs, because, well, we own that code, and that code is responsible for the wrapping of the tools and making sure the output is consistent regardless of using compiled or installed tools.

XAMPPRocky:
filed https://github.com/EmbarkStudios/spirv-tools-rs/issues/19

XAMPPRocky:
Closing in favour of #546

PR #513, opened 2021-03-21
Fix Location calculation for arrays in entrypoint interface

Arc-blroth:
This PR fixes the calulation for Location decorations on array entrypoint parameters.

Both Vulkan ([spec][2]) and MLSL ([page 44 of the spec][3]) specify that array elements must take up consecutive locations, one for each element. Currently however, rust-gpu will always allocate 1 Location per entrypoint parameter, meaning that this code ([playground link][1]):
```rust
#[spirv(vertex)]
pub fn main_vs(
    data: Input<[f32; 4]>,
    more_data: Input<[f32; 4]>,
    mut out_data: Output<[f32; 4]>,
) {
```
will generate invalid SPIRV:
```
OpEntryPoint Vertex %1 "main_vs" %data %more_data %out_data
OpName %data "data"
OpName %more_data "more_data"
OpName %out_data "out_data"
OpDecorate %_arr_float_uint_4 ArrayStride 4
OpDecorate %data Location 0
OpDecorate %more_data Location 1
OpDecorate %out_data Location 0
```
since data and more_data have overlapping locations.

I'm not exactly that familiar with rustc and rustc_codegen_spirv, so there might be a better way to tell if an entrypoint parameter maps to a SPIRV array type than inspecting the generated type - please review :)

[1]: http://shader-playground.timjones.io/f15e88399b8c03dc131f7f758d53a1c2/vkspec.html#interfaces-iointerfaces-locations
[2]: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#interfaces-iointerfaces-locations
[3]: https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf

### DIFF
--- a/crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
@@ -457,7 +457,7 @@ impl<'tcx> CodegenCx<'tcx> {
                 std::iter::once(Operand::LiteralInt32(*location)),
             );
             // Arrays take up multiple locations
-            *location += if let SpirvType::Array { count, .. } = self.lookup_type(value_ty) {
+            *location += if let SpirvType::Array { count, .. } = self.lookup_type(value_spirv_type) {
                 self.builder
                     .lookup_const_u64(count)
                     .expect("Array type has invalid count value") as u32
--- a/crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
@@ -400,7 +400,15 @@ impl<'tcx> CodegenCx<'tcx> {
                 Decoration::Location,
                 std::iter::once(Operand::LiteralInt32(*location)),
             );
-            *location += 1;
+            // Arrays take up multiple locations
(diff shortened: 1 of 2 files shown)


XAMPPRocky (changes requested):
Thank you for your PR! This looks good to me, can you also add two tests? One to ensure that the behaviour works, and one that generates the `invalid entry param type` error? You can see similar tests in `tests/ui`.

Arc-blroth:
@eddyb pointed out that this PR is essentially blocked on #443 - I'll update this PR after that one is merged.

Arc-blroth:
Sorry for taking so long, but I've updated to HEAD and added a test for fixing array locations. (The CI seems to be failing because of commits that don't have any relationship to this PR)

XAMPPRocky (approved):
LGTM!

khyperia (approved):
LGTM, just adding a note to our future selves that this is incomplete, the vulkan spec says:

> 64-bit three- and four-component vectors consume two consecutive locations.

so we'll need to fix that too eventually (not needed in this PR).

khyperia:
Closing this for inactivity, if you'd like to revive this, please do!

PR #514, opened 2021-03-21
Clean up `recreate_swapchain` and some other things in the ash runner

expenses:


expenses:
I've checked the validation layers and they don't give out any warnings, but I didn't manage to get `heaptrack` to work with the `librustc_codegen_spirv.so` so it's possible (but unlikely) that I missed something and there's an tiny memory leak.

VZout (approved):

PR #515, opened 2021-03-22
Add error tests for #[spirv(...)] checks, and improve attribute handling ergonomics.

eddyb:
Note: the commits in this PR should be reviewed separately, especially since some definitions got moved around.

In terms of user-facing changes, we now disallow:
* duplicate attributes, e.g. `#[spirv(vertex, vertex)]`
* conflicting attributes, e.g. `#[spirv(vertex, fragment)]`
  * this is still conservative, in that it's based on the `SpirvAttribute` variant being the same between two different attributes, and we could have more precise checking for specific interactions

For both the pre-existing checks, and the new ones, compile error tests are added (now that we have `compiletest`).

Finally, the last commit replaces iterating over `enum SpirvAttribute`, with `struct AggregatedSpirvAttributes`, which only keeps one value for each variant of `SpirvAttribute` - this is because once we consider an error to have more than one of each, there's no reason to track any more, and it simplifies reasoning about attributes during codegen.


khyperia (approved):
#[spirv(:tada:)]

PR #516, opened 2021-03-22
Update nightly

XAMPPRocky:


crates/rustc_codegen_spirv/src/attr.rs
```diff
@@ -161,11 +161,6 @@ impl<'tcx> Visitor<'tcx> for CheckSpirvAttrVisitor<'tcx> {
         intravisit::walk_trait_item(self, trait_item)
     }
 
-    fn visit_struct_field(&mut self, struct_field: &'tcx hir::StructField<'tcx>) {
```
khyperia, line 164: This should be replaced with `visit_field_def`, right? It was renamed, not deleted: https://github.com/rust-lang/rust/commit/b25d3ba78118033b3f25b6de7a32e210d113872c#diff-90854f27d73f89fc2997be31fa36f58cb9af532d1387ac39751134afdd2ff8e2L154

khyperia (changes requested):

khyperia (approved):

PR #520, opened 2021-03-22
Add primitive set of arch instructions

XAMPPRocky:
These were small enough to just add.


khyperia (approved):
This looks good, but we should probably focus on more important things - geometry shaders are not on the roadmap any time soon.

XAMPPRocky:
> This looks good, but we should probably focus on more important things - geometry shaders are not on the roadmap any time soon.

Yep of course. I'm currently blocked on a fix for an ICE landing on nightly so I was working on other instructions and this seemed easy to do and complete. What else would be required to support geometry? AFAICT these are the only geometry specific instructions.

Issue #442, opened 2021-02-22
Support Atomics


XAMPPRocky:
Currently if you try to use an atomic you get the following error.

### Code
[Playground](http://shader-playground.timjones.io/f3d2124248dcb956f6139aa7f006d7b8)

```rust
#![no_std]
#![feature(register_attr)]
#![register_attr(spirv)]

use core::sync::atomic::AtomicI32;

#[allow(unused_attributes)]
#[spirv(vertex)]
pub fn main_vs() {
    let atomic = AtomicI32::from(65);
    
    atomic.fetch_add(35, core::sync::atomic::Ordering::AcqRel);
    assert!(atomic.load(core::sync::atomic::Ordering::AcqRel) == 100);
}
```

### Error

```
error: Cannot use AtomicOrdering=SequentiallyConsistent on Vulkan memory model. Check if AcquireRelease fits your needs.

error: Cannot use AtomicOrdering=SequentiallyConsistent on Vulkan memory model. Check if AcquireRelease fits your needs.

error: Cannot cast between pointer types
     |
     = note: from: *{Function} struct core::sync::atomic::AtomicI32 { v: i32 }
     = note: to: *{Function} u32
     = note: this error originates in a macro (in Nightly builds, run with -Z macro-backtrace for more info)

error: Cannot cast between pointer types
  --> D:\local\Temp\2b65e857-11aa-4ed1-a4a1-8dfa7f30591a\lib.rs:12:18
   |
12 |     let atomic = AtomicI32::new(65);
   |                  ^^^^^^^^^^^^^^^^^^
   |
   = note: from: *{Function} struct core::sync::atomic::AtomicI32 { v: i32 }
   = note: to: *{Function} u32
```

khyperia:
> Is there a particular reason atomics have not been merged?

We haven't had the time to do proper research/design/evaluation of the various possibility spaces, because it's not a priority for us. For example...

> since that would more or less require that shader inputs be AtomicU32 etc, which makes it hard to be generic over using atomic ops

I'm assuming rust had good reasons for creating the AtomicU32 etc. types instead of having them be operations on "regular" scalars (e.g. saftey, etc.). We should make sure we understand those reasons before blasting ahead with the status quo - rust-gpu is supposed to strive to "be as safe as regular rust", not blindly follow the (potentially unsafe) status quo.

dvc94ch:
[...]

PR #522, opened 2021-03-22
Add atomic arch instructions

XAMPPRocky:
This PR adds nearly all of the SPIR-V atomic instructions, the only ones not implemented are the ones around atomic flags as we don't have a type for them at the moment.

Depends on #519

### DIFF
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -201,7 +201,7 @@ fn invoke_rustc(builder: &SpirvBuilder, multimodule: bool) -> Result<PathBuf, Sp
         .unwrap_or_default();
 
     let rustflags = format!(
-        "-Z codegen-backend={} -Zsymbol-mangling-version=v0{}",
+        "-Z codegen-backend={}{}",
         rustc_codegen_spirv.display(),
         llvm_args,
     );
--- a/crates/spirv-std/src/arch.rs
+++ b/crates/spirv-std/src/arch.rs
@@ -7,12 +7,16 @@ use crate::{scalar::Scalar, vector::Vector};
 
 mod arithmetic;
 #[cfg(feature = "const-generics")]
+mod atomics;
+#[cfg(feature = "const-generics")]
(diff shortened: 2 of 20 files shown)


khyperia:
I thought we had discussed this already in https://github.com/EmbarkStudios/rust-gpu/issues/442 and the plan was to expose these through the rust stdlib, instead of custom instructions?

khyperia:
To be clear, if I recall correctly, the entire purpose of the arch module is intended to let us prototype high level abstractions over SPIR-V concepts without getting into compiler work. There is already a high level abstraction in place for atomics (`sync::atomic`), no prototyping is necessary, and so I don't want to waste work to add instructions to the arch module for sake of adding instructions to the arch module. I think our effort would be better directed towards things like fixing #442.

XAMPPRocky:
> I thought we had discussed this already in #442 and the plan was to expose these through the rust stdlib, instead of custom instructions?

Well there's no discussion in that thread, so if we did discuss it , it wasn't there and I don't remember it beyond "yes we should have them". I don't have a strong opinion on how we do it, but as I've read the spec more I'm not sure stdlib is complete abstraction for SPIR-V's atomics. For example there are way more memory semantics in SPIR-V than just `Relaxed`, `Release`, `Acquire`, `AcqRel`, and `SeqCst`. It also doesn't have any concept of `Scope`, and there's no `AtomicBool` equivalent in SPIR-V. So to me it seems like we'd be better providing our own abstraction.

eddyb:
(Re-opened temporarily to force push a rebase - the presence of the `atomics` branch, with a massive merge anomaly, was getting in the way of using `jj` on this repo, so I felt I had to untangle it)

eddyb:
> I can delete the branch if that helps.

@XAMPPRocky Only if you think it's been obsoleted by e.g.:
- #877

(even if you delete the branch now, I still thing my cleanup of it was worth it, because the PR doesn't look extremely broken - idk how much that persists after deleting the branch, but better safe than sorry)

PR #528, opened 2021-03-22
Follow best practice hints in the ash runner and reset the command pool instead of the buffer.

expenses:
This is honestly very nitpicky, but think of it as a learning experience :P If you run the ash runner with the environment variable `VK_LAYER_ENABLES=VK_VALIDATION_FEATURE_ENABLE_BEST_PRACTICES_EXT` then it recommends the following:

`PERFORMANCE [UNASSIGNED-BestPractices-vkCreateCommandPool-command-buffer-reset (-2027362524)] : Validation Performance Warning: [ UNASSIGNED-BestPractices-vkCreateCommandPool-command-buffer-reset ] Object 0: handle = 0x55dc63c69e40, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x8728e724 | vkCreateCommandPool(): VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT is set. Consider resetting entire pool instead.
`

This is mentioned on the page https://arm-software.github.io/vulkan_best_practice_for_mobile_developers/samples/performance/command_buffer_usage/command_buffer_usage_tutorial.html#resetting-the-command-pool, where supposedly, 
>Resetting the pool with vkResetCommandPool automatically resets all the command buffers allocated by it. Doing this periodically will allow the pool to reuse the memory allocated for command buffers with lower CPU overhead.
>
>To reset the pool the flag RESET_COMMAND_BUFFER_BIT is not required. In fact, it is better to avoid since it prevents using a single large allocator for all buffers in the pool, thus increasing memory overhead.

It probably only makes the most minuscule difference, but best practices are best practices 🤷🏻‍♀️

### DIFF
--- a/examples/runners/ash/src/main.rs
+++ b/examples/runners/ash/src/main.rs
@@ -950,10 +950,17 @@ impl RenderCtx {
                 .reset_fences(&[self.sync.draw_commands_reuse_fence])
                 .expect("Reset fences failed.");
 
+            // As we only have a single command buffer, we can simply reset the entire pool instead of just the buffer.
+            // Doing this is a little bit faster, see
+            // https://arm-software.github.io/vulkan_best_practice_for_mobile_developers/samples/performance/command_buffer_usage/command_buffer_usage_tutorial.html#resetting-the-command-pool
+            self.base
+                .device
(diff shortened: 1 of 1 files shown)


XAMPPRocky (approved):
Thank you for your PR!

Issue #42, opened 2020-10-01
Capability computation UX


khyperia:
Design and implement what I've been calling "capability computation":

1) Let the user specify what capabilities to allow. Exactly _how_ they do it is open right now - e.g. attributes, or just a `-C target-feature` temp hack.
2) Validate those capabilities are correct. Error on missing capability, warning on not-needed capability?

Note validation is surprisingly complex, e.g. some capabilities say "you can't opbitcast this particular type to that particular type", it's not just simple "this instruction is/isn't allowed" rules.

~~@Jasper-Bekkers is working on part 2 right now.~~

khyperia:
I really like the system that @XAMPPRocky came up with, although there's a chunk of bikeshedding left to do. The system goes something roughly like this:

1) Functions that require capabilities declare that they **need** specific capabilities:
    ```rust
    #[spirv(needs_capability = "DerivativeControl")]
    fn derivative(x: f32) -> f32 { ... }
    ```
2) End-user modules (the crate that builds a binary, not an rlib) declare that they **have** capabilities:
    ```rust
    #![spirv(declare_capability = "DerivativeControl")]
    ```
3) If a module uses a function that `needs_capability` that it does not `declare_capability`, an error occurs ("ya need to add `declare_capability = "blah"` to your module", etc.)
4) The capabilities declared by `declare_capability` (and _only_ those capabilities) are emitted to the SPIR-V binary.

This prevents issues like #468 being extremely confusing - if you accidentally use some function (in the case of #468, deep within a library) that `needs_capability = Int64` (perhaps functions using `u64` are automatically declared with `needs_capability = Int64` implicitly, to avoid needing to modify libraries), then you get a **compiler error**, rather than **silently adding `OpCapibility Int64` to your module**.

Bikesheddy questions:

1) If `a()` calls `b()` calls `c()`, and `c` has `#[spirv(needs_capability = "blah")]`, does `b` *also* need to declare `needs_capability = "blah"`? It's annoying to do so, but would be more clear.
2) `needs_capability` and `declare_capability` and the `=` syntax are temporary placeholders, what should they actually be?

XAMPPRocky:
For declaring capabilities I was thinking more about it and I think it could be good if attached the declaring of capabilities to entrypoint functions. That way if you use `cfg` to remove that entrypoint, you cfg out all of the capabilities it uses as well.

```rust
#[spirv(fragment)]
#[spirv(capabilities(kernel))]
[...]

PR #529, opened 2021-03-23
Add capability attribute

XAMPPRocky:
This PR adds the `#[spirv(capabilities)]` attribute, which accepts a list of capabilities on an entrypoint function and emits those capabilities.

### To Do

- [x] Documentation
- [x] Ensure you can only use `#[spirv(capabilities)]` with entry point functions.

### DIFF
--- a/crates/rustc_codegen_spirv/src/attr.rs
+++ b/crates/rustc_codegen_spirv/src/attr.rs
@@ -5,8 +5,8 @@
 use crate::codegen_cx::CodegenCx;
 use crate::symbols::Symbols;
 use rspirv::spirv::{
-    AccessQualifier, BuiltIn, Dim, ExecutionMode, ExecutionModel, ImageFormat, StorageClass,
-    Capability
+    AccessQualifier, BuiltIn, Capability, Dim, ExecutionMode, ExecutionModel, ImageFormat,
+    StorageClass,
 };
 use rustc_ast::Attribute;
 use rustc_hir as hir;
@@ -198,7 +198,9 @@ impl AggregatedSpirvAttributes {
             StorageClass(value) => {
                 try_insert(&mut self.storage_class, value, span, "storage class")
             }
-            Capability(value) => try_insert(&mut self.capability, value, span, "#[spirv(capability)]"),
+            Capability(value) => {
+                try_insert(&mut self.capability, value, span, "#[spirv(capability)]")
+            }
             Builtin(value) => try_insert(&mut self.builtin, value, span, "builtin"),
             DescriptorSet(value) => try_insert(
                 &mut self.descriptor_set,
@@ -371,8 +373,10 @@ impl CheckSpirvAttrVisitor<'_> {
 
         match aggregated_attrs.capability {
             Some(spanned) if aggregated_attrs.entry.is_none() => {
-                self.tcx.sess.span_err(spanned.span, "`#[spirv(capability)]` can only be used with entry point functions.");
-
+                self.tcx.sess.span_err(
+                    spanned.span,
+                    "`#[spirv(capability)]` can only be used with entry point functions.",
+                );
             }
             _ => {}
         }
--- a/crates/rustc_codegen_spirv/src/attr.rs
+++ b/crates/rustc_codegen_spirv/src/attr.rs
@@ -85,7 +85,7 @@ pub enum SpirvAttribute {
 
     // `fn` attributes:
     Entry(Entry),
-    Capabilities(Vec<Capability>),
+    Capability(Vec<Capability>),
 
     // (entry) `fn` parameter attributes:
     StorageClass(StorageClass),
@@ -117,7 +117,7 @@ pub struct AggregatedSpirvAttributes {
 
     // `fn` attributes:
     pub entry: Option<Spanned<Entry>>,
-    pub capabilities: Option<Spanned<Vec<Capability>>>,
+    pub capability: Option<Spanned<Vec<Capability>>>,
 
     // (entry) `fn` parameter attributes:
     pub storage_class: Option<Spanned<StorageClass>>,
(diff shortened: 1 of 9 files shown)


dragostis:
Just out of curiosity, could capabilities be detected at compile time and would that be desirable?

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -596,6 +600,147 @@ fn parse_image_type(
...
+    let list = match arg.meta_item_list() {
+        Some(list) => list,
+        None => return Err((arg.span(), "`#[spirv(capabilities())]` requires a list.".to_string())),
+    };
+
+    Ok(list.iter().filter_map(NestedMetaItem::ident).filter_map(ident_to_capability).collect())
+}
+
+fn ident_to_capability(ident: Ident) -> Option<Capability> {
+    Some(match &*ident.to_string() {
```
eddyb, line 613: `to_string` is unnecessarily expensive (only `proc_macro` suffers from this limitation).

But you don't even need to match on `ident.as_str()` - check out what `const BUILTINS` & friends do, I suspect you want something similar.

tests/src/main.rs
```diff
@@ -60,11 +60,21 @@ fn main() {
     let codegen_backend_path = find_rustc_codegen_spirv();
     let libs = build_deps(&deps_target_dir, &codegen_backend_path);
 
+    run_mode(
+        "compile-fail",
```
eddyb, line 64: Aren't non-UI tests deprecated? If you want `//~ ERROR` annotations, you can still use them with UI, but IMO it's really valuable to keep the output around, to be able to catch it changing in unexpected ways.

crates/rustc_codegen_spirv/src/attr.rs
```diff
@@ -163,6 +166,10 @@ impl AggregatedSpirvAttributes {
             }
         }
 
+        if aggregated_attrs.capabilities.is_some() && aggregated_attrs.entry.is_none() {
+            cx.tcx.sess.err("#[spirv(capabilities)] can only be used with entry point functions.");
+        }
```
eddyb, line 171: For the record, this belongs at the end of `check_spirv_attributes`.

To avoid confusion around this in the future, maybe `AggregatedSpirvAttributes::parse` should be renamed to `AggregatedSpirvAttributes::gather` or something?

XAMPPRocky:
@dragostis In theory, I believe so, I believe you probably could visit each instruction, see what capabilities it requires, and construct a set of a capabilities to emit at the end.

However whether that's desirable is another part, because as I understand, these capabilities affect platform availability, meaning that some platforms don't have certain capabilities. So unintentionally enabling them can be frustrating experience as your code will compile successfully to SPIR-V, but will fail when that SPIR-V is compiled on the target platform.

khyperia:
> @dragostis In theory, I believe so, I believe you probably could visit each instruction, see what capabilities it requires, and construct a set of a capabilities to emit at the end.

FWIW we kind of do that right now, and we're moving away from that. https://github.com/EmbarkStudios/rust-gpu/blob/908487ae1dcbc0a3bf119294a4f12dd7bc87c218/crates/rustc_codegen_spirv/src/linker/capability_computation.rs#L28-L30

(that file should also go away with this PR)

XAMPPRocky:
@khyperia Removing `capability_computation` is fine, however removing the defaults in `builder_spirv.rs` prevents `glam` from being compiled.

```
  error: f64 without OpCapability Float64
     --> /Users/erin.power/.rustup/toolchains/nightly-2021-03-21-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/num/f64.rs:640:5
      |
  640 |     pub fn recip(self) -> f64 {
      |     ^^^^^^^^^^^^^^^^^^^^^^^^^

  error: could not compile `glam`
```

khyperia:
> @khyperia Removing `capability_computation` is fine, however removing the defaults in `builder_spirv.rs` prevents `glam` from being compiled.

Yes, those systems in the compiler will need to be updated to the new capability computation system, in this case, my comment

> perhaps functions using `u64` are automatically declared with `needs_capability = Int64` implicitly, to avoid needing to modify libraries

in #42.

(Don't remove `capability_computation` without removing the defaults in `builder_spirv.rs`, the two complement each other, and unexpected capabilities will arise if you remove one without the other)

XAMPPRocky:
> (Don't remove capability_computation without removing the defaults in builder_spirv.rs, the two complement each other, and unexpected capabilities will arise if you remove one without the other)

Okay, well I think we should make removing those a separate PR, as they aren't incompatible with the new attribute, and it keeps both PRs smaller and easier to test.

khyperia:
> Okay, well I think we should make removing those a separate PR, as they aren't incompatible with the new attribute, and it keeps both PRs smaller and easier to test.

Sorry, I guess I was confused by the scope of this PR

XAMPPRocky:
> Sorry, I guess I was confused by the scope of this PR

No worries, I don't think there was a explicit scope when I opened, and this discussion helped clarified it. 😄

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -596,6 +600,249 @@ fn parse_image_type(
...
+        }
+    };
+
+    Ok(list
+        .iter()
+        .filter_map(NestedMetaItem::ident)
+        .filter_map(|i| {
+            CAPABILITIES
+                .iter()
+                .find(|x| Symbol::intern(x.0) == i.name)
```
khyperia, line 620: This should be cached and put in a HashMap, similar to how `const BUILTINS`/`Symbols::attributes` is done - interning is relatively expensive, especially in a O(n) find loop. The array should probably also be defined alongside the other large chunks of data, to consolidate them into something more manageable.

khyperia (changes requested):
I think `multiple.rs` and `invalid-target.rs` need to be updated with this. `not_entry_point.rs` can probably be folded into `invalid-target.rs`.

khyperia:
Going with https://github.com/EmbarkStudios/rust-gpu/issues/555 instead of attributes, so closing this

khyperia:
rip, race condition on closing it, hit the button literally the moment github flipped it to "reopen", haha

PR #533, opened 2021-03-24
Add support for multiple target environments

XAMPPRocky:
This PR adds support for multiple target environments such as `vulkan1.1`, and fixes our tests to pass `vulkan1.1` validation.

You can also now pass `--target-env` to compiletest to test SPIR-V environments.

```
cargo compiletest --target-env=vulkan1.1
# Or multiple
cargo compiletest --target-env=vulkan1.1,spv1.3
```

### DIFF
--- a/crates/rustc_codegen_spirv/src/lib.rs
+++ b/crates/rustc_codegen_spirv/src/lib.rs
@@ -211,7 +211,7 @@ fn is_blocklisted_fn<'tcx>(
     false
 }
 
-fn target_options() -> Target {
+fn target_options(env: Option<spirv_tools::TargetEnv>) -> Target {
     Target {
         llvm_target: "no-llvm".to_string(),
         pointer_width: 32,
@@ -228,6 +228,10 @@ fn target_options() -> Target {
             linker_flavor: LinkerFlavor::Ld,
             panic_strategy: PanicStrategy::Abort,
             os: "unknown".to_string(),
+            env: env
+                .as_ref()
+                .map(ToString::to_string)
+                .unwrap_or_else(|| "unknown".to_string()),
             // TODO: Investigate if main_needs_argc_argv is useful (for building exes)
             main_needs_argc_argv: false,
             ..Default::default()
@@ -290,12 +294,20 @@ impl CodegenBackend for SpirvCodegenBackend {
 
     fn target_override(&self, opts: &config::Options) -> Option<Target> {
         match opts.target_triple {
-            TargetTriple::TargetTriple(ref target_triple) => match &**target_triple {
-                // TODO: Do we want to match *everything* here, since, well, we only support one thing? that will allow
-                // folks to not specify --target spirv-unknown-unknown on the commandline.
-                "spirv-unknown-unknown" => Some(target_options()),
-                _ => None,
-            },
+            TargetTriple::TargetTriple(ref target_triple) => {
+                const ARCH_VENDOR: &str = "spirv-unknown-";
+                if !target_triple.starts_with(ARCH_VENDOR) {
+                    return None;
+                }
+
+                let env = &target_triple[ARCH_VENDOR.len()..];
+
+                match env.parse() {
+                    Ok(env) => Some(target_options(Some(env))),
+                    Err(_) if env == "unknown" => Some(target_options(None)),
+                    Err(_) => None,
+                }
+            }
             TargetTriple::TargetPath(_) => None,
         }
     }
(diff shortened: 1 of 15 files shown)


crates/rustc_codegen_spirv/src/lib.rs
```diff
@@ -284,12 +288,16 @@ impl CodegenBackend for SpirvCodegenBackend {
...
-                _ => None,
-            },
+            TargetTriple::TargetTriple(ref target_triple) => {
+                const ARCH_VENDOR: &str = "spirv-unknown-";
+                if !target_triple.starts_with(ARCH_VENDOR) {
+                    return None;
+                }
+
+                Some(target_options(
+                    target_triple[ARCH_VENDOR.len()..].parse().ok(),
```
khyperia, line 298: Should this fail, or produce some diagnostic (warning?) when it fails to parse? seems unfortunate to typo something and have it silently default to `unknown`

tests/ui/storage_class/push_constant.rs
```diff
@@ -7,6 +7,7 @@
 use spirv_std as _;
 
 #[derive(Copy, Clone)]
+#[spirv(block)]
```
eddyb, line 10: This test has a comment about how it should *only* pass as `spirv-unknown-unknown`, effectively.

I wonder if you can filter the target with a compiletest directive, or maybe add `#[cfg(target_env = "unknown")]` on the `struct` and the `fn`?

eddyb: Also there's a separate old test that you can now move into `tests/ui`, which tests the Vulkan validation side of this.

tests/src/main.rs
```diff
@@ -20,9 +20,22 @@ struct Opt {
     /// Only run tests that match these filters
     #[structopt(name = "FILTER")]
     filters: Vec<String>,
+
+    /// The environment to compile to the SPIR-V tests.
+    #[structopt(long)]
+    target_env: Option<String>,
```
eddyb, line 26: No structopt shorthand for comma-separated? Also, it might make more sense to always add new flags before `filters`, so that it matches the order they show up in the usage, but that's really just my pedantry.

tests/src/main.rs
```diff
@@ -58,90 +71,99 @@ fn main() {
...
-        &libs,
-    );
+        deps_target_dir,
+        codegen_backend_path,
+    };
+
+    runner.run_mode("ui");
 }
 
 // FIXME(eddyb) a bunch of these functions could be nicer if they were methods.
```
eddyb, line 86: At least one function is now a method, feel free to remove this comment or move it further down.

eddyb (changes requested):

tests/ui/storage_class/push_constant.rs
```diff
@@ -1,12 +1,10 @@
...
-// NOTE(eddyb) this won't pass Vulkan validation (see `push_constant_vulkan`),
-// but should still pass the baseline SPIR-V validation.
-
 // build-pass
-
 use spirv_std as _;
 
 #[derive(Copy, Clone)]
+// `Block` decoration is required for push constants when compiling for Vulkan.
+#[cfg_attr(target_env = "vulkan1.1", spirv(block))]
```
eddyb, line 7: Will this work for other environments? I suspect not, maybe check for `unknown` instead and omit `spirv(block)` in that case?

crates/spirv-builder/src/test/basic.rs
```diff
@@ -200,29 +200,6 @@ OpDecorate %4 Binding 0
     );
 }
 
-// NOTE(eddyb) we specifically run Vulkan validation here, as the default
-// validation rules are more lax and don't require a `Block` decoration
-// (`#[spirv(block)]` here) on `struct ShaderConstants`.
-#[test]
-fn push_constant_vulkan() {
-    val_vulkan(
```
eddyb, line 208: Can `val_vulkan` be removed now? There's some hacks IIRC, between `rustc_codegen_spirv` and `spirv-builder`, just to make it work.

eddyb (changes requested):

crates/spirv-builder/src/test/mod.rs
```diff
@@ -99,19 +99,6 @@ fn val(src: &str) {
     build(src);
 }
 
-/// While `val` runs baseline SPIR-V validation, for some tests we want the
-/// stricter Vulkan validation (`vulkan1.2` specifically), which may produce
-/// additional errors (such as missing Vulkan-specific decorations).
-fn val_vulkan(src: &str) {
-    use rustc_codegen_spirv::{spirv_tools_validate as validate, SpirvToolsTargetEnv as TargetEnv};
```
eddyb, line 106: Sorry to keep pulling on this, but can you remove these two exports from `rustc_codegen_spirv`?

eddyb (approved):

PR #547, opened 2021-03-29
[experiment] #546 + "ci: run `cargo test` with `--release`".

eddyb:
This is effectively a repeat of (the first iteration of #494), on top of #546 (also see its description).
Only opening to see what CI does with it and how long it takes.

### DIFF
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -12,6 +12,7 @@ os=$1
 function cargo_test() {
     echo ::group::"$1 build"
     cargo test \
+        --release \
         --manifest-path "$1/Cargo.toml" \
         --no-default-features \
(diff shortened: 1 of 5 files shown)


eddyb:
Alright, so the reason this is slower, despite being faster in parts, is that all of the "host code" of the runners (including dependencies, basically everything in the dependency tree that `rustc_codegen_spirv` doesn't reach) takes longer to optimize - however, it's also the part we don't care about optimizing, so there may be a way around this, in the form of making debug mode itself be optimized and using *that*.

khyperia:
I think some of this got merged with https://github.com/EmbarkStudios/rust-gpu/pull/637, so closing this for now.

PR #551, opened 2021-03-29
Add support for outputting multiple modules, one per entry point

### LINKED ISSUE #539: Add option to output multiple files, one per entry point
Discussed in https://github.com/EmbarkStudios/rust-gpu/issues/524

Right now, the ecosystem for SPIR-V is quite buggy when handling multiple entry points per spir-v module. We've fought it a lot, but it's time to give up :fire: and add an option to emit multiple files per crate, one entry point per file.

1) Add an option to spirv-builder to let the user specify that they want multiple files
2) Plumb that through to rust-gpu
3) Near the end of linking (before ID compaction? idk), clone the module N times, where N is the number of entry points. In each one, remove all but one entry point, then run [DCE](https://github.com/EmbarkStudios/rust-gpu/blob/c3757cb9c799872ec25b5ea835dcd4f0ff9e8941/crates/rustc_codegen_spirv/src/linker/dce.rs) on it.
4) Output the collection of files somewhere(?). @eddyb said that there's pre-existing code to translate a rust name (e.g. `foo::bar`) to a filesystem-okay path in MIR dumping, we should re-use that.
5) Return a `HashMap<function name, Path>` from spirv-builder to the user :sparkles:somehow:sparkles:. @eddyb had the idea of making the rustc artifact actually be a json file that contains that map, and then spirv-builder would read that json and return that map.

khyperia:
Fixes #539

Not sure if I should actually include the `examples/multibuilder` crate, I'm fine with deleting it (also for some reason it only works when cd'ing into the dir and running `cargo run --release`, it fails to find the `cargo` exe when running with `cargo run --release -p multibuilder`)


khyperia:
thanks, my brain is mush right now :P

VZout (approved):
Lets keep the example. in the future we might want to use a toggle for that in the examples tho 🤷

crates/rustc_codegen_spirv/src/lib.rs
```diff
@@ -380,6 +380,9 @@ impl CodegenBackend for SpirvCodegenBackend {
     ) -> Result<(), ErrorReported> {
         // TODO: Can we merge this sym with the one in symbols.rs?
         let legalize = !sess.target_features.contains(&Symbol::intern("kernel"));
+        let emit_multiple_modules = sess
+            .target_features
+            .contains(&Symbol::intern("multimodule"));
```
XAMPPRocky, line 385: Would it be more appropriate if we accepted this as an argument from `-Cllvm-args`([`sess.cg.llvm_args`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_session/config/struct.CodegenOptions.html#structfield.llvm_args))? It's not really something that's specific to the the target as it is specific to our backend. Another benefit is since it is just `Vec<String>`, we could use clap here for getting these arguments, that's make it to have key value arguments like `--module-output=[multiple|single]`.

XAMPPRocky (approved):

PR #552, opened 2021-03-30
Allow #[spirv(block)] on DSTs

Hentropy:
Types decorated with Block may be unsized.


eddyb (approved):

eddyb:
I should've asked you to write a test but I thought we should just merge it because I was already doing that.

Turns out it's not that useful:
`error: invalid id:0:0 - For Vulkan, an OpTypeStruct variable containing an OpTypeRuntimeArray must be decorated with BufferBlock if it has storage class Uniform.`

PR #556, opened 2021-03-30
Run DCE twice

### LINKED ISSUE #531: Dead types not stripped
In SPIR-V output, one can find various types that should not remain after optimization and dead code elimination, that are actually invalid and will trip up some drivers. 

This can be things like:
```
%_ptr_Function__ptr_Input_v3float = OpTypePointer Function %_ptr_Input_v3float
```
but I think there are other variations too.

These violate the following restriction:

```
A variable pointer must point to one of the following storage classes:
* StorageBuffer
* Workgroup (if the VariablePointers capability is declared
```

Unfortunately these do seem to pass spir-val, so that's a separate problem.

khyperia:
Fixes #531

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/mod.rs
+++ b/crates/rustc_codegen_spirv/src/linker/mod.rs
@@ -277,7 +277,10 @@ pub fn link(sess: &Session, mut inputs: Vec<Module>, opts: &Options) -> Result<L
                 .write_all(spirv_tools::binary::from_binary(&output.assemble()))
                 .unwrap();
         }
-        if opts.dce && opts.emit_multiple_modules {
+        // Run DCE again, even if emit_multiple_modules==false - the first DCE ran before
+        // structurization and mem2reg (for perf reasons), and mem2reg may remove references to
+        // invalid types, so we need to DCE again.
+        if opts.dce {
             let _timer = sess.timer("link_dce_2");
             dce::dce(output);
         }
--- a/crates/spirv-builder/src/test/basic.rs
+++ b/crates/spirv-builder/src/test/basic.rs
@@ -189,15 +189,14 @@ OpDecorate %5 Binding 0
 %6 = OpTypeVoid
 %7 = OpTypeFunction %6
 %8 = OpTypeInt 32 0
-%9 = OpTypePointer Function %8
-%10 = OpConstant %8 1
-%11 = OpTypeFloat 32
-%12 = OpTypeImage %11 2D 0 0 0 1 Unknown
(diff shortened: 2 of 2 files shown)


XAMPPRocky (approved):

khyperia:
Disabled automerge because I'd like @eddyb's feedback on if the last part of this comment is correct before merging this https://github.com/EmbarkStudios/rust-gpu/issues/531#issuecomment-809948120

crates/rustc_codegen_spirv/src/linker/mod.rs
```diff
@@ -277,7 +277,10 @@ pub fn link(sess: &Session, mut inputs: Vec<Module>, opts: &Options) -> Result<L
                 .write_all(spirv_tools::binary::from_binary(&output.assemble()))
                 .unwrap();
         }
-        if opts.dce && opts.emit_multiple_modules {
+        // Run DCE again, even if emit_multiple_modules==false - the first DCE ran before
+        // structurization and mem2reg (for perf reasons), and mem2reg may remove references to
+        // invalid types, so we need to DCE again.
```
eddyb, line 282: Not sure if you want to remove the comment or tweak it, but DCE is really cheap, and removing dead code makes other passes faster (especially `mem2reg` IIUC).

For reference, I just checked on `mouse-shader` (built during the build script of `cargo build --release -p example-runner-wgpu`) and this PR gets:
```
time:   0.008; rss:  183MB ->  183MB (   +0MB)  link_dce
time:   0.000; rss:  183MB ->  183MB (   +0MB)  link_structurize
time:   0.297; rss:  183MB ->  183MB (   +0MB)  link_block_ordering_pass_and_mem2reg
time:   0.000; rss:  183MB ->  183MB (   +0MB)  link_sort_globals
time:   0.002; rss:  183MB ->  183MB (   +0MB)  link_dce_2
```
Whereas if I remove the first DCE it pretty much takes twice as long in `mem2reg`:
```
time:   0.596; rss:  183MB ->  183MB (   +0MB)  link_block_ordering_pass_and_mem2reg
time:   0.000; rss:  183MB ->  183MB (   +0MB)  link_sort_globals
time:   0.008; rss:  183MB ->  183MB (   +0MB)  link_dce_2
```

eddyb (approved):

crates/rustc_codegen_spirv/src/linker/mod.rs
khyperia: thanks for the timing numbers! haven't looked at these in far too long, forgot how they looked~ :heart:

PR #559, opened 2021-03-30
Add new handling for SPIR-V targets

XAMPPRocky:
This PR adds a target argument that you can set a different target environment from `spirv-unknown-unknown` with spirv-builder.


khyperia (approved):
Approved pending comment

examples/runners/wgpu/build.rs
```diff
@@ -4,6 +4,7 @@ use std::error::Error;
 fn build_shader(path_to_create: &str) -> Result<(), Box<dyn Error>> {
     SpirvBuilder::new(path_to_create)
         .spirv_version(1, 0)
+        .target("spirv-unknown-vulkan1.1")
```
khyperia, line 7: whoops, didn't hit "submit comment", I thought wgpu used vulkan 1.0?

khyperia:
Also, we should think about deriving `spirv_version` from the target triple in the compiler, instead of having it be a separate option

XAMPPRocky:
> Also, we should think about deriving spirv_version from the target triple in the compiler, instead of having it be a separate option

Is that possible? I don't think cargo exposes the target you're compiling for to the build script. Otherwise then we should probably always do that.

Eventually when we can set this in Cargo with artifact deps it would be just part of the dependency.
```toml
sky-shader = { path = "./sky-shader", target = "spirv-unknown-vulkan1.1" }
```

examples/runners/wgpu/build.rs
XAMPPRocky: > I thought wgpu used vulkan 1.0?

I don't know, really. I tried googling to see if there was but couldn't find anything that mentions a minimum compatible version. I did run the wgpu example, so it definitely works at least.

khyperia: Currently the target doesn't actually affect the binary, only which validator is run (AFAIK), so specifying vulkan1.1 here will still produce a vulkan1.0 binary (due to `.spirv_version(1, 0)`), so it'll run on wgpu. It should still be `vulkan1.0`, though, I think.

XAMPPRocky:
> Also, we should think about deriving spirv_version from the target triple in the compiler, instead of having it be a separate option

How would we account for `unknown`? Do we have a default, or should we remove `unknown` and require people to always specify a target environment?

khyperia:
> How would we account for `unknown`? Do we have a default, or should we remove `unknown` and require people to always specify a target environment?

I think we should remove `unknown`. If someone objects, though, we currently default to 1.3 if `spirv_version` isn't specified, so defaulting to `spirv-unknown-spv1.3` is probably reasonable (might be nice to default in spirv-builder instead of the compiler, too)

examples/multibuilder/src/main.rs
```diff
@@ -1,7 +1,7 @@
 use spirv_builder::SpirvBuilder;
 
 fn main() {
-    let result = SpirvBuilder::new("../shaders/sky-shader")
+    let result = SpirvBuilder::new("../shaders/sky-shader", "spirv-unknown-spv1.3")
```
VZout, line 4: Any chance we can turn this into an enum?

XAMPPRocky: Well the environment is already an enum, but I was deliberately keeping this stringly typed as providing the whole target triple is more flexible, as it allows you to put any target here. So if there was an `spirv-embark-spv1.3` target for example it would be valid.

examples/runners/wgpu/build.rs
```diff
@@ -2,7 +2,7 @@ use spirv_builder::SpirvBuilder;
 use std::error::Error;
 
 fn build_shader(path_to_create: &str) -> Result<(), Box<dyn Error>> {
-    SpirvBuilder::new(path_to_create)
+    SpirvBuilder::new(path_to_create, "spirv-unknown-vulkan1.1")
```
XAMPPRocky, line 5:
```suggestion
    SpirvBuilder::new(path_to_create, "spirv-unknown-vulkan1.0")
```

examples/runners/ash/src/main.rs
```diff
@@ -171,10 +171,13 @@ pub fn main() {
 }
 
 pub fn compile_shaders() -> Vec<SpvFile> {
-    let spv_paths: Vec<PathBuf> = vec![SpirvBuilder::new("examples/shaders/sky-shader")
-        .print_metadata(false)
-        .build()
-        .unwrap()];
+    let spv_paths: Vec<PathBuf> =
+        vec![
+            SpirvBuilder::new("examples/shaders/sky-shader", "spirv-unknown-vulkan1.1")
```
XAMPPRocky, line 176:
```suggestion
            SpirvBuilder::new("examples/shaders/sky-shader", "spirv-unknown-vulkan1.0")
```

khyperia: The ash runner uses vulkan1.1, so this should stay vulkan 1.1. https://github.com/EmbarkStudios/rust-gpu/blob/8c89b49b6639fed7e8ef3c13b5c17c2b472ae680/examples/runners/ash/src/main.rs#L255

crates/rustc_codegen_spirv/src/target.rs
```diff
@@ -0,0 +1,144 @@
...
+            return Err(error());
+        }
+
+        let vendor = iter.next().map(From::from).ok_or_else(error)?;
+
+        let env = iter
+            .next()
+            .and_then(|env| env.parse().ok())
+            .ok_or_else(error)?;
+        Ok(Self { env, vendor })
```
khyperia, line 126: Checking `iter.next() == None` would be good, otherwise `spirv-unknown-spv1.0-shenanigans` would parse as valid

khyperia (approved):

PR #561, opened 2021-03-30
structurizer: find and/or cache `OpConstant{False,True}`.

eddyb:


crates/rustc_codegen_spirv/src/linker/new_structurizer.rs
```diff
@@ -37,6 +44,32 @@ pub fn structurize(
...
+        }
+
+        if existing_const_false.is_some() && existing_const_true.is_some() {
+            break;
+        }
+    }
+
+    // Create new `OpConstant{False,True}` if they're missing.
+    let const_false = existing_const_false.unwrap_or_else(|| builder.constant_false(type_bool));
+    let const_true = existing_const_true.unwrap_or_else(|| builder.constant_true(type_bool));
```
khyperia, line 71: ah, was going to comment on this, but with the new DCE pass happening after the structurizer, this is fine if these aren't used anywhere :P

khyperia (approved):

PR #563, opened 2021-03-31
Add support for SPV_KHR_ray_tracing

XAMPPRocky:
This PR adds initial support for ray tracing in Rust-GPU.

~~Currently the tests don't pass because I ran into some ICE's when testing `OpTerminateRayKHR`. Filed as https://github.com/EmbarkStudios/rust-gpu/issues/564~~


XAMPPRocky:
cc @hrydgard @repi

repi:
Cool!

hrydgard:
Very cool!

To actually do real shading once we hit stuff with rays though, we need to be able to look up mesh data by mesh ID, which will likely end up requiring some of https://github.com/EmbarkStudios/rust-gpu/pull/450 .

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -129,21 +129,21 @@ const BUILTINS: &[(&str, BuiltIn)] = {
...
-        ("ray_tmax_nv", RayTmaxNV),
-        ("instance_custom_index_nv", InstanceCustomIndexNV),
-        ("object_to_world_nv", ObjectToWorldNV),
-        ("world_to_object_nv", WorldToObjectNV),
-        ("hit_t_nv", HitTNV),
-        ("hit_kind_nv", HitKindNV),
-        ("incoming_ray_flags_nv", IncomingRayFlagsNV),
+        ("launch_id_khr", LaunchIdNV),
+        ("launch_size_khr", LaunchSizeNV),
+        ("instance_custom_index_khr", InstanceCustomIndexNV),
```
hrydgard, line 134: khr on left side, NV on right? were some NV names reused in the KHR extension?

XAMPPRocky: It's just because rspirv has the NV versions but not the KHR versions.

expenses: Probably best to wait for an rspirv update then. The NV things are _probably_ the same as the KHR ones but it's best no to mix them up. Looking foward to this though!

khyperia: Huh? rspirv has the KHR versions already though https://docs.rs/spirv_headers/1.5.0/spirv_headers/enum.BuiltIn.html#associatedconstant.LaunchIdKHR

XAMPPRocky: Ah, I figured it out. They're not variants but constants. So you can't use `use Builtin::*;` and then say `LaunchIdKHR` because it's not imported with the variants. I've updated the PR to use those constants.

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -129,21 +129,21 @@ const BUILTINS: &[(&str, BuiltIn)] = {
...
+        ("world_ray_direction_khr", BuiltIn::WorldRayDirectionKHR),
+        ("object_ray_origin_khr", BuiltIn::ObjectRayOriginKHR),
+        ("object_ray_direction_khr", BuiltIn::ObjectRayDirectionKHR),
+        ("ray_tmin_khr", BuiltIn::RayTminKHR),
+        ("ray_tmax_khr", BuiltIn::RayTmaxKHR),
+        ("object_to_world_khr", BuiltIn::ObjectToWorldKHR),
+        ("world_to_object_khr", BuiltIn::WorldToObjectKHR),
+        ("hit_kind_khr", BuiltIn::HitKindKHR),
+        ("incoming_ray_flags_khr", BuiltIn::IncomingRayFlagsKHR),
         ("hit_t_nv", HitTNV),
```
hrydgard, line 146: Shouldn't we just get rid of the NV raytracing stuff entirely in favor of KHR? 

There's little reason for anyone to use it for new software, KHR is the only future.

XAMPPRocky: Maybe? `HitTNV` seems to be the only builtin that wasn't brought over to the KHR extension. So there might be valid uses for it, but I don't know enough about it to be able to say one way or another.

expenses: Supposedly it's just an alias of a variable that is in KHR:
https://github.com/KhronosGroup/GLSL/blob/ab928c80851b96f3889f1287d4ce0003a9bb86e4/extensions/nv/GLSL_NV_ray_tracing.txt#L657-L658

hrydgard: Yeah it's nothing special and is only part of the NV extension, which can now be considered deprecated.

hrydgard: Please just delete it along with the other NV stuff.

tests/ui/arch/trace_ray_khr.rs
```diff
@@ -2,7 +2,7 @@
 
 #[spirv(ray_generation_khr)]
 pub fn main(
-    acceleration_structure: &spirv_std::ray_tracing::AccelerationStructureKhr,
+    #[spirv(uniform, descriptor_set = 0, binding = 0)] acceleration_structure: &spirv_std::ray_tracing::AccelerationStructureKhr,
```
hrydgard, line 5: wasn't uniform in this position only for uniform buffers?

crates/rustc_codegen_spirv/src/symbols.rs
XAMPPRocky: Well, I've removed `HitTNV`, I'm going to leave any other pre-existing NV builtins however, as the rest are unrelated to ray tracing, and should be migrated separately as their own PRs.

tests/ui/arch/trace_ray_khr.rs
XAMPPRocky: Yep, I was just rushing to fix it and didn't type `_constant`, it should be `uniform_constant` but that shouldn't be needed once #567 is merged.

crates/rustc_codegen_spirv/Cargo.toml
```diff
@@ -37,7 +37,7 @@ syn = { version = "1", features = ["visit", "visit-mut"] }
 # Normal dependencies.
 bimap = "0.6"
 indexmap = "1.6.0"
-rspirv = { git = "https://github.com/gfx-rs/rspirv.git", rev = "ee1e913" }
+rspirv = { git = "https://github.com/EmbarkStudios/rspirv.git", rev = "7c72c65" }
```
khyperia, line 40: Marking this as request-change just to make sure we don't merge this (it's okay for a draft, but we shouldn't merge it)

crates/spirv-std/src/ray_tracing.rs
```diff
@@ -0,0 +1,9 @@
+//! Ray-tracing data types
+
+/// An acceleration structure type which is an opaque reference to an
+/// acceleration structure handle as defined in the client API specification.
+#[spirv(acceleration_structure_khr)]
+#[derive(Copy, Clone)]
+pub struct AccelerationStructureKhr {
+    pub(crate) _private: (),
```
khyperia, line 8: I'm a little worried about weird edge cases with rustc thinking it's a ZST, I think I'd prefer this to be a `u32` like all the other fake types

khyperia (changes requested):

crates/spirv-std/src/ray_tracing.rs
XAMPPRocky: Shouldn't it be treated as a ZST though? The spec says it has no defined size or bit pattern. I'm fine with changing it though.

khyperia: "no defined size" is very different than "zero size" - it's the difference between `[T]` and `[T; 0]`. Images are also "no defined size or bit pattern", and we use u32 for those.

crates/spirv-std/src/arch/ray_tracing.rs
```diff
@@ -0,0 +1,174 @@
...
+            {ray_direction} \
+            {ray_tmax} \
+            {payload}",
+        acceleration_structure = in(reg) &acceleration_structure,
+        ray_flags = in(reg) ray_flags,
+        cull_mask = in(reg) cull_mask,
+        sbt_offset = in(reg) sbt_offset,
+        sbt_stride = in(reg) sbt_stride,
+        miss_index = in(reg) miss_index,
+        ray_origin = in(reg) &ray_origin,
```
khyperia, line 86: ... hold on, looking up the docs for this, `ray_origin` should be a straight vector, not a pointer to a vector, yet validation succeeds. I'm scared now, spirv-val probably doesn't validate this instruction at all, and we should probably throw this at an actual implementation to be sure?

XAMPPRocky: Oh yeah definitely, I don't think there's much validation of ray tracing instructions happening at all.

crates/spirv-std/src/arch/ray_tracing.rs
```diff
@@ -0,0 +1,175 @@
...
+#[inline]
+pub unsafe fn convert_u_to_acceleration_structure_khr(id: u64) -> AccelerationStructureKhr {
+    // Since we can't represent an uninitalized opaque type in Rust at the
+    // moment, we need to create and return the acceleration structure entirely
+    // in assembly.
+    asm! {
+        "%ret = OpTypeAccelerationStructureKHR",
+        "%result = OpConvertUToAccelerationStructureKHR %ret {id}",
+        "OpReturnValue %result",
+        "%blah = OpLabel",
```
Jasper-Bekkers, line 18: Small nit :-)

XAMPPRocky: Nit that it's called blah or that it's there at all? The `OpLabel` has to be there because currently we don't correctly handle terminators in inline assembly. https://github.com/EmbarkStudios/rust-gpu/issues/564

Jasper-Bekkers: Yeah just about the name.

crates/spirv-std/src/arch/ray_tracing.rs
```diff
@@ -0,0 +1,81 @@
...
+#[inline]
+pub unsafe fn report_intersection(hit: f32, hit_kind: u32) -> bool {
+    let mut result: u8 = 0;
+
+    asm! {
+        "%bool = OpTypeBool",
+        "%u8 = OpTypeInt 8 0",
+        "%result = OpReportIntersectionKHR %u8 {hit} {hit_kind}",
+        "OpStore {result} %result",
+        result = in(reg) &mut result,
```
khyperia, line 29: If the result is a u8, it can be an `out(reg)`, right? No need to do the default value and `&mut` dance.

crates/spirv-std/src/arch/ray_tracing.rs
```diff
@@ -0,0 +1,81 @@
...
+#[spirv_std_macros::gpu_only]
+#[doc(alias = "OpReportIntersectionKHR")]
+#[inline]
+pub unsafe fn report_intersection(hit: f32, hit_kind: u32) -> bool {
+    let mut result: u8 = 0;
+
+    asm! {
+        "%bool = OpTypeBool",
+        "%u8 = OpTypeInt 8 0",
+        "%result = OpReportIntersectionKHR %u8 {hit} {hit_kind}",
```
khyperia, line 27: Result type should be a `%bool` - from the spec:

> Result Type must be a scalar boolean.

crates/spirv-std/src/arch/ray_tracing.rs
```diff
@@ -0,0 +1,81 @@
...
+#[spirv_std_macros::gpu_only]
+#[doc(alias = "OpExecuteCallableKHR")]
+#[inline]
+pub unsafe fn execute_callable<T, const ID: usize>(data: &T) {
+    asm! {
+        "OpExecuteCallableKHR {id} {data}",
+        id = const ID,
+        data = in(reg) data,
+    };
+}
```
khyperia, line 81: I don't think there's a test for this function

khyperia (changes requested):

crates/spirv-std/src/arch/ray_tracing.rs
XAMPPRocky: Yeah it should, but it doesn't seem to work. Using bool seems to insert an `OpINotEqual` instruction, which causes it error.

```rust
pub unsafe fn report_intersection(hit: f32, hit_kind: u32) -> bool {
    let result: u8;

    asm! {
        "%bool = OpTypeBool",
        "%u8 = OpTypeInt 8 0",
        "{result} = OpReportIntersectionKHR %bool {hit} {hit_kind}",
        result = out(reg) result,
        hit = in(reg) hit,
        hit_kind = in(reg) hit_kind,
    };

    result != 0
}
```

```spirv
; SPIR-V
; Version: 1.3
; Generator: Google rspirv; 0
; Bound: 17
; Schema: 0
               OpCapability Shader
               OpCapability Int8
               OpCapability RayTracingKHR
               OpExtension "SPV_KHR_ray_tracing"
               OpMemoryModel Logical Simple
               OpEntryPoint IntersectionNV %1 "main"
               OpName %report_intersection_khr__main "report_intersection_khr::main"
       %bool = OpTypeBool
      %float = OpTypeFloat 32
       %uint = OpTypeInt 32 0
      %uchar = OpTypeInt 8 0
    %uchar_0 = OpConstant %uchar 0
       %void = OpTypeVoid
          %9 = OpTypeFunction %void
    %float_2 = OpConstant %float 2
     %uint_4 = OpConstant %uint 4
%report_intersection_khr__main = OpFunction %void None %9
         %12 = OpLabel
         %13 = OpReportIntersectionKHR %bool %float_2 %uint_4
         %14 = OpINotEqual %bool %13 %uchar_0
               OpReturn
               OpFunctionEnd
          %1 = OpFunction %void None %9
         %15 = OpLabel
         %16 = OpFunctionCall %void %report_intersection_khr__main
               OpReturn
               OpFunctionEnd
```

khyperia: I'm confused, the spec isn't something you can ignore and make it be a u8 instead, the spec states

> Result Type **must** be a scalar boolean.

maybe I shouldn't have said "should", it _must_ be a boolean.

khyperia: This seems to work just fine

```rust
pub unsafe fn report_intersection(hit: f32, hit_kind: u32) -> bool {
    let result: u32;

    asm! {
        "%bool = OpTypeBool",
        "%u32 = OpTypeInt 32 0",
        "%zero = OpConstant %u32 0",
        "%one = OpConstant %u32 0",
        "%result = OpReportIntersectionKHR %bool {hit} {hit_kind}",
        "{result} = OpSelect %u32 %result %one %zero",
        result = out(reg) result,
        hit = in(reg) hit,
        hit_kind = in(reg) hit_kind,
    };

    result != 0
}
```

khyperia (approved):
LGTM, I'd also like a review from @eddyb

crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
```diff
@@ -727,15 +727,15 @@ pub fn instruction_signatures(op: Op) -> Option<&'static [InstSig<'static>]> {
         | Op::ExecuteCallableKHR
         | Op::ConvertUToAccelerationStructureKHR
         | Op::IgnoreIntersectionKHR
-        | Op::TerminateRayKHR => reserved!(SPV_KHR_ray_tracing),
+        | Op::TerminateRayKHR => {}
```
eddyb, line 730: Can you change these to be like the other ones which have a comment like this:
```rust
// NOTE(eddyb) we actually use these despite not being in the standard yet.
```
(and the `reserved!` being commented out, not removed)

The reason I want to keep this structure is that the spec still lists these instructions are reserved, and I wanted to be able to track changes to the spec long-term, by mirroring its order and categorization.

crates/rustc_codegen_spirv/src/linker/new_structurizer.rs
```diff
@@ -168,7 +168,12 @@ impl Structurizer<'_> {
             let block_id = self.func.blocks()[block].label_id().unwrap();
             let terminator = self.func.blocks()[block].instructions.last().unwrap();
             let mut region = match terminator.class.opcode {
-                Op::Return | Op::ReturnValue | Op::Kill | Op::Unreachable => Region {
+                Op::Return
+                | Op::ReturnValue
+                | Op::Kill
+                | Op::IgnoreIntersectionKHR
+                | Op::TerminateRayKHR
+                | Op::Unreachable => Region {
```
eddyb, line 176: There's a comment in this file that goes like this:
```rust
    /// The `merge` block must be terminated by one of `OpReturn`, `OpReturnValue`,
    /// `OpKill`, or `OpUnreachable`.
```

Can you update it, or at least make the phrasing less complete (i.e. turn those specific instructions into examples more than an exhaustive list)?

eddyb (changes requested):

eddyb (approved):

crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
```diff
@@ -754,7 +754,10 @@ pub fn instruction_signatures(op: Op) -> Option<&'static [InstSig<'static>]> {
         | Op::TerminateRayNV
         | Op::TraceNV
         | Op::TypeAccelerationStructureNV
-        | Op::ExecuteCallableNV => reserved!(SPV_NV_ray_tracing),
+        | Op::ExecuteCallableNV => {
+            // NOTE(eddyb) Some KHR variants are aliased to the the NV instructions.
+            // reserved!(SPV_KHR_ray_tracing)
```
eddyb, line 759: Wrong name, should be `SPV_NV_ray_tracing`.

eddyb (changes requested):

crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
```diff
@@ -748,7 +754,10 @@ pub fn instruction_signatures(op: Op) -> Option<&'static [InstSig<'static>]> {
...
```
XAMPPRocky, line 759:
```suggestion
            // reserved!(SPV_NV_ray_tracing)
```

eddyb (approved):

PR #567, opened 2021-03-31
Infer the storage class for images/samplers/imagesamplers

khyperia:


XAMPPRocky (approved):
LGTM! Do we also want to remove the `uniform_constant` attribute? Are there reasons to keep it?

khyperia:
Not sure if it's used anywhere else, but if we do remove it, then we should also remove `input`/`output`/etc. (right now it's a consistent "include everything" thing, even things that aren't currently valid to place anywhere)

XAMPPRocky:
> Not sure if it's used anywhere else, but if we do remove it, then we should also remove input/output/etc. (right now it's a consistent "include everything" thing, even things that aren't currently valid to place anywhere)

Sound good to me. I'll file a separate issue for that.

Also the tests need formatting 🙃

khyperia:
haha yeah, kinda sucks that `cargo fmt` doesn't catch the tests, have to run `rustfmt tests/ui/**/*.rs`, but whatever, at least a check is in CI~

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -264,25 +269,67 @@ impl<'tcx> CodegenCx<'tcx> {
...
+        hir_param: &hir::Param<'tcx>,
+        decoration_locations: &mut HashMap<StorageClass, u32>,
+    ) -> (Word, StorageClass) {
+        let attrs = AggregatedSpirvAttributes::parse(self, self.tcx.hir().attrs(hir_param.hir_id));
+
+        // FIXME(eddyb) attribute validation should be done ahead of time.
+        // FIXME(eddyb) also take into account `&T` interior mutability,
+        // i.e. it's only immutable if `T: Freeze`, which we should check.
+        // FIXME(eddyb) also check the type for compatibility with being
+        // part of the interface, including potentially `Sync`ness etc.
```
eddyb, line 330: These comments belong in `infer_param_ty_and_storage_class` (they pertain to the `TyKind::Ref` part of the Rust type, more or less).

eddyb (changes requested):

eddyb:
> Not sure if it's used anywhere else, but if we do remove it, then we should also remove `input`/`output`/etc. (right now it's a consistent "include everything" thing, even things that aren't currently valid to place anywhere)

I've been told OpenGL uses `UniformConstant` the same way Vulkan uses `PushConstant` (which explains the name).

Removing `input`/`output` and `function`/`private`/`generic` should probably be done regardless.

eddyb (approved):

PR #572, opened 2021-03-31
Add support for SPV_KHR_ray_query

XAMPPRocky:
This PR adds initial support for `SPV_KHR_ray_query` in rust-gpu. The only current exception is `OpRayQueryGetIntersectionObjectToWorldKHR` and `OpRayQueryGetIntersectionWorldToObjectKHR`, because we don't have a `Matrix` trait yet, and I wanted to think about how that would work more.

Depends on #563 since it needs `AccelerationStructure` for some functions.

cc @hrydgard @repi

### DIFF
--- a/crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
+++ b/crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
@@ -811,7 +811,10 @@ pub fn instruction_signatures(op: Op) -> Option<&'static [InstSig<'static>]> {
         | Op::RayQueryGetWorldRayDirectionKHR
         | Op::RayQueryGetWorldRayOriginKHR
         | Op::RayQueryGetIntersectionObjectToWorldKHR
-        | Op::RayQueryGetIntersectionWorldToObjectKHR => reserved!(SPV_KHR_ray_query),
+        | Op::RayQueryGetIntersectionWorldToObjectKHR => {
+            // NOTE(eddyb) we actually use these despite not being in the standard yet.
+            // reserved!(SPV_KHR_ray_query)
+        }
 
         // Instructions not present in current SPIR-V specification
         // SPV_INTEL_function_pointers
--- a/crates/spirv-std/src/ray_tracing.rs
+++ b/crates/spirv-std/src/ray_tracing.rs
@@ -159,3 +159,492 @@ bitflags::bitflags! {
         const SKIP_AABBS = 512;
     }
 }
+
+/// A ray query type which is an opaque object representing a ray traversal.
+#[spirv(ray_query)]
+pub struct RayQuery {
+    _private: u32,
+}
+
+/// Constructs an uninitialized ray query variable. Using the syntax
+/// `let (mut)? <name>`. Where `name` is the name of the ray query variable.
+#[macro_export]
+macro_rules! ray_query {
+    (let $name:ident) => {
+        $crate::ray_query!(@inner $name)
+    };
+    (let mut $name:ident) => {
+        $crate::ray_query!(@inner $name, mut)
+    };
+    (@inner $name:ident $(, $mut:tt)?) => {
+        let $name: &$($mut)? RayQuery = unsafe {
+            let $name : *mut RayQuery;
+            asm! {
+                "%ray_query = OpTypeRayQueryKHR",
+                "%ray_query_ptr = OpTypePointer Generic %ray_query",
+                "{name} = OpVariable %ray_query_ptr Function",
+                name = out(reg) $name,
+            }
+
+            &$($mut)? *$name
+        };
+    }
+}
+
+impl RayQuery {
+    /// Initialize a ray query object, defining parameters of traversal. After this
+    /// call, a new ray trace can be performed with [`Self::proceed`]. Any
+    /// previous traversal state stored in the object is lost.
+    ///
+    /// - `ray_query` is a pointer to the ray query to initialize.
+    /// - `acceleration_structure` is the descriptor for the acceleration structure
+    ///   to trace into.
+    /// - `ray_flags` contains one or more of the Ray Flag values.
+    /// - `cull_mask` is the mask to test against the instance mask.  Only the 8
+    ///   least-significant bits of `cull_mask` are used by this instruction - other
+    ///   bits are ignored.
+    /// - `ray_origin`, `ray_tmin`, `ray_direction`, and `ray_tmax` control the
+    ///   basic parameters of the ray to be traced.
+    #[spirv_std_macros::gpu_only]
+    #[doc(alias = "OpRayQueryInitializeKHR")]
+    #[inline]
+    #[allow(clippy::too_many_arguments)]
+    pub unsafe fn initialize(
+        &mut self,
+        acceleration_structure: &AccelerationStructure,
+        ray_flags: RayFlags,
+        cull_mask: u32,
+        ray_origin: impl Vector<f32, 3>,
+        ray_tmin: f32,
+        ray_direction: impl Vector<f32, 3>,
+        ray_tmax: f32,
+    ) {
+        asm! {
+            "%origin = OpLoad _ {ray_origin}",
+            "%direction = OpLoad _ {ray_direction}",
+            "OpRayQueryInitializeKHR \
+                {ray_query} \
+                {acceleration_structure} \
+                {ray_flags} \
+                {cull_mask} \
+                %origin \
+                {ray_tmin} \
+                %direction \
+                {ray_tmax}",
+            ray_query = in(reg) self,
+            acceleration_structure = in(reg) acceleration_structure,
+            ray_flags = in(reg) ray_flags.bits(),
+            cull_mask = in(reg) cull_mask,
+            ray_origin = in(reg) &ray_origin,
+            ray_tmin = in(reg) ray_tmin,
+            ray_direction = in(reg) &ray_direction,
+            ray_tmax = in(reg) ray_tmax,
+        }
+    }
+
+    /// Allow traversal to proceed. Returns `true` if traversal is incomplete,
+    /// and `false` when it has completed. A previous call to [`Self::proceed`]
+    /// with the same ray query object must not have already returned `false`.
+    #[spirv_std_macros::gpu_only]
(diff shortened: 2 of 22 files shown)


crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -179,7 +179,8 @@ impl<'tcx> CodegenCx<'tcx> {
             SpirvType::Image { .. }
             | SpirvType::Sampler
             | SpirvType::SampledImage { .. }
-            | SpirvType::AccelerationStructureKhr => {
+            | SpirvType::AccelerationStructureKhr
+            | SpirvType::RayQueryKhr => {
```
eddyb, line 183: I don't think it should be added here, since it's invalid to have in the interface.

crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
```diff
@@ -811,7 +811,7 @@ pub fn instruction_signatures(op: Op) -> Option<&'static [InstSig<'static>]> {
         | Op::RayQueryGetWorldRayDirectionKHR
         | Op::RayQueryGetWorldRayOriginKHR
         | Op::RayQueryGetIntersectionObjectToWorldKHR
-        | Op::RayQueryGetIntersectionWorldToObjectKHR => reserved!(SPV_KHR_ray_query),
+        | Op::RayQueryGetIntersectionWorldToObjectKHR => {}
```
eddyb, line 814: Same objection as with the other PR, regarding the commenting and the note.

crates/spirv-std/src/ray_tracing.rs
```diff
@@ -159,3 +159,448 @@ bitflags::bitflags! {
...
+        acceleration_structure: &AccelerationStructure,
+        ray_flags: RayFlags,
+        cull_mask: u32,
+        ray_origin: impl Vector<f32, 3>,
+        ray_tmin: f32,
+        ray_direction: impl Vector<f32, 3>,
+        ray_tmax: f32,
+    ) -> Self {
+        unsafe {
+            let this = Self { _private: () };
```
eddyb, line 198: IMO this should be mutable (and a mutable reference passed in `asm!`), but also `MaybeUninit::uninit()` if possible.

I thought opaque types wrapped an `u32` or something, why does this have `()`?

crates/spirv-std/src/ray_tracing.rs
```diff
@@ -159,3 +159,448 @@ bitflags::bitflags! {
...
+                "OpRayQueryInitializeKHR \
+                {ray_query} \
+                {acceleration_structure} \
+                {ray_flags} \
+                {cull_mask} \
+                {ray_origin} \
+                {ray_tmin} \
+                {ray_direction} \
+                {ray_tmax}",
+                ray_query = in(reg) &this,
```
eddyb, line 209: This should *probably* be `&mut this`, given how it's accessed by pointer.

eddyb (changes requested):

crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
```diff
@@ -811,7 +811,10 @@ pub fn instruction_signatures(op: Op) -> Option<&'static [InstSig<'static>]> {
         | Op::RayQueryGetWorldRayDirectionKHR
         | Op::RayQueryGetWorldRayOriginKHR
         | Op::RayQueryGetIntersectionObjectToWorldKHR
-        | Op::RayQueryGetIntersectionWorldToObjectKHR => reserved!(SPV_KHR_ray_query),
+        | Op::RayQueryGetIntersectionWorldToObjectKHR => {
+            // NOTE(eddyb) we actually use these despite not being in the standard yet.
+            // reserved!(SPV_KHR_ray_tracing)
```
eddyb, line 816: Heh, it's not the same identifier.

eddyb (changes requested):

crates/spirv-std/src/ray_tracing.rs
XAMPPRocky: Well because using `u32` causes the type to not be able to be constructed in rust because of pointer casts.

```
   --> /Users/erin.power/src/rust-gpu2/crates/spirv-std/src/ray_tracing.rs:198:28
    |
198 |             let mut this = Self { _private: 0 };
    |                            ^^^^^^^^^^^^^^^^^^^^
    |
    = note: from: *RayQueryKhr
    = note: to: *u32

```

I tried using the same hack that used for `AccelerationStructure`, however this causes some issues as there seem to issues around using `OpTypePointer` with `RayQueryKHR` missing from the type cache. 

```rust
pub fn initialize(
    acceleration_structure: &AccelerationStructure,
    ray_flags: RayFlags,
    cull_mask: u32,
    ray_origin: impl Vector<f32, 3>,
    ray_tmin: f32,
    ray_direction: impl Vector<f32, 3>,
    ray_tmax: f32,
) -> Self {
    unsafe {
        asm! {
            "%ray = OpTypeRayQueryKHR",
            "%ptr = OpTypePointer Generic %ray",
            "%handle = OpVariable %ptr Generic",
            "OpRayQueryInitializeKHR \
            %handle \
            {acceleration_structure} \
            {ray_flags} \
            {cull_mask} \
            {ray_origin} \
            {ray_tmin} \
            {ray_direction} \
            {ray_tmax}",
            "OpReturnValue %handle",
            "%blah = OpLabel",
            acceleration_structure = in(reg) acceleration_structure,
            ray_flags = in(reg) ray_flags.bits(),
            cull_mask = in(reg) cull_mask,
            ray_origin = in(reg) &ray_origin,
            ray_tmin = in(reg) ray_tmin,
            ray_direction = in(reg) &ray_direction,
            ray_tmax = in(reg) ray_tmax,
        }

        loop {}
    }
}
```

eddyb: Did you add the special handling necessary in `spirv_asm` for `OpTypeRayQueryKHR`? The other type instructions get recognized and converted into `SpirvType` instead of simply being emitted.

Either way, `_private: 0` is not what I had in mind, but rather using `MaybeUninit` (if possible).

XAMPPRocky: > Either way, _private: 0 is not what I had in mind, but rather using MaybeUninit (if possible).

MaybeUninit doesn't work, I've tried using it before. If you do you get the following error.

```
thread 'rustc' panicked at 'assertion failed: `(left != right)`
  left: `0`,
 right: `0`: TyAndLayout {
    ty: core::mem::MaybeUninit<spirv_std::ray_tracing::RayQuery>,
    layout: Layout {
        fields: Union(
            2,
        ),
        variants: Single {
            index: 0,
        },
        abi: Aggregate {
            sized: true,
        },
        largest_niche: None,
        align: AbiAndPrefAlign {
            abi: Align {
                pow2: 0,
            },
            pref: Align {
                pow2: 3,
            },
        },
        size: Size {
            raw: 0,
        },
    },
}', crates/rustc_codegen_spirv/src/abi.rs:565:13
```

crates/spirv-std/src/ray_tracing.rs
```diff
@@ -159,3 +159,473 @@ bitflags::bitflags! {
...
+macro_rules! ray_query {
+    (let $name:ident) => {
+        $crate::ray_query!(@inner $name)
+    };
+    (let mut $name:ident) => {
+        $crate::ray_query!(@inner $name, mut)
+    };
+    (@inner $name:ident $(, $mut:tt)?) => {
+        #[allow(deprecated)]
+        let mut $name : *mut RayQuery = core::mem::uninitialized();
```
eddyb, line 182: You don't need to make it mutable nor initialize it - you're initializing it through the `asm!`.

crates/spirv-std/src/ray_tracing.rs
```diff
@@ -159,3 +159,473 @@ bitflags::bitflags! {
         const SKIP_AABBS = 512;
     }
 }
+
+/// A ray query type which is an opaque object representing a ray traversal.
+#[spirv(ray_query)]
+#[derive(Clone, Default)]
```
eddyb, line 165: Neither of these traits is applicable - you can't have a by-value `RayQuery`.
Also, it should either be unsized *or* used only through `Pin<&mut RayQuery>`, otherwise `mem::swap` may work.

crates/spirv-std/src/ray_tracing.rs
```diff
@@ -159,3 +159,473 @@ bitflags::bitflags! {
...
+    /// - `cull_mask` is the mask to test against the instance mask.  Only the 8
+    ///   least-significant bits of `cull_mask` are used by this instruction - other
+    ///   bits are ignored.
+    /// - `ray_origin`, `ray_tmin`, `ray_direction`, and `ray_tmax` control the
+    ///   basic parameters of the ray to be traced.
+    #[spirv_std_macros::gpu_only]
+    #[doc(alias = "OpRayQueryInitializeKHR")]
+    #[inline]
+    #[allow(clippy::too_many_arguments)]
+    pub fn initialize(
```
eddyb, line 214: Given all other methods are `unsafe`, shouldn't this one be too?

eddyb (changes requested):

crates/spirv-std/src/ray_tracing.rs
```diff
@@ -159,3 +159,493 @@ bitflags::bitflags! {
...
+#[macro_export]
+macro_rules! ray_query {
+    (let $name:ident) => {
+        $crate::ray_query!(@inner $name)
+    };
+    (let mut $name:ident) => {
+        $crate::ray_query!(@inner $name, mut)
+    };
+    (@inner $name:ident $(, $mut:tt)?) => {
+        #[allow(deprecated)]
```
eddyb, line 180: You shouldn't need `#[allow(deprecated)]` if you're not calling `mem::uninitialized`. Also, you can move this into the `unsafe` block, and even change its type to `&$($mut)? RayQuery`.

Or even just initialize the user variable with `asm!`, no need for 2 variables.

eddyb (changes requested):

eddyb (approved):

eddyb (approved):

PR #574, opened 2021-04-01
structurizer: split regions into `Divergent` vs `Convergent`, and optimize based on that.

eddyb:
The first commit is a pretty large refactor that ended up a bit worse than I expected.
We might just want to have this instead (and call it in a few places):
```rust
impl Region {
    fn diverges(&self) -> bool {
        self.exits.is_empty()
    }
}
```

The second commit is just an on-the-fly `OpPhi` simplification that we should probably merge completely separately.

And the third commit is "clever hack" (to remove some unnecessary empty blocks, see example below) that could also probably be done without the first commit, it just didn't hit me that it was possible before the refactor.

<details>
<summary>(click to open example code & before/after graphs for the third commit)</summary>

```rust
pub fn marker() {}

pub fn shortcircuit_and(a: bool, b: bool) {
    if a && b {
        marker();
    }
}
```

### Before
[](http://build.lyken.rs/~eddy/tmp/rFdU/_.svg)

### After
[](http://build.lyken.rs/~eddy/tmp/8dZP/_.svg)

</details>

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/new_structurizer.rs
+++ b/crates/rustc_codegen_spirv/src/linker/new_structurizer.rs
@@ -441,12 +441,8 @@ impl Structurizer<'_> {
                     );
                     let repeat_case = (exit.condition.unwrap_or(const_true), while_body_merge_id);
                     let phi_cases = [first_entry_case, repeat_case];
-                    exit.condition = Some(
-                        self.func
-                            .builder
-                            .phi(type_bool, None, phi_cases.iter().copied())
-                            .unwrap(),
-                    );
+                    exit.condition =
+                        Some(self.emit_phi_or_simplfy(type_bool, phi_cases.iter().copied()));
                 }
 
                 // Choose whether to keep looping, in the `while`-like loop header.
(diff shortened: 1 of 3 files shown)


crates/rustc_codegen_spirv/src/linker/new_structurizer.rs
```diff
@@ -168,11 +191,7 @@ impl Structurizer<'_> {
             let block_id = self.func.blocks()[block].label_id().unwrap();
             let terminator = self.func.blocks()[block].instructions.last().unwrap();
             let mut region = match terminator.class.opcode {
-                Op::Return | Op::ReturnValue | Op::Kill | Op::Unreachable => Region {
-                    merge: block,
-                    merge_id: block_id,
-                    exits: indexmap! {},
-                },
+                Op::Return | Op::ReturnValue | Op::Kill | Op::Unreachable => Region::Divergent,
```
khyperia, line 194: Just a note here that this will conflict with https://github.com/EmbarkStudios/rust-gpu/pull/563, and it might not be immediately obvious if a merge goes wonky here, so note to double check a merge goes well

khyperia (commented):
Seems good to me! Type saftey with different types for divergent/convergent is probably kinda nice, makes you think about things instead of writing generic code for both, so I'd probably be in favor of keeping it if you like it.

eddyb:
Closing in favor of SPIR-T (which has its own structurizer, with a different design).

PR #576, opened 2021-04-05
Deprecate `#[spirv(block)]` and auto-wrap in "interface blocks" instead.

### LINKED ISSUE #391: Make #[spirv(block)] decoration force ABI type into Struct
In a situation where one wants to use a non-struct primitive type as a push constant or uniform, you need to put it into a struct and apply the `block` decoration to it. However, if we try to just put it in a wrapper struct the ABI will not actually change. For example this

```rust
#[spirv(block)]
#[derive(Clone, Copy)]
#[allow(unused_attributes)]
pub struct Constants {
    direction: Vec2,
    _make_struct: f32,
}
```

without the `_make_struct` field, this will still be an AbiVector and so it's illegal to apply the `block` decoration to it. This also occurs for two f32s in which case it will be a ScalarPair, like here:

```rust
#[spirv(block)]
#[derive(Copy, Clone)]
#[allow(unused_attributes)]
pub struct TonemapConstants {
    pub exposure: f32,
    pub aspect: f32,
    _make_struct: f32,
}
```

It would be nice to not have to add this extra member to the struct.

eddyb:
As per [Vulkan](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#interfaces-resources) and [OpenGL](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_gl_spirv.txt) requirements:
* `PushConstant`, `Uniform` and `StorageBufer` interface variables *must* have a `Block`-decorated `OpTypeStruct` type
  * this "struct" type is called an "interface block" and serves mostly as a grouping construct
  * as per https://github.com/EmbarkStudios/rust-gpu/issues/391#issuecomment-813090156, having a `struct` (with some fields) inside an "interface block" shouldn't differ from having those same fields in the "interface block" itself
  * after this PR, we automatically generate a single-field "interface block" for these storage classes, instead of requiring the user to declare a `struct` and annotate it with `#[spirv(block)]`
* `Input`/`Output` can also use "interface blocks", but it's not mandatory
  * we'd probably want to think a bit more about this once/if we have usecases in mind

Most of the commits in this PR are small refactors, leading up to the last commit, which is the actual change.

Because of `#[spirv(block)]` being deprecated instead of outright removed, this isn't a breaking change, except for slices, where you have to remove the `struct` wrapping the `[T]` that you previously had to add - it's now just:
```rust
#[spirv(storage_buffer, descriptor_set = 0, binding = 0)] slice: &mut [f32],
```

The auto-generated "interface block" also allows simpler uniforms like this to "just work":
```rust
#[spirv(uniform, descriptor_set = 1, binding = 0)] u_view_proj: &Mat4,
```

Closes #391.

**EDIT**: temporarily rebased on top of #577 to unbreak CI.

### DIFF
--- a/crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
@@ -30,19 +30,23 @@ impl<'tcx> CodegenCx<'tcx> {
         name: String,
         entry: Entry,
     ) {
-        let local_id = match instance.def_id().as_local() {
-            Some(id) => id,
-            None => {
-                self.tcx
-                    .sess
-                    .err(&format!("Cannot declare {} as an entry point", name));
-                return;
-            }
+        let span = self.tcx.def_span(instance.def_id());
+        let hir_params = {
+            let fn_local_def_id = match instance.def_id().as_local() {
+                Some(id) => id,
+                None => {
+                    self.tcx
+                        .sess
+                        .span_err(span, &format!("Cannot declare {} as an entry point", name));
+                    return;
+                }
+            };
+            let fn_hir_id = self.tcx.hir().local_def_id_to_hir_id(fn_local_def_id);
+            let body = self.tcx.hir().body(self.tcx.hir().body_owned_by(fn_hir_id));
+            body.params
         };
-        let fn_hir_id = self.tcx.hir().local_def_id_to_hir_id(local_id);
-        let body = self.tcx.hir().body(self.tcx.hir().body_owned_by(fn_hir_id));
         const EMPTY: ArgAttribute = ArgAttribute::empty();
-        for (abi, arg) in fn_abi.args.iter().zip(body.params) {
+        for (abi, hir_param) in fn_abi.args.iter().zip(hir_params) {
             match abi.mode {
                 PassMode::Direct(_)
                 | PassMode::Indirect { .. }
@@ -63,7 +67,7 @@ impl<'tcx> CodegenCx<'tcx> {
                     },
                 ) => {}
                 _ => self.tcx.sess.span_err(
-                    arg.span,
+                    hir_param.ty_span,
                     &format!("PassMode {:?} invalid for entry point parameter", abi.mode),
                 ),
             }
@@ -71,7 +75,7 @@ impl<'tcx> CodegenCx<'tcx> {
         if let PassMode::Ignore = fn_abi.ret.mode {
         } else {
             self.tcx.sess.span_err(
-                self.tcx.hir().span(fn_hir_id),
+                span,
                 &format!(
                     "PassMode {:?} invalid for entry point return type",
                     fn_abi.ret.mode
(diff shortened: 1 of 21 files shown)


termhn:
Very nice :D

termhn (approved):

eddyb:
@khyperia @XAMPPRocky For the record I think this could be avoided in the future if GitHub has an option to require reviews from everyone who was added as a reviewer.

crates/rustc_codegen_spirv/src/spirv_type.rs
```diff
@@ -248,6 +249,20 @@ impl SpirvType {
...
+            Self::InterfaceBlock { inner_type } => {
+                let mut emit = cx.emit_global();
+                let id = emit.id();
+                let result = emit.type_struct_id(Some(id), iter::once(inner_type));
+                emit.decorate(result, Decoration::Block, iter::empty());
+                emit.member_decorate(
+                    result,
+                    0,
+                    Decoration::Offset,
+                    [Operand::LiteralInt32(0)].iter().cloned(),
```
XAMPPRocky, line 262: What happens if there's multiple interface blocks in the function? Won't this point them to the same offset?

crates/rustc_codegen_spirv/src/spirv_type.rs
```diff
@@ -499,6 +516,12 @@ impl fmt::Debug for SpirvTypePrinter<'_, '_> {
                 .field("id", &self.id)
                 .field("image_type", &self.cx.debug_type(image_type))
                 .finish(),
+
+            SpirvType::InterfaceBlock { inner_type } => f
+                .debug_struct("SampledImage")
```
XAMPPRocky, line 521:
```rust
.debug_struct("InterfaceBlock")
```

crates/rustc_codegen_spirv/src/spirv_type.rs
eddyb: This is just what you would get from this kind of struct:
```rust
#[spirv(block)]
struct InterfaceBlock<T> {
    inner: T,
}
```

So this isn't a regression, and I believe offsets are per interface variable.

There's an unrelated issue that we allow multiple push constants and multiple other bindings with the same `(descriptor_set, binding)`, that's the only situations I can think of in which anything would overlap.

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -108,124 +113,136 @@ impl<'tcx> CodegenCx<'tcx> {
...
+
+                                    // Unsized pointee with length (i.e. `&[T]`).
+                                    PassMode::Pair(..) => {
+                                        // FIXME(eddyb) shouldn't this be `usize`?
+                                        let len_spirv_type = self.type_isize();
+
+                                        let len = bx
+                                            .emit()
+                                            .array_length(
+                                                len_spirv_type,
```
khyperia, line 184: 12 levels of indentation makes me sad :( would be nice to split this out

khyperia (commented):
LGTM

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
eddyb: I guess the newly introduce use of `Builder` means that we can interleave things a bit more and not have to separate them out so much - e.g. `declare_interface_global_for_param` could also return an `OperandValue` for the call args.

PR #577, opened 2021-04-05
Unbreak CI (fix control_barrier and memory_barrier tests).

eddyb:
#519 was landed with a stale (2 week old at this point) CI, predating the introduction of Vulkan validation on CI, and now `main` itself doesn't pass CI anymore.

I don't like my `Semantics` bodge, but it being a bitset means using `enum` is just wrong.

I could've switched it to `bitflags!` but it seemed easier to just work around it for now.


eddyb:
Well, this is awkward but it landed when @termhn approved #576.

PR #580, opened 2021-04-07
Address review comments on #576 (Deprecate #[spirv(block)] and auto-wrap in "interface blocks" instead).

eddyb:
(This is needed because #576 got merged before all the reviews could be done)

Same as last time, most of the commits are small(er) refactors. The last one consolidates all the logic to do with adjusting the global interface `OpVariable` pointer to match the entry `fn` call argument, into `declare_shader_interface_for_param`, instead of being nested in some closure in `shader_entry_stub`.


khyperia (approved):
nice!!

PR #586, opened 2021-04-09
Allow constant references (to constant data), when they're valid SPIR-V.

eddyb:
This is trickier than it should reasonably be, because of some SPIR-V limitations.

A constant like `&&123` is valid in SPIR-V and generates something like this (types omitted for brevity):
```llvm
        %int_123 = OpConstant _ 123
    %ref_int_123 = OpVariable _ Private %int_123
%ref_ref_int_123 = OpVariable _ Private %ref_int_123
```

But a constant like `&(&1, &2)` isn't - its SPIR-V would have to be this, which doesn't pass validation:
```llvm
    %int_1 = OpConstant _ 1
%ref_int_1 = OpVariable _ Private %int_1
    %int_2 = OpConstant _ 2
%ref_int_2 = OpVariable _ Private %int_2
     %pair = OpConstantComposite _ %ref_int_1 %ref_int_2
 %ref_pair = OpVariable _ Private %ref_pair
```

The problem isn't even `%ref_pair`, but `%pair` itself: `OpConstantComposite` *must* have `OpConstant...` fields, but not (global, i.e. module-scoped) `OpVariable` fields (even if the storage class is `Private`).

This is pretty bizarre IMO, and I'm not 100% convinced it's intentional - it's possible the first case (with the nested `OpVariable`s) was never meant to be legal either.

<hr/>

Regardless of whether the rules make sense or not, this PR accounts for them, and you can see in the UI tests some simple examples what is allowed and what isn't (the `==` examples were the original motivation for this PR).

I've even tested that adding multiple extraneous references to the color constants in `mouse-shader` still works, but I suspect the loads get constant-folded anyway, so it's not much of a test.

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder_spirv.rs
+++ b/crates/rustc_codegen_spirv/src/builder_spirv.rs
@@ -171,16 +171,22 @@ impl SpirvValueExt for Word {
 
 #[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
 pub enum SpirvConst {
-    U32(Word, u32),
-    U64(Word, u64),
+    U32(u32),
+    U64(u64),
     /// f32 isn't hash, so store bits
-    F32(Word, u32),
+    F32(u32),
     /// f64 isn't hash, so store bits
-    F64(Word, u64),
-    Bool(Word, bool),
-    Composite(Word, Vec<Word>),
-    Null(Word),
-    Undef(Word),
+    F64(u64),
+    Bool(bool),
+    Composite(Vec<Word>),
+    Null,
+    Undef,
+}
+
+#[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+struct WithType<V> {
+    ty: Word,
+    val: V,
 }
 
 /// Cursor system:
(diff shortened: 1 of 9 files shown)


crates/rustc_codegen_spirv/src/builder_spirv.rs
```diff
@@ -329,44 +414,143 @@ impl BuilderSpirv {
...
+            | SpirvConst::F32(_)
+            | SpirvConst::F64(_)
+            | SpirvConst::Bool(_) => Ok(()),
+
+            SpirvConst::Null => {
+                // FIXME(eddyb) check that the type supports `OpConstantNull`.
+                Ok(())
+            }
+            SpirvConst::Undef => {
+                // FIXME(eddyb) check that the type supports `OpUndef`.
```
khyperia, line 465: Just curious, are there any types that _cannot_ be made by OpUndef? Intuitively, there absolutely should be, but I don't think the spec actually mentions anything.

crates/rustc_codegen_spirv/src/builder_spirv.rs
```diff
@@ -214,7 +292,13 @@ pub struct BuilderCursor {
 
 pub struct BuilderSpirv {
     builder: RefCell<Builder>,
-    constants: RefCell<BiHashMap<SpirvConst, SpirvValue>>,
+
+    // Bidirectional maps between `SpirvConst` and the ID of the defined global
+    // (e.g. `OpConstant...`) instruction.
+    // NOTE(eddyb) both maps have `WithConstLegality` around their keys, which
+    // allows getting that legality information without additional lookups.
```
khyperia, line 299: nit: around their **values**?

khyperia (approved):
what the fuck

> [OpVariable instruction spec]: Initializer is optional. If Initializer is present, it will be the initial value of the variable’s memory content. Initializer must be an <id> from a constant instruction **or a global (module scope) OpVariable instruction**.

PR #591, opened 2021-04-12
Disable variable pointers by default

### LINKED ISSUE #530: Shouldn't require `OpCapability VariablePointers`
rust-gpu currently always emits

```
               OpCapability VariablePointers
```

at the start of the SPIR-V, whether it's needed or not. And it should not be needed, which is a separate issue.

Related to #531

khyperia:
Fixes #530

I have not tested to see if ark shaders still build with this, I'm not sure how, so if someone could check that'd be great.


XAMPPRocky:
I have built ark shaders with these changes, and it it worked fine.

XAMPPRocky (approved):

eddyb (approved):

PR #592, opened 2021-04-12
Add compile_spirv proc macro to replace build.rs

Hentropy:
Adds a proc macro that builds and executes a SpirvBuilder at compile time.  This replaces the build.rs use-case of SpirvBuilder, but does not replace using SpirvBuilder directly if runtime shader compilation is desired.  The macro returns the result of SpirvBuilder::build(), ie. the path to the .spv file.

Syntax:
```rust
compile_spirv! {
    // required
    path: "path/to/crate",
    // optional, defaults to 1.3
    version: 1.5,
    // optional, defaults to Vulkan
    memory_model: GLSL450,
    // either...
    debug,
    // ...or, default is release
    release,
}
```


Hentropy:
This will unblock #423

lachlansneff:
Does this work with emitting a spirv file for each entry point?

Hentropy:
No, though it could be extended to do so.  Another useful extension would be to have it return the spirv directly, instead of a path to a .spv file.

XAMPPRocky:
Thank you for your PR! Can you update this to have the same options as the current build.rs version? E.g. Remove the version and `memory_model` arguments, and add the `target` and multi_module arguments?

Also I think the parsing logic for the arguments is a bit too overly complex in terms of the amount of duplicated code, I would use some macros to clean up the validation logic. You can see similar logic in my [image macro PR](https://github.com/EmbarkStudios/rust-gpu/pull/359/files#diff-a20b7fca1c7f0f3bb8577a84235e0fe3b864c220df8e43cf4a78a2c9f9b46f7e).

khyperia (changes requested):
We should discuss this in the next meeting, this is a pretty major change with large tradeoffs and should go through at least the MCP process (or even an RFC, tbh) before being approved. Doing a `request changes` to make sure it's blocked before that happens.

Hentropy:
Closing, unclear how #423 will end up working/if it will.

PR #595, opened 2021-04-15
asm: infer the result type of OpAccessChain (when indexing into arrays).

eddyb:
See added test for example usecase.

The changes done here will also make it easier in the future to support `asm!` inference in more cases.


XAMPPRocky (approved):

XAMPPRocky:
Whoops, didn't see that you had automerge on.

Issue #570, opened 2021-03-31
Implement `options(noreturn)` for `asm!`.


XAMPPRocky:
Currently you cannot correctly write an inline assembly that uses a terminator, as there's no way to specify `options(noreturn)`. This means for example that we can't set functions that use these terminators to `-> !` to be able to behave like `panic!` from std.

PR #596, opened 2021-04-16
Add runtime array type

khyperia:
This allows runtime descriptor arrays, which aren't wrapped in a block struct, and have no length.


XAMPPRocky (approved):

crates/spirv-std/src/runtime_array.rs
```diff
@@ -0,0 +1,39 @@
...
+    #[allow(clippy::empty_loop)]
+    pub unsafe fn index(&self, index: usize) -> &T {
+        asm! {
+            "%result = OpAccessChain _ {arr} {index}",
+            "OpReturnValue %result",
+            "%unused = OpLabel",
+            arr = in(reg) self,
+            index = in(reg) index,
+        }
+        loop {}
```
lachlansneff, line 24: You should be able to replace the `loop {}` with `options(noreturn)` in the asm block.

https://doc.rust-lang.org/nightly/unstable-book/library-features/asm.html#options
https://doc.rust-lang.org/nightly/unstable-book/library-features/asm.html#options-1

crates/spirv-std/src/runtime_array.rs
```diff
@@ -0,0 +1,39 @@
...
+    #[allow(clippy::empty_loop)]
+    pub unsafe fn index_mut(&mut self, index: usize) -> &mut T {
+        asm! {
+            "%result = OpAccessChain _ {arr} {index}",
+            "OpReturnValue %result",
+            "%unused = OpLabel",
+            arr = in(reg) self,
+            index = in(reg) index,
+        }
+        loop {}
```
lachlansneff, line 37: Same here.

crates/spirv-std/src/runtime_array.rs
msiglreith: There is an open issue for supporting it: https://github.com/EmbarkStudios/rust-gpu/issues/570

lachlansneff: Ah, my bad. I should've seen that beforehand.

crates/spirv-std/src/runtime_array.rs
```diff
@@ -0,0 +1,39 @@
+use core::marker::PhantomData;
+
+#[spirv(runtime_array)]
+pub struct RuntimeArray<T> {
+    // spooky! this field does not exist, so if it's referenced in rust code, things will explode
+    _do_not_touch: u32,
+    _phantom: PhantomData<T>,
+}
```
Hentropy, line 8: Calling this `RuntimeArray` is a bit confusing, someone was already in chat asking whether this was used for runtime data arrays. `RuntimeDescriptorArray` is a bit verbose, but it's also unambiguous, and for descriptor arrays of buffers this type needs to insert a block decorated struct around the buffer and have special indexing behaviour to handle that.

The `u32` field is a bit strange, descriptor arrays, like descriptors, don't really have size/alignment per se.  Since the compiler is handling this specially anyways, is it necessary?

crates/spirv-std/src/runtime_array.rs
```diff
@@ -0,0 +1,39 @@
...
+    #[spirv_std_macros::gpu_only]
+    #[allow(clippy::empty_loop)]
+    pub unsafe fn index(&self, index: usize) -> &T {
+        asm! {
+            "%result = OpAccessChain _ {arr} {index}",
+            "OpReturnValue %result",
+            "%unused = OpLabel",
+            arr = in(reg) self,
+            index = in(reg) index,
+        }
```
Hentropy, line 23: This only works for the opaque resource types, extending this to support buffers requires a compiler implemented function.

crates/rustc_codegen_spirv/src/abi.rs
```diff
@@ -810,5 +810,25 @@ fn trans_intrinsic_type<'tcx>(
...
+                    .sess
+                    .err("#[spirv(runtime_array)] type must have size 4");
+                return Err(ErrorReported);
+            }
+
+            // We use a generic to indicate the underlying element type.
+            // The spirv type of it will be generated by querying the type of the first generic.
+            if let Some(elem_ty) = substs.types().next() {
+                let element = trans_type_impl(cx, span, cx.layout_of(elem_ty), false);
+                Ok(SpirvType::RuntimeArray { element }.def(span, cx))
```
Hentropy, line 825: `.def()` produces an array stride decoration, which isn't necessary for opaque resource types ([shader-playground GLSL example](http://shader-playground.timjones.io/a2a0f437a19f698843917b7f3ac21c03)) and is explicitly forbidden for buffers.

Hentropy (commented):
Descriptor indexing also works for buffers, both sized and unsized as in #389. For that to work, there needs to be a separate type for `DescriptorArray` and `RuntimeDescriptorArray` because arrays and slices in storage buffer or uniform storage are inferred to be buffers that need to be wrapped in `SpirvType::InterfaceBlock`.  It also makes it a little bit clearer what the type is for, someone was lil confused in chat on the weekend about it.  I didn't know this was in progress, and it's something I need so I had started working on it myself about the same time you did.  Kind of a shame to do duplicate work, but it's essentially the extension of this that allows for (possibly variably sized) buffers as well as opaque resources.

crates/spirv-std/src/runtime_array.rs
eddyb: > Since the compiler is handling this specially anyways, is it necessary?

You have to avoid making it a ZST, because those can be assumed to be data-less tokens, i.e. they have only "existence" but no memory/register representation. For example, `(SomeZST, f32)` is represented as *just* an `f32`, and so there is no way to recover any data the backend may want to keep in the ZST type.

The other option (instead of a dummy field) is a `!Sized` type, via `extern { type Foo; }`, which we should maybe move to at some point.

khyperia: Yeah, I'd love to move to `extern { type Foo; }`, but IIRC it requires a `#[feature]` and I'd like to keep those limited if at all possible (and AFAIK there's no other way to create an unsized yet no wide pointer type)

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -320,11 +329,49 @@ impl<'tcx> CodegenCx<'tcx> {
...
             }
+            StorageClass::UniformConstant => {
+                var_ptr_spirv_type = self.type_ptr_to(value_spirv_type);
+
+                match self.lookup_type(value_spirv_type) {
+                    SpirvType::RuntimeArray { .. } => {
+                        if is_pair {
+                            self.tcx.sess.span_err(
+                                hir_param.ty_span,
+                                "uniform_constant buffers must use &RuntimeArray<T>, not &[T]",
```
eddyb, line 352: > `uniform_constant buffers`

I thought `UniformConstant` was only for non-buffers?

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -290,6 +296,9 @@ impl<'tcx> CodegenCx<'tcx> {
         // which we represent with `SpirvType::InterfaceBlock` (see its doc comment).
         // This "interface block" construct is also required for "runtime arrays".
         let is_unsized = self.lookup_type(value_spirv_type).sizeof(self).is_none();
+        let is_pair = matches!(entry_arg_abi.mode, PassMode::Pair(..));
+        // is_pair implies is_unsized (but is_unsized doesn't imply is_pair, for RuntimeArray<T>)
+        assert!(!is_pair || is_unsized);
```
eddyb, line 301: Can't this become really confusing if someone uses some type that just happens to be represented as a pair? I think you could replace `if is_pair {` below with `if is_unsized_with_len {` given:
```rust
let is_unsized_with_len = is_unsized && matches!(entry_arg_abi.mode, PassMode::Pair(..));
```

eddyb (approved):
I think this is fine, but feel free to do some of the suggested changes.

Issue #581, opened 2021-04-07
barrier `SEMANTICS` const generic as bitflag


msiglreith:
The `SEMANTICS` used in `control_barrier` and `memory_barrier` is a bitflag, but currently only `memory_barrier` handles it as such. `control_barrier` only allows to set a single enum/flag.

https://github.com/EmbarkStudios/rust-gpu/blob/main/crates/spirv-std/src/arch/barrier.rs#L33-L36
https://github.com/EmbarkStudios/rust-gpu/blob/main/crates/spirv-std/src/arch/barrier.rs#L69

PR #598, opened 2021-04-20
Fix semantics type for control barriers to support multiple flags

msiglreith:
Fixes https://github.com/EmbarkStudios/rust-gpu/issues/581

`control_barrier` API now matches `memory_barrier` by using a `u32` instead of an enum type for specifying `Semantics` bitflags.

### DIFF
--- a/crates/spirv-std/src/arch/barrier.rs
+++ b/crates/spirv-std/src/arch/barrier.rs
@@ -30,7 +30,6 @@ use crate::memory::{Scope, Semantics};
 #[spirv_std_macros::gpu_only]
 #[doc(alias = "OpControlBarrier")]
 #[inline]
-// FIXME(eddyb) use a `bitflags!` `Semantics` for `SEMANTICS`.
 pub unsafe fn control_barrier<
     const EXECUTION: Scope,
     const MEMORY: Scope,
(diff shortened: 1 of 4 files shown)


XAMPPRocky:
Thank you for PR! I think we might as well switch the `Semantics` type to using `bitflags` than go to `u32`, I've already had to add bitflags for `RayFlags`.

msiglreith:
Pushed a change with converting the enum to bitflags after the discussion on discord.
As const trait impls are behind a nightly feature flag in Rust, the `from_bits_truncate` workaround is needed in case of multiple flags.

crates/spirv-std/src/arch/barrier.rs
```diff
@@ -30,6 +30,7 @@ use crate::memory::{Scope, Semantics};
 #[spirv_std_macros::gpu_only]
 #[doc(alias = "OpControlBarrier")]
 #[inline]
+// FIXME(eddyb) use a `bitflags!` `Semantics` for `SEMANTICS`.
```
XAMPPRocky, line 33: This can be removed right?
```suggestion
```

XAMPPRocky (approved):

PR #601, opened 2021-04-23
CI check that docs are able to be built without warnings

khyperia:
I got annoyed that tons of warnings get spewed when running `cargo doc`, so, made it a CI check.

I've intentionally not fixed one instance of ``[`AccelerationStructureKHR`]`` to make sure CI properly fails with the new check, I'll fix after it fails.

(The diff here is a bit annoying since I reflowed all the comments I touched to 80 columns - I _think_ rust convention is that they should actually reflow at 100 columns, but all the comments I touched seemed to be consistent at reflowing at roughly around 80 columns, so I kept that)


XAMPPRocky (approved):
LGTM, and yeah I've always just wrote everything in 80 columns.

PR #609, opened 2021-04-29
Move disassemble tests to compiletest

khyperia:
I got real tired of waiting for `cargo test -p spirv-builder`, so, printf hacks go brrrr


crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
```diff
@@ -233,10 +237,114 @@ impl CodegenArgs {
...
+                .functions
+                .iter()
+                .find(|f| f.def_id().unwrap() == id)
+                .unwrap()
+                .clone();
+            // Compact to make IDs more stable
+            compact_ids(&mut func);
+            eprintln!("{}", func.disassemble());
+        }
+        if self.disassemble_globals {
```
XAMPPRocky, line 343: Some newlines between each statement, would help with readability.

XAMPPRocky (approved):
LGTM

crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
```diff
@@ -233,10 +237,118 @@ impl CodegenArgs {
...
+        );
+        opts.optflagopt("", "disassemble-globals", "print globals to stderr", "");
         let matches = opts.parse(args)?;
         let module_output_type =
             matches.opt_get_default("module-output", ModuleOutputType::Single)?;
-        Ok(Self { module_output_type })
+        let disassemble = matches.opt_present("disassemble");
+        let disassemble_fn = matches.opt_str("disassemble-fn");
+        let disassemble_entry = matches.opt_str("disassemble-entry");
+        let disassemble_globals = matches.opt_present("disassemble-globals");
```
XAMPPRocky, line 255: Should some of these flags be incompatible with each other? As in if you pass `--disassemble-fn` and `--disassemble`, you'll get both outputs concatenated.

khyperia: yea probably, but shrug, doesn't matter too much IMO, just adds complexity to an already slightly annoyingly big chunk of code that shouldn't be used anywhere outside of tests

XAMPPRocky: Yeah, it probably doesn't matter much, though I do think this has some utility outside of tests. Mainly debugging and sharing snippets, since this does expose `spirv-dis` functionality without needing spirv tools installed.

eddyb (approved):

PR #610, opened 2021-04-30
Add support for using target-features for extensions and capabilities

### LINKED ISSUE #555: Use target features for capabilities
# Proposal
While working on #529 I was thinking about other approaches we could take, and one that I would like to propose we do instead adding a new attribute is to use `-Ctarget-features` for describing what capabilities to use. As I've thought over I think this has a few important advantages over the current proposed approach.

- More "Rust-tic" — Since target-features is built-in to the compiler, more people are familiar with using `target-feature` for controlling CPU capabilities so using it for GPU capabilities is more intuitive than a custom attribute. This also helps with development as we can use any ecosystem features or integrations that use `target-feature`.
- Cfgable — Using target-feature gives us a free cfg predicate to use in our GPU code. For example you'd be able to write `#[cfg(target_feature = "MultiView")]` to write code that only works when `MultiView` is available.
- Binary focused — Target features can really only be set when you're building the final binary, so a library can't currently enable it for you behind the scenes. This matches the behaviour we wanted to have where you only set the capabilities you need when compiling the final shader.

XAMPPRocky:
resolves #555

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder_spirv.rs
+++ b/crates/rustc_codegen_spirv/src/builder_spirv.rs
@@ -2,6 +2,7 @@ use crate::builder;
 use crate::codegen_cx::CodegenCx;
 use crate::spirv_type::SpirvType;
 use crate::target::SpirvTarget;
+use crate::target_feature::TargetFeature;
 use rspirv::dr::{Block, Builder, Module, Operand};
 use rspirv::spirv::{AddressingModel, Capability, MemoryModel, Op, StorageClass, Word};
 use rspirv::{binary::Assemble, binary::Disassemble};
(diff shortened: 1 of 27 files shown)


tests/ui/dis/target_features_err.rs
```diff
@@ -0,0 +1,10 @@
+// build-fail
+// compile-flags: -Ctarget-feature=+rayTracingKHR,+ext:SPV_KHR_ray_tracing
+
+use spirv_std as _;
+
+#[spirv(any_hit)]
+pub fn main() {
+    unsafe { spirv_std::arch::terminate_ray() }
+}
+
```
khyperia, line 10: Feels kind of weird to have this in the `dis` folder since it's not actually a disassembly test, but shrug, doesn't matter too much (we should probably reorganize each file in `dis` to be in its actual proper place, I kind of just slapped everything here and said "let's deal with it later")

khyperia (approved):
We'll need to nuke both `capability_computation.rs` and the default `Int8` and friends capabilities, but that can be done in a follow-up. Also, there's a couple `asm!("OpCapability")` in spirv-std that need to be removed (and replaced with `#[cfg]`)

PR #611, opened 2021-05-03
fix rustdoc with hack

XAMPPRocky:
See https://github.com/rust-lang/rust/issues/84738 for details, but essentially having this function's docs reference `spirv_types` it fixes the ICE when building docs.

### DIFF
--- a/crates/spirv-std/src/lib.rs
+++ b/crates/spirv-std/src/lib.rs
@@ -122,3 +122,8 @@ fn panic(_: &core::panic::PanicInfo<'_>) -> ! {
 #[cfg(all(not(test), target_arch = "spirv"))]
 #[lang = "eh_personality"]
 extern "C" fn rust_eh_personality() {}
+
+// See: https://github.com/rust-lang/rust/issues/84738
+#[doc(hidden)]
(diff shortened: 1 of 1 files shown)


khyperia (approved):
nice! yay hacks :sparkles:

crates/spirv-std/src/lib.rs
```diff
@@ -122,3 +122,8 @@ fn panic(_: &core::panic::PanicInfo<'_>) -> ! {
 #[cfg(all(not(test), target_arch = "spirv"))]
 #[lang = "eh_personality"]
 extern "C" fn rust_eh_personality() {}
+
+// See: https://github.com/rust-lang/rust/issues/84738
+#[doc(hidden)]
+/// [`spirv_types`]
```
jyn514, line 128: The backticks aren't necessary, rustdoc literally does `s.replace("`", "")`: https://github.com/rust-lang/rust/blob/c825bc431ee5b815847b9bab693c59c43986fc4b/src/librustdoc/passes/collect_intra_doc_links.rs#L936
```suggestion
/// [spirv_types]
```

eddyb (approved):

PR #612, opened 2021-05-04
Add is_helper_invocation, and cleanup arch & asm functions

XAMPPRocky:
Cleaning up and removing all the asm uses of `OpCapability`, and also added `is_helper_invocation`, also made `demote_to_helper_invocation` unsafe, as using it has side effects that make variables undefined.

### DIFF
--- a/crates/spirv-std/src/arch.rs
+++ b/crates/spirv-std/src/arch.rs
@@ -8,6 +8,7 @@ use crate::{scalar::Scalar, vector::Vector};
 mod arithmetic;
 #[cfg(feature = "const-generics")]
 mod barrier;
+mod demote_to_helper_invocation_ext;
 mod derivative;
 mod primitive;
 mod ray_tracing;
@@ -15,6 +16,7 @@ mod ray_tracing;
 pub use arithmetic::*;
 #[cfg(feature = "const-generics")]
 pub use barrier::*;
+pub use demote_to_helper_invocation_ext::*;
 pub use derivative::*;
 pub use primitive::*;
 pub use ray_tracing::*;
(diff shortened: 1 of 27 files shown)


crates/spirv-std/src/arch/demote_to_helper_invocation_ext.rs
```diff
@@ -0,0 +1,48 @@
+/// Demote fragment shader invocation to a helper invocation. Any stores to
+/// memory after this instruction are suppressed and the fragment does not write
+/// outputs to the framebuffer.
+///
+/// Unlike [super::kill], this does not necessarily terminate the invocation. It
+/// is not considered a flow control instruction (flow control does not become
+/// non-uniform) and does not terminate the block.
+///
+/// - **Required Capabilities** `DemoteToHelperInvocationEXT`
+/// - **Required Extensions** `SPV_EXT_demote_to_helper_invocation`
```
khyperia, line 10: Turning these into `#[cfg]` on the method would be nice and is something we should do eventually, but not required right now.

crates/spirv-std/src/arch.rs
```diff
@@ -140,3 +142,17 @@ pub unsafe fn vector_insert_dynamic<T: Scalar, V: Vector<T, N>, const N: usize>(
 
     result
 }
+
+/// Fragment-shader discard.
+///
+/// Ceases all further processing in any invocation that executes it: Only
+/// instructions these invocations executed before [kill] have observable side
+/// effects.
```
khyperia, line 150: I think I would prefer to keep the comments about how OpKill corresponds to discard in GLSL, and OpDemoteToHelperInvocationEXT corresponds to discard in HLSL, as that's what a lot of people are familiar with and would help those people understand what the difference between the two is.

khyperia (approved):

crates/spirv-std/src/arch/demote_to_helper_invocation_ext.rs
XAMPPRocky: Yeah, and I think we'd want to do more than `#[cfg]`, as when it's cfg'ed out, the error will just be "`demote_to_helper_invocation` not found", when ideally the error message would tell you, you need those capabilities and extensions.

khyperia: ... as in, doing a custom cfg attribute? we already decided against that in our discussions though, I'm confused

XAMPPRocky: I'm not sure how it would be exactly, but I meant more like a macro which just expands to two versions of the function containing the `cfg` and a better message similar to `gpu_only`.

PR #618, opened 2021-05-16
Always emit cargo dependency information

msiglreith:
When using the `multimodule` features in spirv-builder, the user currently has to disable `print_metadata`. This disable the spir-v module env variable but also emitting the cargo dependency information to correctly handle when the shader should be rebuilt.
Ideally, the second part shouldn't be affected.

### DIFF
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -86,9 +86,9 @@ impl fmt::Display for SpirvBuilderError {
                 write!(f, "Crate path {} does not exist", path.display())
             }
             SpirvBuilderError::BuildFailed => f.write_str("Build failed"),
-            SpirvBuilderError::MultiModuleWithPrintMetadata => {
-                f.write_str("Multi-module build cannot be used with print_metadata = true")
-            }
+            SpirvBuilderError::MultiModuleWithPrintMetadata => f.write_str(
+                "Multi-module build cannot be used with print_metadata = MetadataPrintout::Full",
+            ),
             SpirvBuilderError::MetadataFileMissing(_) => {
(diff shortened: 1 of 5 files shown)


khyperia:
We should be more in-depth than this. Currently, this change means that if the user requests metadata to not be printed, it... prints metadata.

msiglreith:
Would sth like this be fine?
```rust
enum MetadataPrintout {
    /// Prints nothing.
    None,
    /// Only output dependency metadata (e.g for multimodule).
    DepsOnly,
    /// Emit dependency metadata and env var for spirv module.
    Full,
}
```

msiglreith:
I will update it after the json metdata PR gets through

msiglreith:
Should be ready, not sure why the tests got stuck for 6h /:

khyperia:
CI should theoretically not do that anymore after https://github.com/EmbarkStudios/rust-gpu/pull/637, could you rebase beyond that?

msiglreith:
nice, CI is green now!

crates/spirv-builder/src/lib.rs
```diff
@@ -166,6 +178,7 @@ impl SpirvBuilder {
         self
     }
 
+    /// Builds the module. Returns the path to the built spir-v file. If `print_metadata` is [`MetadataPrintout::Full`],
```
khyperia, line 181: Hmm, weird comment here?

khyperia (approved):
LGTM pending the one thing!

crates/spirv-builder/src/lib.rs
msiglreith: thanks, fixed and rebased again.

Issue #374, opened 2021-01-13
The Vec3/Vec3A byte size conundrum


hrydgard:
So, we currently have the Vec3 and Vec3A types from glam available.

Vec3 is simply three scalar float32 values, and Vec3A is a SIMD type where an actual vector with three floats can fit.

On CPU, this results in:
* Vec3 is (f32, f32, f32), 12 bytes
* Vec3A is NEON or SIMD register, 16 bytes

On GPU/SPIR-V, we have:  (EDITED)
* Vec3/Vec3A are both a SPIR-V vec3 value (`TypeVector x(float) 3`). 12 bytes (but with special alignment requirements depending on buffer type/layout mode).
  * Having this type (and not just scalar) is required as an argument for some texture sampling functions, etc

The biggest issue I have with this is of course that on CPU a Vec3A occupies 16 bytes while on GPU it may occupy 12 or 16 bytes depending on surrounding declaration and buffer layout mode.

This all feels quite unsatisfactory, but I am not sure what the best course of action to clean this up would be.

khyperia:
As of https://github.com/bitshifter/glam-rs/pull/106 (early december), `Vec3` and `Vec3A` are identical on the GPU, both are `TypeVector`.

Hentropy:
The [Vulkan spec requires Vec3 to be `align_of<Vec4>()`](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#interfaces-resources-layout) but the spare 4 bytes are usable, expressed like so in CPU code:
```Rust
#[repr(align(16))]
#[repr(C)]
struct Foo {
  v: Vec3, // align_of<Vec3>() == 4 on the CPU, 16 on GPU with VK, 4 on GPU with VK_EXT_scalar_layout
  f: f32,
}
```
But the spirv from that is:
```
OpMemberDecorate %Foo 0 Offset 0
OpMemberDecorate %Foo 1 Offset 16    // This Offset could be 12, and size_of::<Foo>() could be 16 instead of 32
%float = OpTypeFloat 32
%v3float = OpTypeVecctor %float 3
%Foo = OpTypeStruct %v3float %float
```
because Vec3 needs to be `#[repr(simd)]` so it's represented as an `OpTypeVector`, which makes it `align(16)`, size is an integer multiple of alignment, and offset is the sum of the sizes of the previous members.

[...]

PR #619, opened 2021-05-16
mcp: Add `vector` intrinsic type attribute

msiglreith:
## Background

Types in SPIR-V have different alignment requirements depending on their usage location. For uniform/storage/push constants these need to follow for Vulkan the base, extended or scalar buffer alignment rules.
In case of vector types, `repr(simd)` is currently used for indication and trying to align with the aforementioned rules. In particular, aligning `vec3` to 16bytes. See also https://github.com/EmbarkStudios/rust-gpu/issues/374

## Proposal

(As this only affects Vec3 in reality, I will ignore Vec2/Vec4 here)
Add an intrinsic type attribute `vector` which translate following type definition to a `TypeVector` in SPIR-V:
```rust
#[derive(Debug, Copy, Clone)]
#[cfg_attr(target_arch = "spirv", spirv(vector))]
pub struct Vec3<T> {
    pub x: T,
    pub y: T,
    pub z: T,
}
```
Compared to the repr(simd) layout the vector types the ones generated with `vector` attribute will follow the scalar layout (12 bytes).

Pros:
- Fits better to `bindless` API with structured loading, as in full bindless only push constant layouts are of interest regarding alignment rules. For example loading manual vertex fetching of Vec3 requires an extra element compared to using a R32G32B32 vertex format.
- Vec3 can be manually padded and rust-gpu could verify the correct alignment when translating the entry point

Cons:
- 3x3 matrix requires explicit padding for non-scalar layouts to not improperly straddle. (Might as well just use a 4x3 matrix in this case.)
- Most of the time when using a vec3 the user needs explicit add a field for padding compared to 16bytes aligned types

## TODO
- Update glam
- Adjust tests

## Future work
- Check alignments on entry point translation


khyperia (changes requested):
This has significant implications more wide-reaching than what's dealt with here, unfortunately. The issue with this is that rustc's understanding and rust-gpu's understanding of size and alignment does not match with this change, and that will cause significant problems down the line - we need significant design and implementation work to make this work, likely including upstream (rustc) design and implementation work as well, as rustc needs to be aware of what is effectively a size=12 align=16 type, something that the architecture for is severely lacking at the moment. It might be best to look into rustc support for this first.

msiglreith:
> what is effectively a size=12 align=16 type, something that the architecture for is severely lacking at the moment

One part of this proposal would be to treat it simply as size=12 align=4 type, assuming a scalar layout. The user would have to manually layout the structs themself with optional verification by the spirv codegen backend in case the storage class imposes any alignment rules on the struct. It would basically allow for unaligned load operations as the current buffer load/store for bindless also kinda does afaik.

Jasper-Bekkers:
Over the course of this project we've gone back and forth on this, early on we've established that adding a new repr to the rustc compiler or doing something like what you're suggesting would be significantly involved and beyond the scope of this project (for now).

I personally think that instead of supporting super old std140/std430 layouts we should focus ahead and support VK_EXT_scalar_block_layout which just doesn't have this problem anymore.

PR #622, opened 2021-05-18
Always emit json metadata, and emit entry point names

### LINKED ISSUE #620: Always emit json, even for non-multimodule, and embed metadata in the json
We discussed this in the meeting on May 6th, but I want to write it down and assign it to me so I can track it/remember it.

tl;dr we want to do a whole host of stuff (emitting entry point names as string constants in CPU-side codegen, doing CPU-side static asserts for entry point interface struct layout, etc.) that requires passing metadata about the shader out of the compiler and back into the consumer of spirv-builder (so it can do CPU-side codegen). A nice, clean, easy way of passing that metadata back is to use our lil json system we invented for multimodule, and always use it, and embed extra shenanigans into that json.

khyperia:
Fixes #620

### DIFF
--- a/crates/rustc_codegen_spirv/src/compile_result.rs
+++ b/crates/rustc_codegen_spirv/src/compile_result.rs
@@ -1,0 +1,169 @@
+use rustc_data_structures::fx::FxHashMap;
+use serde::{Deserialize, Serialize};
+use std::fmt::Write;
+use std::path::{Path, PathBuf};
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum ModuleResult {
+    SingleModule(PathBuf),
+    MultiModule(FxHashMap<String, PathBuf>),
+}
+
+impl ModuleResult {
+    pub fn unwrap_single(&self) -> &Path {
+        match self {
+            ModuleResult::SingleModule(result) => result,
+            ModuleResult::MultiModule(_) => {
+                panic!("called `ModuleResult::unwrap_single()` on a `MultiModule` result")
+            }
+        }
+    }
+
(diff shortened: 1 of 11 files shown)


VZout (approved):
Awesome! and looks quite easy to add interface variables as well to the metadata. But this does not work with non-multimodule?

khyperia:
> But this does not work with multimodule?

yeah, it does work? `cargo run -p multibuilder` outputs this:

```
CompileResult {
    module: MultiModule(
        {
            "main_fs": "/home/khyperia/me/rust-gpu/target/spirv-unknown-spv1.3/release/deps/sky_shader.spv.dir/main_fs",
            "main_vs": "/home/khyperia/me/rust-gpu/target/spirv-unknown-spv1.3/release/deps/sky_shader.spv.dir/main_vs",
        },
    ),
    entry_points: [
        "main_fs",
        "main_vs",
    ],
}
```

the code does take a little shortcut and yoinks the entrypoint out of the precomputed dictionary (which itself yoinks the entrypoint out of the module) instead of digging around in the module itself, because it's a bit simpler code, but it could easily be expanded to dig through the modules themselves if we wanted to extract more metadata (or carry information through the compiler and emit it, instead of reflecting the output)

VZout:
Ah I saw this and thought metadata was disabled in its entirety for multimodule.
```
if self.print_metadata && self.multimodule {
    return Err(SpirvBuilderError::MultiModuleWithPrintMetadata);
}
```

khyperia:
Ah, that's printing cargo metadata, and is no change in behavior from the current system - it already produces the same error before this PR. "Metadata" in the context of that code means things like printing `cargo:rerun-if-changed=blah` to stdout.

PR #360, opened 2020-12-31
Flesh out compute example

tombh:
Just a copy of wgpu-rs's 'hello-compute' example: https://github.com/gfx-rs/wgpu-rs/tree/v0.6/examples/hello-compute

I'm not sure if this is even possible at the moment? Not to mention that this is my first time using rust-gpu (which is such a great project). I get a compiler error:

```
RUST_BACKTRACE=full cargo run --bin example-runner-wgpu -- --shader Compute
   Compiling example-runner-wgpu v0.1.0 (/home/tombh/Software/rust-gpu/examples/runners/wgpu)
thread 'rustc' panicked at 'Failed to recover key for generics_of(b445f41739b0d39c-316db2fba229b0a9) with hash b445f41739b0d39c-316db2fba229b0a9', compiler/rustc_middle/src/ty/query/mod.rs:235:5
stack backtrace:
   0:     0x7f763a45a6b7 - std::backtrace_rs::backtrace::libunwind::trace::h746c3e9529d524bc
                               at /rustc/d32c320d7eee56706486fef6be778495303afe9e/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x7f763a45a6b7 - std::backtrace_rs::backtrace::trace_unsynchronized::h86340908ff889faa
                               at /rustc/d32c320d7eee56706486fef6be778495303afe9e/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x7f763a45a6b7 - std::sys_common::backtrace::_print_fmt::h43f85f9b18230404
                               at /rustc/d32c320d7eee56706486fef6be778495303afe9e/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x7f763a45a6b7 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hc132ae1a5b5aa7cd
                               at /rustc/d32c320d7eee56706486fef6be778495303afe9e/library/std/src/sys_common/backtrace.rs:46:22
   4:     0x7f763a4ce5ac - core::fmt::write::hdf023a0036d2a25f
                               at /rustc/d32c320d7eee56706486fef6be778495303afe9e/library/core/src/fmt/mod.rs:1078:17
   5:     0x7f763a44c6a2 - std::io::Write::write_fmt::h8580846154bcb66a
                               at /rustc/d32c320d7eee56706486fef6be778495303afe9e/library/std/src/io/mod.rs:1519:15
[...]

PR #623, opened 2021-05-18
Compute a collatz sequence in the compute example

DJMcNab:
These changes were made at the behest of @raphlinus.

The expected values can be found at https://oeis.org/A006877

Should we rename the example to `compute_collatz`?

I'm also planning on integrating the overflow detection from wgpu. (Edit: Done - unfortunately can't use `checked_add` and friends (yet?))

We also should insert the timing queries into this example, since wgpu now supports them.

### DIFF
--- a/examples/runners/wgpu/src/compute.rs
+++ b/examples/runners/wgpu/src/compute.rs
@@ -1,34 +1,15 @@
+use wgpu::util::DeviceExt;
+
 use super::{shader_module, Options};
-use core::num::NonZeroU64;
+use futures::future::join;
+use std::{convert::TryInto, future::Future, num::NonZeroU64, time::Duration};
 
-fn create_device_queue() -> (wgpu::Device, wgpu::Queue) {
-    async fn create_device_queue_async() -> (wgpu::Device, wgpu::Queue) {
-        let instance = wgpu::Instance::new(wgpu::BackendBit::PRIMARY);
-        let adapter = instance
-            .request_adapter(&wgpu::RequestAdapterOptions {
-                power_preference: wgpu::PowerPreference::default(),
-                compatible_surface: None,
-            })
-            .await
-            .expect("Failed to find an appropriate adapter");
-
-        adapter
-            .request_device(
-                &wgpu::DeviceDescriptor {
-                    label: None,
-                    features: wgpu::Features::empty(),
-                    limits: wgpu::Limits::default(),
-                },
-                None,
-            )
-            .await
-            .expect("Failed to create device")
-    }
+fn block_on<T>(future: impl Future<Output = T>) -> T {
     cfg_if::cfg_if! {
         if #[cfg(target_arch = "wasm32")] {
-            wasm_bindgen_futures::spawn_local(create_device_queue_async())
+            wasm_bindgen_futures::spawn_local(future)
         } else {
-            futures::executor::block_on(create_device_queue_async())
+            futures::executor::block_on(future)
         }
     }
 }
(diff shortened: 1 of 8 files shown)


DJMcNab:
I will change the SpirvBuilder invocation to only use the vulkan version which they need - i.e. make only the compute example use vulkan 1.1. 

I'm also not sure whether the  dummy bind group still needs to exist,  and ideally I'd remove the new unsafe code (i.e. using `flat_map` `core::Array::IntoIter` on `char::to_ne_bytes()`. 

Additionally, now that multiple things are async it might be worth threading that through further.

I should also create a CPU runner, because this shader is simple enough that the naïve implementation of that would work fine (it seems to me that the CPU runner could live in `examples/shaders/compute/main.rs`, but I'm not certain). Putting it in with the existing CPU runner would be fine, although the dependencies thereof are a bit heavy weight, since this runner doesn't need any capabilities apart from `stdout` access.

I found the change/run/execute cycle to be excruciatingly slow - am I missing some tricks I can use to get it faster? For reference, my current invocation is:
```powershell
$ cargo run -p example-runner-wgpu -- --shader Compute
```

examples/runners/wgpu/src/compute.rs
```diff
@@ -69,24 +74,29 @@ pub fn start(options: &Options) {
         module: &module,
         entry_point: "main_cs",
     });
-
-    let buf = device.create_buffer(&wgpu::BufferDescriptor {
+    let size = (src.len() * std::mem::size_of::<u32>()).try_into().unwrap();
+    let staging_buffer = device.create_buffer(&wgpu::BufferDescriptor {
```
termhn, line 84: I think this should be called `readback_buffer`. "staging" usually implies that it's being used as an intermediate to copy data to cpu side *before* copying to a gpu only buffer.

examples/runners/wgpu/src/compute.rs
```diff
@@ -69,24 +74,29 @@ pub fn start(options: &Options) {
...
-        usage: wgpu::BufferUsage::STORAGE,
+        size,
+        // Can be read to the CPU, and can be copied from the shader's storage buffer
+        usage: wgpu::BufferUsage::MAP_READ | wgpu::BufferUsage::COPY_DST,
         mapped_at_creation: false,
     });
 
+    let storage_buffer = device.create_buffer_init(&wgpu::util::BufferInitDescriptor {
+        label: Some("Collatz Conjecture Input"),
+        contents: unsafe { core::slice::from_raw_parts(src.as_ptr().cast::<u8>(), size as usize) },
```
termhn: I would be fine with bringing in `bytemuck` for this. Should be able to `cast_slice` safely.

examples/runners/wgpu/src/compute.rs
```diff
@@ -97,8 +107,33 @@ pub fn start(options: &Options) {
...
+        cpass.dispatch(src.len() as u32, 1, 1);
     }
 
+    encoder.copy_buffer_to_buffer(&storage_buffer, 0, &staging_buffer, 0, size);
     queue.submit(Some(encoder.finish()));
+    let buffer_slice = staging_buffer.slice(..);
+    let buffer_future = buffer_slice.map_async(wgpu::MapMode::Read);
+    device.poll(wgpu::Maintain::Wait);
+    if let Ok(()) = {
+        cfg_if::cfg_if! {
```
termhn, line 131: this feels a bit hard to read. Maybe extract this into a binding above like `let buffer_readback_result = ...`

examples/shaders/compute-shader/src/lib.rs
```diff
@@ -1,17 +1,31 @@
...
+pub fn collatz(mut n: u32) -> u32 {
+    let mut i = 0;
+    while n != 1 {
+        n = if n % 2 == 0 { n / 2 } else { (3 * n) + 1 };
+        i += 1;
+    }
+    i
+}
+
+// LocalSize/numthreads of (x = 1, y = 1, z = 1)
```
termhn, line 35: it would be great to increase this for perf. Probably something like (x = 16, y = 1, z = 1) (or maybe 32) as we'll eventually be ALU bound for high enough n.

termhn (changes requested):

examples/runners/wgpu/build.rs
```diff
@@ -2,7 +2,7 @@ use spirv_builder::SpirvBuilder;
 use std::error::Error;
 
 fn build_shader(path_to_create: &str) -> Result<(), Box<dyn Error>> {
-    SpirvBuilder::new(path_to_create, "spirv-unknown-vulkan1.0").build()?;
+    SpirvBuilder::new(path_to_create, "spirv-unknown-vulkan1.1").build()?;
```
raphlinus, line 5: It's odd to me that this requires Vulkan 1.1. Playing with it, it seems it needs the `VK_KHR_storage_buffer_storage_class` extension, probably because the SPV refers to StorageBuffer. If you look at the [GLSL version](https://github.com/linebender/piet-gpu/blob/master/piet-gpu-hal/examples/shader/collatz.comp) and the [spv it compiles to](https://github.com/linebender/piet-gpu/blob/master/piet-gpu-hal/examples/shader/collatz.spv), it looks pretty different - it's a BufferBlock.

I'm not super expert on this, but it seems like it's using more advanced features than it really needs to, which might hurt compatibility.

khyperia: I forget if we documented this anywhere, but a decision made in rust-gpu is to be pretty opinionated in that we should only support non-deprecated SPIR-V features, and be pretty "modern" in that sense (_for now_ at least - perhaps when we have more fundamentals in place, we can look at supporting older things). `BufferBlock` is explicitly deprecated, replaced with `Block`/`StorageBuffer`, and so we don't plan on supporting `BufferBlock` right now, and so require vulkan 1.1+ for this.

khyperia (approved):

examples/runners/wgpu/src/compute.rs
DJMcNab: I've converted it to a safe version which uses `u32::to_ne_bytes` in 01b6076715e000b3efc65ef2834200f485929eca

examples/runners/wgpu/build.rs
raphlinus: @khyperia Good to know thanks! I've been spending a lot of time looking at compatibility recently because I'm trying to target [phones,](https://developer.android.com/about/dashboards#Vulkan) and very much appreciate the load of dealing with older stuff. Having a clean, well documented line is a good strategy.

DJMcNab:
If these performance counters are done correctly, then we should also add them to the other examples https://github.com/EmbarkStudios/rust-gpu/pull/259

I have also only now realised that I have inadvertently duplicated https://github.com/EmbarkStudios/rust-gpu/pull/360. I was just trying to show @raphlinus that rust-gpu can be used to make compute shaders 😅

VZout (approved):

termhn (approved):

khyperia (approved):

PR #627, opened 2021-05-21
Build the wgpu shader at runtime

khyperia:
This _dramatically_ improves the IDE experience when working on rust-gpu.

### DIFF
--- a/examples/runners/wgpu/build.rs
+++ b/examples/runners/wgpu/build.rs
@@ -1,15 +1,43 @@
+use std::env;
 use std::error::Error;
+use std::path::PathBuf;
 
 fn main() -> Result<(), Box<dyn Error>> {
     let target_os = std::env::var("CARGO_CFG_TARGET_OS")?;
     let target_arch = std::env::var("CARGO_CFG_TARGET_ARCH")?;
     // Always build on CI to make sure the shaders can still build
     let is_on_ci = std::env::var("CI");
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-env-changed=CARGO_CFG_TARGET_OS");
+    println!("cargo:rerun-if-env-changed=CARGO_CFG_TARGET_ARCH");
+    println!("cargo:rerun-if-env-changed=CI");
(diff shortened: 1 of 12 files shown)


VZout:
I think you should leave an option to pick between the two. in the end this is a example for other people and they might want to know they can compile at compile time.

khyperia:
The android version does compile at compile time with this change, that's already an example.

VZout (approved):

termhn (approved):
Nice.

examples/runners/wgpu/src/lib.rs
```diff
@@ -57,6 +57,46 @@ pub enum RustGPUShader {
...
+        let crate_path = [
+            Path::new(manifest_dir),
+            Path::new(".."),
+            Path::new(".."),
+            Path::new("shaders"),
+            Path::new(crate_name),
+        ]
+        .iter()
+        .copied()
+        .collect::<PathBuf>();
```
eddyb, line 81: Isn't this shorter as a chain of `.join(...)` calls? Tho I suppose that may be less efficient hmpf.

khyperia: yeah, probably, idk, I tend to over-optimize things that really I probably should just write the clean readable way :|

examples/runners/wgpu/build.rs
```diff
@@ -1,15 +1,24 @@
...
+    // Always build on CI to make sure the shaders can still build
+    let is_on_ci = std::env::var("CI");
+    if target_os != "android" && target_arch != "wasm32" && is_on_ci.is_err() {
+        return Ok(());
+    }
+    let status = std::process::Command::new("cargo")
+        .args(["run", "--release", "-p", "example-runner-wgpu-builder"])
+        .stderr(std::process::Stdio::inherit())
+        .stdout(std::process::Stdio::inherit())
+        .status()?;
```
eddyb, line 15: Doesn't this deadlock Cargo without passing a different `--target-dir`? Have you tried building `example-runner-wgpu` with `--release` and `CI` being set? It should get stuck AFAICT (given e.g. `touch examples/runners/wgpu/builder/src/main.rs` to force a rebuild of it).

eddyb (approved):

PR #630, opened 2021-05-24
Add API to enable ext/capabilities, and remove default capabilities

khyperia:
I had to add a decent amount of debugging infrastructure to figure out wtf was going wrong when the default capabilities weren't enabled, and I figured I should clean up that infra and leave it in, so there's a decent amount of error message changes in this PR.

Also included is adding `glam` to system crates, because otherwise it barfs on all usages of e.g. `f64` in glam if `Float64` isn't enabled, and I don't want to force glam to write `#[cfg(any(not(target_arch = "spirv"), target_feature = "Float64"))]` everywhere (and Int8, etc.) - it'd be a maintenance nightmare, easily broken.

Also, compiletest gets all of Int8, Int16, Int64, Float64 automatically enabled for both the "library" crates (everything except for the compiletest'ed crate), as well as the final crate. I ranted about this in discord for as to why, I'm just going to copy that in instead of rewriting it:

> uugggh, so, I'm nuking the "default Int8 capability", and running into really really annoying things in compiletest. The "standard library" crates in compiletest (everything except the end crate) is only compiled once, for every test. Say, it's compiled without the Int8 capability, and a test does enable the Int8 capability. This means that in the "standard library" crates, Int8 is zombie'd, however, in the test crate, it is not zombied. This means that if, say, the test crate calls a stdlib function taking a u8, the export in the stdlib has a zombie symbol, but the import does not have a zombie symbol, and so there's a linker type mismatch because two identical types where one is zombie and the other are not are considered distinct types. But, even if we merged, say, only Int8 types where one is zombie and the other isn't, the zombie status of the stdlib would "leak" into the end crate, and u8 would get zombie'd even if the test enables the Int8 capability.
> 
> So uh, running this by [eddyb] for sanity: I've thought about it a bit and I think the best solution is to enable all of Int8,Int16,Int64,Float64 for the compiletest "stdlib" and every test, to make sure there's no mismatches. It does mean that testing that particular part of the capability system would be impossible (since it's always enabled, we can't test an expected-failure), but, it should be fine, idk. This situation **won't ever happen in the real world**, since all crates in the graph are always compiled with the same capabilities.


crates/rustc_codegen_spirv/src/builder_spirv.rs
```diff
@@ -425,6 +421,18 @@ impl BuilderSpirv {
...
+    pub fn has_extension(&self, extension: &str) -> bool {
+        self.builder
+            .borrow()
+            .module_ref()
+            .extensions
+            .iter()
+            .any(|inst| {
+                inst.class.opcode == Op::Extension
+                    && inst.operands[0].unwrap_literal_string() == extension
+            })
```
eddyb, line 433: This can be `O(1)` as a set of interned strings, maybe open an issue about simple algorithmic inefficiencies like this and leave `// FIXME(#1234)` comments? At least that's what I ~~would~~ should do.

crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
```diff
@@ -96,12 +96,22 @@ impl<'tcx> CodegenCx<'tcx> {
...
             .iter()
             .filter(|s| *s != &sym.bindless)
-            .map(|s| s.as_str().parse())
+            .map(|s| s.as_str())
+            .collect::<Vec<_>>();
+
+        // target_features is a HashSet, not a Vec, so we need to sort to have deterministic
+        // compilation - otherwise, the order of capabilities in binaries depends on the iteration
+        // order of the hashset. Sort by the string, since that's easy.
+        feature_names.sort();
```
eddyb, line 110: It's a `FxHashSet`, right? Its order should already be deterministic. But yes this is probably a good idea just to keep the output clean, if nothing else.

examples/runners/wgpu/builder/src/main.rs
```diff
@@ -15,9 +23,13 @@ fn build_shader(path_to_create: &str, codegen_names: bool) -> Result<(), Box<dyn
...
-    build_shader("../../shaders/simplest-shader", false)?;
-    build_shader("../../shaders/compute-shader", false)?;
-    build_shader("../../shaders/mouse-shader", false)?;
+    build_shader("../../shaders/sky-shader", true, &[])?;
+    build_shader("../../shaders/simplest-shader", false, &[])?;
+    build_shader("../../shaders/compute-shader", false, &[])?;
+    build_shader(
+        "../../shaders/mouse-shader",
+        false,
+        &[Capability::Int8, Capability::Int16, Capability::Int64],
```
eddyb, line 32: Maybe open an issue for this? I think it's just one of those `num-traits` things like `.abs()`?

examples/runners/wgpu/src/lib.rs
```diff
@@ -70,11 +70,14 @@ fn shader_module(shader: RustGPUShader) -> wgpu::ShaderModuleDescriptor<'static>
...
-            RustGPUShader::Sky => "simplest-shader",
-            RustGPUShader::Compute => "compute-shader",
-            RustGPUShader::Mouse => "mouse-shader",
+        let (crate_name, capabilities): (_, &[Capability]) = match shader {
+            RustGPUShader::Simplest => ("sky-shader", &[]),
+            RustGPUShader::Sky => ("simplest-shader", &[]),
+            RustGPUShader::Compute => ("compute-shader", &[]),
+            RustGPUShader::Mouse => (
+                "mouse-shader",
+                &[Capability::Int8, Capability::Int16, Capability::Int64],
```
eddyb, line 79: Oof, the duplication here isn't great. There's no easy way to share this information, is there?

eddyb (approved):

examples/runners/wgpu/src/lib.rs
khyperia: My only thought was to make a common file and `include!()` it from both (or `#[path = "..."] mod blah;` or whatever), but I'm not sure if that makes it better or worse.

examples/runners/wgpu/builder/src/main.rs
khyperia: https://github.com/EmbarkStudios/rust-gpu/issues/634

khyperia:
I'm going to leave https://github.com/EmbarkStudios/rust-gpu/issues/633 as just an issue instead of a comment in code ~~so I don't have to run 45-minute CI again~~

PR #632, opened 2021-05-26
Use `rust-toolchain.toml` instead

DJMcNab:
Note: rustup 1.24.0 (which adds support for this) was released on 2021-04-27, i.e. 29 days ago

Unfortunately, I don't think this will degrade gracefully for those using older rustups, but I expect that should be rare.

Note also that we started to depend on the toml format in #284, on 2020-12-07, around 10 days after rustup 1.23 (2020-11-27) released.

### DIFF
--- a/rust-toolchain
+++ b/rust-toolchain
@@ -1,9 +1,0 @@
-# If you see this, run `rustup self update` to get rustup 1.23 or newer.
-
-# NOTE: above comment is for older `rustup` (before TOML support was added),
-# which will treat the first line as the toolchain name, and therefore show it
-# to the user in the error, instead of "error: invalid channel name '[toolchain]'".
-
-[toolchain]
-channel = "nightly-2021-05-24"
-components = ["rust-src", "rustc-dev", "llvm-tools-preview"]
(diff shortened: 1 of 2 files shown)


khyperia:
Not sure if we want to wait longer on this, there was a _lot_ of pain due to outdated rustup on the original toml-ification, and we even had the graceful fallback of the error message telling you to update.

> Note also that we started to depend on the toml format in #284, on 2020-12-07, around 10 days after rustup 1.23 (2020-11-27) released.

The reason the original change was merged only 10 days after release is because it contained critical behavior differences (inclusion of required components), whereas this change is much less urgent as it doesn't contain any new functionality.

Is it possible to include both `rust-toolchain` and `rust-toolchain.toml` files, and have the `rust-toolchain` file contain a comment that causes a syntax error that tells you to update rustup? That would solve my concerns.

DJMcNab:
Unfortunately, rustup prioritises `rust-toolchain` over `rust-toolchain.toml` - that is, we could do that, but the result would be that rustup only uses the rust-toolchain file.

It would appear that this is not going to be workable at the moment.

Issue #628, opened 2021-05-21
Need a way to disable uniform buffer layout validation in spirv-val


hrydgard:
Vulkan implementations can support more relaxed layouts for uniform buffers (and also storage buffers) than the usual std140 or std430. spirv-val needs to be told, though, to not validate, if you intend to run your SPIR-V output on a Vulkan implementation that supports these (which is most, nowadays).

See https://github.com/KhronosGroup/SPIRV-Tools/issues/3153

PR #635, opened 2021-05-26
Add spirv-val flags to spirv-builder

khyperia:
Fixes https://github.com/EmbarkStudios/rust-gpu/issues/628

Went a little overboard and added all the spirv-val flags instead of just `--scalar-block-layout`, figured we'd want them eventually.

Also, fixed a nasty bug where spirv-opt wasn't being passed the target environment (but spirv-val was), that was a little spooky to find.

### DIFF
--- a/crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
@@ -277,6 +277,14 @@ pub struct CodegenArgs {
     pub disassemble_fn: Option<String>,
     pub disassemble_entry: Option<String>,
     pub disassemble_globals: bool,
+
+    // spirv-val flags
+    pub relax_struct_store: bool,
+    pub relax_logical_pointer: bool,
+    pub relax_block_layout: Option<bool>,
+    pub uniform_buffer_standard_layout: bool,
+    pub scalar_block_layout: bool,
+    pub skip_block_layout: bool,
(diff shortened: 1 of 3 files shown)


crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
```diff
@@ -277,6 +277,14 @@ pub struct CodegenArgs {
     pub disassemble_fn: Option<String>,
     pub disassemble_entry: Option<String>,
     pub disassemble_globals: bool,
+
+    // spirv-val flags
+    pub relax_struct_store: bool,
+    pub relax_logical_pointer: bool,
+    pub relax_block_layout: Option<bool>,
```
hrydgard, line 284: So this is basically a tri-state? It stands out a bit here, maybe deserves a comment (although this is kind of a deep location).

khyperia: Yeah, it's... a little weird? Even weirder is that I don't think the CLI of `spirv-val` exposes this tri-state, it's only `Some(true)` and `None`. I thiiiink the tri-state is:

- `Some(true)` -> force relax-block-layout on
- `Some(false)` -> force relax-block-layout off
- `None` -> relax-block-layout is on if vulkan version >=1.1, off if ==1.0

hrydgard (approved):

PR #637, opened 2021-05-27
Try reducing CI times

khyperia:
Trying to build rustc_codegen_spirv fewer times, and running cargo check rather than cargo build/test on the examples

There's still work left to do, as e.g. `syn` is built five (5!!) times, but it's better.


khyperia:
Shoot, sorry, meant to open this as a draft

khyperia:
Looks like CI times are cut in roughly half with this change! \o/

PR #638, opened 2021-05-31
Use `legacy` mangling instead of `v0`.

ElectronicRU:
`v0` mangling mangles generics, but can only handle const generic
arguments of simple types. As `Image!` types use const generic enums,
things break horribly and compiler panics.
`legacy` doesn't even attempt to mangle generics, which is probably
fine.

### DIFF
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -264,6 +264,7 @@ fn invoke_rustc(builder: &SpirvBuilder) -> Result<PathBuf, SpirvBuilderError> {
         format!(" -C target-feature={}", target_features.join(","))
     };
 
+    //FIXME: reintroduce v0 mangling, see issue #642
     let rustflags = format!(
         "-Z codegen-backend={} -Zsymbol-mangling-version=legacy{}{}",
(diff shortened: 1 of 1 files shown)


crates/spirv-builder/src/lib.rs
```diff
@@ -265,7 +265,7 @@ fn invoke_rustc(builder: &SpirvBuilder) -> Result<PathBuf, SpirvBuilderError> {
     };
 
     let rustflags = format!(
-        "-Z codegen-backend={} -Zsymbol-mangling-version=v0{}{}",
+        "-Z codegen-backend={} -Zsymbol-mangling-version=legacy{}{}",
```
eddyb, line 268: You can just remove it if you want. If you want to keep it in, can you open an issue about switching it back to `v0`, and leave a `// FIXME(#1234) ...` comment here? Thanks.

eddyb (changes requested):

crates/spirv-builder/src/lib.rs
ElectronicRU: Done and done!

eddyb (approved):

PR #643, opened 2021-06-01
Added read_subpass for SubpassData images, and necessary attrs for it.

ElectronicRU:
Decided to call the attribute attachment_index since that is shorter than input_attachment_index, and output attachment indices are just locations, so there's no ambiguity.

Tell me if I should add more tests for error messages!

### DIFF
--- a/crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
@@ -680,25 +680,28 @@ impl<'tcx> CodegenCx<'tcx> {
             ),
             _ => false,
         };
-        if let Some(attachment_index) = attrs.input_attachment_index.map(|attr| attr.value) {
-            if is_subpass_input {
-                self.emit_global().capability(Capability::InputAttachment);
+        if let Some(attachment_index) = attrs.input_attachment_index {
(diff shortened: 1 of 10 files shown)


hrydgard:
I think there's a case for keeping "input_" in the name actually, input attachments are such an oddball Vulkan-only thing while attachments are everywhere.

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -663,6 +665,43 @@ impl<'tcx> CodegenCx<'tcx> {
...
+                SpirvType::Image {
+                    dim: Dim::DimSubpassData,
+                    ..
+                }
+            ),
+            _ => false,
+        };
+        if let Some(attachment_index) = attrs.input_attachment_index.map(|attr| attr.value) {
+            if is_subpass_input {
+                self.emit_global().capability(Capability::InputAttachment);
```
khyperia, line 685: Ah, with https://github.com/EmbarkStudios/rust-gpu/pull/630, we no longer automatically add capabilities for users, as it's a really bad user experience. Instead, check `self.builder.has_capability(Capability::InputAttachment)`, and emit an error if false.

khyperia (commented):
LGTM, just the one thing!

khyperia (approved):

PR #648, opened 2021-06-04
Add SHADERed Rust link to Readme

PrototypeNM1:
Increase visibility of SHADERed's rust-gpu integration by adding it to the Readme.


PrototypeNM1:
SHADERed's rust-gpu integration is sponsored by Embark but is difficult to find short of looking through old release notes. It doesn't currently show up in Google search results for shadertoy-like tools for Rust.

khyperia (approved):
Thanks!

PR #652, opened 2021-06-07
Add Image! docs to book

### LINKED ISSUE #603: Document `Image!` in book
Continuing from #359, there was a desire to also include documentation about the `Image` macro syntax in the rust-gpu book.

khyperia:
Fixes #603

### DIFF
--- a/docs/src/image.md
+++ b/docs/src/image.md
@@ -1,0 +1,20 @@
+# Image type syntax
+
+There are a huge number of combinations of image types in SPIR-V. They are represented by a const
+generic type called `spirv_std::image::Image`, however, specifying the generic parameters of this
+type is incredibly tedious, so a wrapper macro, `spirv_std::Image!` can be used to write the type
+instead.
+
+The specific syntax and meaning of the arguments to the `Image!` macro can be found in
+[rustdoc](https://embarkstudios.github.io/rust-gpu/api/spirv_std/macro.Image.html).
+
+Some type aliases for common image formats can be found in the `spirv_std::image` module. For
+example, `Image2d` is a very commonly used type, corresponding to `texture2D` in GLSL, and is likely
+what you want if you want a regular old sampled texture.
+
+```rust,no_run
+type Image2d = Image!(2D, type=f32, sampled);
+```
+
+Note that the `const-generics` feature must be enabled to use the `Image!` macro at the moment.
+This will likely change in the near future.
--- a/docs/src/image.md
+++ b/docs/src/image.md
@@ -8,13 +8,19 @@ instead.
 The specific syntax and meaning of the arguments to the `Image!` macro can be found in
 [rustdoc](https://embarkstudios.github.io/rust-gpu/api/spirv_std/macro.Image.html).
 
-Some type aliases for common image formats can be found in the `spirv_std::image` module. For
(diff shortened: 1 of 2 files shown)


docs/src/image.md
```diff
@@ -0,0 +1,20 @@
...
+
+There are a huge number of combinations of image types in SPIR-V. They are represented by a const
+generic type called `spirv_std::image::Image`, however, specifying the generic parameters of this
+type is incredibly tedious, so a wrapper macro, `spirv_std::Image!` can be used to write the type
+instead.
+
+The specific syntax and meaning of the arguments to the `Image!` macro can be found in
+[rustdoc](https://embarkstudios.github.io/rust-gpu/api/spirv_std/macro.Image.html).
+
+Some type aliases for common image formats can be found in the `spirv_std::image` module. For
```
DJMcNab, line 11: Could this also link to the module docs? i.e. something like
```suggestion
Some type aliases for common image formats can be found in the [`spirv_std::image`](https://embarkstudios.github.io/rust-gpu/api/spirv_std/index.html) module. For
```

However, something strange is afoot. [That page](https://embarkstudios.github.io/rust-gpu/api/spirv_std/index.html) lists the version as `0.4.0-alpha.9`, the latest version. However, it does not list the `image` module as a seperate module, and lists all the image types as types at the root.

Also, the [src] of `Image2d` is [shown as this](https://embarkstudios.github.io/rust-gpu/api/src/spirv_std/textures.rs.html#19-21), which is afaik not correct.

docs/src/image.md
```diff
@@ -0,0 +1,20 @@
...
+example, `Image2d` is a very commonly used type, corresponding to `texture2D` in GLSL, and is likely
+what you want if you want a regular old sampled texture.
+
+```rust,no_run
+type Image2d = Image!(2D, type=f32, sampled);
```
DJMcNab, line 16:
```suggestion
// In spirv_std::image
pub type Image2d = Image!(2D, type=f32, sampled);
```

docs/src/image.md
```diff
@@ -0,0 +1,20 @@
...
+```
+
+Note that the `const-generics` feature must be enabled to use the `Image!` macro at the moment.
```
DJMcNab, line 19: It [looks like](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=001f2614aba60185e53c11f63baee177) the feature name is `const_generics`.

````suggestion
Note that the `const_generics` feature must be enabled to use the `Image!` macro at the moment.
```rust
#![feature(const_generics)]
// ...
```
````

DJMcNab (approved):
This looks good, although I might have found the need for a debugging mission into our docs generation. Sorry 😅.

docs/src/image.md
khyperia: This is because the docs are not generated with the `const_generics` feature, and so the new `Image2d` doesn't show up (the new one is in the `image` module)

DJMcNab: Oh, I see sorry. That also solves my other confusion.

docs/src/image.md
DJMcNab: I see now that this is a cargo feature, so this should be a Cargo.toml example instead.

hrydgard (approved):

PR #653, opened 2021-06-07
Produce spirv-std API docs with all features

XAMPPRocky:
Fixes issue reported here: https://github.com/EmbarkStudios/rust-gpu/pull/652#discussion_r646601665


DJMcNab (approved):
I don't know if there's any way to validate that it generates the correct thing without merging and seeing what happens, but this looks correct.

khyperia (approved):

PR #654, opened 2021-06-08
fix trace ray acceleration structure type

evopen:
Looks like `TraceRayKHR` accepts acceleration structure as a handle not a pointer. Previously the generated spirv didn't do `OpLoad` for AS param and crashed `vkCreateGraphicsPipelines` with `ERROR_INITIALIZATION_FAILED`.

In SPIRV specs, they use `descriptor` for AS param and `pointer` to describe `payload` param. It's confusing. But I guess they are not the same. It's the first time that I write any SPIRV assembly, so prove me wrong.


khyperia (approved):
Thanks!

PR #655, opened 2021-06-08
👀 hot shader reloading

DJMcNab:
Add a (feature gated) implementation of hot shader reloading to SpirvBuilder

This attempts to handle initial compilation gracefully by watching the shader crate root recursively until we have more detailed dependency information

Unfortunately, notify uses a non-standard license (CC0) so we fail the cargo deny check. I will defer adding a commit which changes `deny.toml` to allow this to pass to the suitable Embarker (@XAMPPRocky ?)

It appears that if there was sufficient demand for it, notify would be willing to update their license ( https://github.com/notify-rs/notify/issues/243#issuecomment-792078532)

In theory, the CC0 license they are under is probably already MIT/Apache compatible, so `notify` could avoid getting new permissions from the contributors, but it would still be wise to notify (ha) them.

(See e.g. [this stackoverflow post](https://webcache.googleusercontent.com/search?q=cache:fophIpDX5Z0J:https://opensource.stackexchange.com/questions/4235/how-do-i-turn-a-public-domain-project-into-a-mit-project+&cd=10&hl=en&ct=clnk&gl=uk&client=firefox-b-d) (this is google's webcache since stackoverflow is down at the time of posting))

I'm also planning to integrate this into the example runner, but want to get the ball rolling on the license issues

### DIFF
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -455,9 +455,9 @@ fn invoke_rustc(builder: &SpirvBuilder) -> Result<PathBuf, SpirvBuilderError> {
     // we do that even in case of an error, to let through any useful messages
     // that ended up on stdout instead of stderr.
     let stdout = String::from_utf8(build.stdout).unwrap();
+    let artifact = get_last_artifact(&stdout);
     if build.status.success() {
-        let artifact = get_last_artifact(&stdout);
-        Ok(artifact)
+        Ok(artifact.expect("Artifact created when compilation succeeded"))
     } else {
         Err(SpirvBuilderError::BuildFailed)
     }
@@ -469,7 +469,7 @@ struct RustcOutput {
     filenames: Option<Vec<String>>,
 }
 
-fn get_last_artifact(out: &str) -> PathBuf {
+fn get_last_artifact(out: &str) -> Option<PathBuf> {
     let last = out
         .lines()
         .filter_map(|line| match serde_json::from_str::<RustcOutput>(line) {
@@ -489,9 +489,9 @@ fn get_last_artifact(out: &str) -> PathBuf {
         .unwrap()
         .into_iter()
         .filter(|v| v.ends_with(".spv"));
-    let filename = filenames.next().expect("Crate had no .spv artifacts");
+    let filename = filenames.next()?;
     assert_eq!(filenames.next(), None, "Crate had multiple .spv artifacts");
-    filename.into()
+    Some(filename.into())
 }
 
 /// Internally iterate through the leaf dependencies of the artifact at `artifact`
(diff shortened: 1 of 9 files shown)


DJMcNab:
I have produced a video showing how this works in the wgpu example runner:



https://user-images.githubusercontent.com/36049421/121253502-186cb780-c8a1-11eb-8755-1960157c8809.mp4

crates/spirv-builder/src/lib.rs
```diff
@@ -423,13 +455,8 @@ fn invoke_rustc(builder: &SpirvBuilder) -> Result<PathBuf, SpirvBuilderError> {
...
     // that ended up on stdout instead of stderr.
     let stdout = String::from_utf8(build.stdout).unwrap();
-    let artifact = get_last_artifact(&stdout);
-
     if build.status.success() {
-        match builder.print_metadata {
-            MetadataPrintout::Full | MetadataPrintout::DependencyOnly => print_deps_of(&artifact),
-            MetadataPrintout::None => (),
-        }
+        let artifact = get_last_artifact(&stdout);
```
khyperia, line 459: Please leave this outside the `build.status.success` if statement, due to the comment above it.

examples/runners/wgpu/src/lib.rs
```diff
@@ -56,10 +58,15 @@ pub enum RustGPUShader {
     Mouse,
 }
 
-fn shader_module(shader: RustGPUShader) -> wgpu::ShaderModuleDescriptor<'static> {
+fn maybe_watch(
+    shader: RustGPUShader,
+    force_no_watch: bool,
+) -> Receiver<wgpu::ShaderModuleDescriptor<'static>> {
+    // This bound needs to be 1, because when we don't watch, we
```
khyperia, line 65: Strange comment

khyperia (commented):
LGTM, just the two things!

khyperia:
> Unfortunately, notify uses a non-standard license (CC0) so we fail the cargo deny check. I will defer adding a commit which changes `deny.toml` to allow this to pass to the suitable Embarker

Pushed the exception!

crates/spirv-builder/src/lib.rs
DJMcNab: I found that when If compilation fails, `get_last_artifact` panics, which meant that hot reloading couldn't work. I guess I can make it return an `Option`?

examples/runners/wgpu/src/lib.rs
DJMcNab: Oh yeah, I'll fix that. It's talking about e.g. line 98 of this file, where the item is added to the channel on the same thread.

crates/spirv-builder/src/lib.rs
khyperia: Ah, haha, yep, that's unfortunate, rip my bad code :P - yes, please make it an `Option`

DJMcNab: Handled

khyperia (approved):

Issue #614, opened 2021-05-05
Bindless follow-up


Jasper-Bekkers:
This is a tracking issue to inventory future work that can/should happen on the bindless API.

- [ ] `StorageTexture`
- [ ] Remaining functions on `Texture`
- [ ] Add `Texture1d` and `Texture3d`
- [ ] Test coverage (and a runner to go along with it)
- [ ] Add NonUniformResourceIndex equivalent
- [ ] Switch from SampledImages to images + static samplers

PR #656, opened 2021-06-08
mcp: dynamic bindless resource access

msiglreith:
## Overview
- Add a `resource_access(idx) -> T` intrinsic which acquires a descriptor/resource from the global bindless runtime descriptor arrays.
- Support templated load/store for all `RuntimeArray<u32>`, not requiring the bindless feature for it anymore
- Support bindless TLAS and samplers
- Auto-generate entry point interface variables from callgraph

## Dynamic Resource Access
Inspired by https://devblogs.microsoft.com/directx/in-the-works-hlsl-shader-model-6-6/#dynamic-resource-binding

Currently, a lot of manual asm is required for adding support for texture types, samplers, tlas in case of bindless. It also fractures libraries as functions taking 'normal' resources as arguments would be not compatible with bindless.
It also allows to easily expand image support using the existing `Image!` macro for bindless allowing to keep the `spirv-std` footprint small.

Currently support types `T` for `resource_access`:
```rust
Image<..>
Sampler
AccelerationStructure

// Buffers
&RuntimeArray<u32>
&mut RuntimeArray<u32>
```

Example:
```rust
#[repr(C)]
#[derive(Copy, Clone)]
pub struct InstanceData {
    sampler: RenderResourceHandle,
    albedo_map: RenderResourceHandle,
}

#[spirv(fragment)]
pub unsafe fn mesh_fs(
    a_texcoord: f32x2,
    output: &mut f32x4,
    #[spirv(push_constant)] constants: &Constants,
) {
    let instance_data = constants.instance.load();
    let albedo: Image2d = instance_data.albedo_map.access();
    let sampler: Sampler = instance_data.sampler.access();

    let color: f32x4 = albedo.sample(sampler, a_texcoord);
    *output = vec4(color.x, color.y, color.z, 1.0);
}
```

##  Templated Store/Load
This feature existed beforehand but was quite intervowen with `bindless`. To couple it with the dynamic resource access functionality, I moved it out of the `bindless` feature and changed it to support it for `RuntimeArray<u32>` in general.

## Entrypoint interface variables

Depending on SPIR-V target version entrypoint need either Input/Output or all non-Function global variables accessed.
To make this a bit more of a non-brainer I added an extra pass at the end to patch up the entry point definitions based on the callgraph. One issue was that not using a specific bindless set could result in specialization errors as it gets not removed by DCE I guess.

Affects https://github.com/EmbarkStudios/rust-gpu/issues/614

_sry for the quite lenghty PR.. ):_

### DIFF
--- a/crates/rustc_codegen_spirv/src/attr.rs
+++ b/crates/rustc_codegen_spirv/src/attr.rs
@@ -102,6 +102,7 @@ pub enum SpirvAttribute {
     UnrollLoops,
     InternalBufferLoad,
     InternalBufferStore,
+    ResourceAccess,
 }
 
 // HACK(eddyb) this is similar to `rustc_span::Spanned` but with `value` as the
@@ -137,6 +138,7 @@ pub struct AggregatedSpirvAttributes {
     pub unroll_loops: Option<Spanned<()>>,
     pub internal_buffer_load: Option<Spanned<()>>,
     pub internal_buffer_store: Option<Spanned<()>>,
+    pub resource_access: Option<Spanned<()>>,
 }
 
 struct MultipleAttrs {
@@ -235,6 +237,12 @@ impl AggregatedSpirvAttributes {
                 (),
                 span,
                 "#[spirv(internal_buffer_store)]",
+            ),
+            ResourceAccess => try_insert(
+                &mut self.resource_access,
+                (),
+                span,
+                "#[spirv(resource_access)]",
             ),
         }
     }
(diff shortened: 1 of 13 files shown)


khyperia:
Unfortunately we don't have the resources to design and evaluate MCPs of this size at the moment, and so we won't be able to accept this. I'm really sorry. In the future, it might be helpful to collaborate with us before writing a lot of code, to make sure we're on the same page and can accept work.

PR #659, opened 2021-06-09
Add sampled to example

khyperia:


hrydgard:
Would be nice to see a direct link to [Image](https://embarkstudios.github.io/rust-gpu/api/spirv_std/image/struct.Image.html) somewhere on the page, so you can see what methods it has, too.

hrydgard (approved):

PR #660, opened 2021-06-09
Add name_variables option to spirv-builder

### LINKED ISSUE #657: Don't strip OpName from variables with descriptor set metadata
As we have seen, stripping OpName from things is needed to avoid huge binary bloat.

However, keeping OpName for things that are clearly exposed, like variables with descriptor set metadata, can be helpful for various automatic binding schemes, or just debugging bindings in general when writing code that interacts with the SPIR-V. And only keeping it for these will not be too bloaty.

I suppose it can also help with using the emitted SPIR-V in OpenGL in various ways, although that's not necessarily a goal right now.

khyperia:
Fixes #657

I opted for the simpler approach of "just keep names for _all_ global OpVariables", to cover the use case of maybe wanting input/output names too, as including those don't bloat the binary much at all (the real issue is all the type/function names, AFAIK)

### DIFF
--- a/crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
@@ -277,6 +277,8 @@ pub struct CodegenArgs {
     pub disassemble_entry: Option<String>,
     pub disassemble_globals: bool,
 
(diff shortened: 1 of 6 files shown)


hrydgard (approved):
Great!

PR #661, opened 2021-06-09
Nuke const generics feature

khyperia:
Our internal engine is now able to use the `Image!` macro fully, which is why this was behind a feature gate in the first place, so the feature is no longer needed.


DJMcNab:
How does this interact with building `spirv-std` CPU side on stable, e.g. for the CPU example runner? 

Is that a property we want to have?

khyperia:
Correct, the CPU example runner (and everything else) will no longer be able to be built on stable. It's already built on nightly, though, so that's not an issue for the rust-gpu repo. Our internal engine is also on stable, yet we have the shaders being built on only nightly (via a decoupling similar to the wgpu example runner's builder crate, and excluding the shader from the workspace), so others can do the same if they have a strong need to run on stable.

PR #662, opened 2021-06-10
Add common image type aliases

### LINKED ISSUE #644: Add type aliases for common `Image!` formats
For example, `pub type Image2d = Image!(2D, type=f32, sampled);`. I'd want to add aliases for all the names we have now, probably plus a few more (`Image1d`, `Image3d`, etc.).

Advantages:

- The `Image!` macro is a little verbose and wordy for something as common as a 2d texture. Users can write their own typedefs, but that seems to be rare to actually do in practice and if the typedef were in spirv-std, the typedef would be used far more often.
- It lets people get up and running faster: instead of needing to figure out the `Image!` macro (reading docs, etc.), they can just use their previous graphics knowledge of a type called "image2d"/"texture2d" and use that off the bat.
- It's a nice learning example for people goto-def'ing: seeing practical differences between how `type Image2d` and `type StorageImage` differ in their `Image!` arguments is a nice teaching example.

Disadvantages:
- It doesn't force people to learn the `Image!` macro, and some people might miss how flexible `Image!` actually is.
- I wouldn't have to write `Image!` with an exclamation point as much, which, [as `SHARKS!` has taught us](https://www.youtube.com/watch?v=_8xhdL8BPvU), is clearly a bad thing

---

Not sure if this is actually deserving of an MCP, but I'm using the tag as basically a reminder that I want to discuss it in the weekly meeting, as I think I do want to discuss this in the weekly meeting.

khyperia:
Fixes #644

I kind of just guessed on what types to include, any suggestions to include/remove some are welcome.

### DIFF
--- a/crates/spirv-std/macros/src/lib.rs
+++ b/crates/spirv-std/macros/src/lib.rs
@@ -67,10 +67,10 @@ use quote::ToTokens;
 ///
 /// The grammar for the macro is as follows:
 ///
-/// ```no_compile
+/// ```rust,no_compile
 /// Image!(
 ///     <dimensionality>,
-///     <type|format>,
+///     <type=...|format=...>,
 ///     [sampled[=<true|false>],]
 ///     [multisampled[=<true|false>],]
 ///     [arrayed[=<true|false>],]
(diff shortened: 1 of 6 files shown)


crates/spirv-std/macros/src/lib.rs
```diff
@@ -67,38 +67,40 @@ use quote::ToTokens;
...
+///    Accepted values: `f32`, `f64`, `u8`, `u16`, `u32`, `u64`, `i8`, `i16`, `i32`, `i64`.
 /// - `format` — The image format of the image, mutually exclusive with `type`.
-///   Accepted values: Camel case versions of [`ImageFormat`].
-/// - `sampled` — Whether it is known that the image will be used with a sampler
-///   at compile time. Accepted values: `true` or `false`. Default: `unknown`.
-/// - `multisampled` — Whether the image contains multisampled content. Accepted
-///    values: `true` or `false`. Default: `false`.
-/// - `arrayed` — Whether the image contains arrayed content. Accepted
-///    values: `true` or `false`. Default: `false`.
+///    Accepted values: Camel case versions of [`ImageFormat`].
```
DJMcNab, line 97: [These](https://github.com/EmbarkStudios/rust-gpu/blob/c3eff4d4a5a52607aa3bc8b4ba8409f6659425ee/crates/spirv-std/macros/src/image.rs#L372-L412) don't look camel case to me

PR #668, opened 2021-06-11
Use bytemuck for the push constants

DJMcNab:


DJMcNab:
If https://github.com/Lokathor/bytemuck/pull/67 lands, then we won't need to have this behind a feature

Lokathor:
it landed!

khyperia (approved):

PR #670, opened 2021-06-13
Use `ControlFlow::Poll` for the Mouse shader

DJMcNab:
I thought that the mouse shader was extremely resource
intensive because it was so laggy

However, it turns out that it was just because it was using `Wait`


khyperia:
I believe this was intentional.

DJMcNab:
Before: 

https://user-images.githubusercontent.com/36049421/121875193-17ec7a80-cd00-11eb-8aa4-6e2f9a726e19.mp4

After:


https://user-images.githubusercontent.com/36049421/121875225-2044b580-cd00-11eb-81f4-aca07c5ea624.mp4

PR #673, opened 2021-06-17
Clean up DUMP environment variables

khyperia:
1) `DUMP_POST_SPLIT` and `DUMP_POST_LINK` didn't support multimodule and would write to the same file over and over, so fixed that.
2) Cleaned up some directory handling and put it into a helper function
3) Documented the various environment variables rust-gpu uses.


eddyb (approved):
LGTM. It also reminds me it might be nice to prefix the env vars with `RUSTC_CG_SPIRV_` or something like that, but that's neither here nor there.

PR #674, opened 2021-06-17
Cargo update

khyperia:
Primarily intended to update glam from 0.15.2 to 0.16.0, but figured I'd update everything else in the process.


khyperia:
Weirdly, this breaks the compute shader example's timer. Probably a wgpu bug.

eddyb (approved):

DJMcNab:
Yeah, I did try using wgpu 0.8, but the timer was broken for me too - haven't looked into what's wrong yet though.

PR #676, opened 2021-06-19
Upgrade wgpu

DJMcNab:
Wgpu 0.9 has released, now under MIT/Apache 2.0 🎉 

~~We can't remove MPL-2.0 from the deny.toml, because wgpu_subscriber is still under MPL.~~

### DIFF
--- a/examples/runners/wgpu/src/compute.rs
+++ b/examples/runners/wgpu/src/compute.rs
@@ -113,8 +113,9 @@ pub async fn start_internal(
         label: Some("Timestamps buffer"),
         size: 16,
         usage: wgpu::BufferUsage::MAP_READ | wgpu::BufferUsage::COPY_DST,
-        mapped_at_creation: false,
+        mapped_at_creation: true,
     });
+    timestamp_buffer.unmap();
 
     let bind_group = device.create_bind_group(&wgpu::BindGroupDescriptor {
         label: None,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -765,6 +765,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9ea835d29036a4087793836fa931b08837ad5e957da9e23886b29586fb9b6650"
 
 [[package]]
+name = "drm-fourcc"
+version = "2.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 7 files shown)


DJMcNab:
I've also tried to enable validation again, without success

Naga still chokes on our shaders, with exceedingly unhelpful error messages.

khyperia:
@VZout I vaguely recall there was a reason you used `wgpu-subscriber` instead of `env_logger` in https://github.com/EmbarkStudios/rust-gpu/pull/92, am I misremembering and this change is fine, or is there a reason?

DJMcNab:
I think wgpu moved to just log in https://github.com/gfx-rs/wgpu/issues/1211#issue-806966118

examples/runners/wgpu/src/compute.rs
```diff
@@ -113,8 +113,9 @@ pub async fn start_internal(
         label: Some("Timestamps buffer"),
         size: 16,
         usage: wgpu::BufferUsage::MAP_READ | wgpu::BufferUsage::COPY_DST,
-        mapped_at_creation: false,
+        mapped_at_creation: true,
     });
+    timestamp_buffer.unmap();
```
VZout, line 118: Why map and then immediately unmap?

VZout (changes requested):

VZout:
> @VZout I vaguely recall there was a reason you used `wgpu-subscriber` instead of `env_logger` in #92, am I misremembering and this change is fine, or is there a reason?

The reason we used `wgpu-subscriber` was because back then the triangle example in the wgpu repo used `wgpu-subscriber`. I'm not aware of any other reason. I think env logger is better anyway. gives you pretty colors.

examples/runners/wgpu/src/compute.rs
DJMcNab: That's required to actually get values from the queries - there's a wgpu issue about it, which I can't find at the moment

VZout: 🤦

DJMcNab: https://github.com/gfx-rs/wgpu/issues/1374#issuecomment-839359307

VZout (approved):

khyperia (approved):

Issue #677, opened 2021-06-19
`core::Option::<u32>::unwrap_or` generates code which requires the Int8 capability


DJMcNab:
### Expected Behaviour
The Int8 capability not to be needed

### Example & Steps To Reproduce
I tried to build our compute example:
https://github.com/EmbarkStudios/rust-gpu/blob/8c05636f5328e1ea5c88194e64c8e657ce572524/examples/shaders/compute-shader/src/lib.rs#L1-L48
without the Int8 Capability:
https://github.com/EmbarkStudios/rust-gpu/blob/8c05636f5328e1ea5c88194e64c8e657ce572524/examples/runners/wgpu/src/lib.rs#L107

And got the following error:
```
error: u8 without OpCapability Int8
    --> C:\Programs\Rustup\toolchains\nightly-2021-06-09-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core\src\ptr\mod.rs:1136:1
     |
1136 | pub(crate) unsafe fn align_offset<T: Sized>(p: *const T, a: usize) -> usize {
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: Stack:
             core::option::Option<T>::unwrap_or
             compute_shader::main_cs
             Unnamed function ID %27
```

(that span also seems wrong?)

The code of `unwrap_or` is 
```rust
    #[inline]
    #[stable(feature = "rust1", since = "1.0.0")]
    pub fn unwrap_or(self, default: T) -> T {
        match self {
            Some(x) => x,
            None => default,
        }
    }
```
which obviously does not use the `align_offset` function.

A minimised reproduction is:
```rust
#![cfg_attr(
    target_arch = "spirv",
    feature(register_attr),
    register_attr(spirv),
    no_std
)]
// HACK(eddyb) can't easily see warnings otherwise from `spirv-builder` builds.
#![deny(warnings)]

extern crate spirv_std;

#[cfg(not(target_arch = "spirv"))]
use spirv_std::macros::spirv;

// Adapted from the wgpu hello-compute example

// LocalSize/numthreads of (x = 64, y = 1, z = 1)
#[spirv(compute(threads(64)))]
pub fn main_cs() {
    None.unwrap_or(0);
}
```
[...]

PR #680, opened 2021-06-21
Handle booleans *slightly* hackily without `Int8`

DJMcNab:
Unlocks use of `unwrap_or` in all shaders

This is obviously a massive hack, and I'm sure this opens up the ability to meet spir-v val errors.

This is also a proof of concept that a solution to #677 might be possible without resorting to codegenning all booleans as u32

It's not clear to me what the precise rules where using booleans causes trouble are, but I've kept the existing codegen for the cases where u8 can be used

~~*ducks*~~


khyperia:
As you have correctly surmised, this is a massive hack that is fundamentally incorrect, and so we can't accept this.

PR #683, opened 2021-06-22
Add option to treat warnings as errors

### LINKED ISSUE #682: Option to treat errors as warnings through the spirv-builder.
Since warnings can't be checked by clippy currently it would be nice to have a option to treat errors as warnings as a replacement. `#![deny(warnings)]` isn't a great solution since in our case we want to depend on a environment variable.

khyperia:
Fixes #682


VZout (approved):
Sweet!

PR #685, opened 2021-06-23
Use target `spirv-unknown-webgpu0` for wgpu runner

VZout:
Looks like spirv-tools optimizer crashes on it though.


khyperia:
Looks like spirv-opt just flatly doesn't support it: https://github.com/KhronosGroup/SPIRV-Tools/blob/0c21e50922e469fa8cd2223ba9eeace58b1e0265/source/opt/optimizer.cpp#L66

In any case, the correct thing to do here is keep using vulkan1.1, as we're not running on the actual webgpu backend of wgpu, we're running on the vulkan backend of wgpu.

Issue #69, opened 2020-10-19
Fix pointers in aggs needing SROA


khyperia:
this makes me want to cry :crying_cat_face:

```rust
// build-pass

use spirv_std as _;

use core::panic::PanicInfo;
use glam::Vec4;

struct S<'a> {
    x: &'a Vec4,
    y: u32,
}

#[spirv(fragment)]
pub fn main(input: Vec4, mut output: &mut Vec4) {
    let s = S { x: &input, y: 5 };
    *output = *s.x;
}
```

### Error

```
error: invalid id:0:0 - OpLoad Pointer <id> '26[%26]' is not a logical pointer.
  %27 = OpLoad %v4float %26

  |
  = note: spirv-val failed
  = note: module `$TEST_BUILD_DIR/hello_world.stage-id.spv`
```

PR #690, opened 2021-07-18
Added a transformation that gets rid of temporary composites.

ElectronicRU:
Those temporary composites result from inlining of multi-argument
closures. Not only are they rather useless, they're also sometimes
invalid, when an argument to said closure is e.g. a pointer.

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
+++ b/crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
@@ -1,8 +1,8 @@
-//! Simplify OpCompositeExtract pointing to OpCompositeConstructs / OpCompositeInserts.
+//! Simplify `OpCompositeExtract` pointing to `OpCompositeConstruct`s / `OpCompositeInsert`s.
 //! Such constructions arise after inlining, when using multi-argument closures
-//! (and other Fn* trait implementations). These composites can frequently be invalid,
-//! containing pointers, OpFunctionArguments, etc. After simplification, components
-//! will become valid targets for OpLoad/OpStore.
+//! (and other `Fn*` trait implementations). These composites can frequently be invalid,
+//! containing pointers, `OpFunctionArgument`s, etc. After simplification, components
+//! will become valid targets for `OpLoad`/`OpStore`.
 use super::apply_rewrite_rules;
 use rspirv::dr::{Instruction, Module, Operand};
 use rspirv::spirv::Op;
@@ -21,7 +21,7 @@ pub fn destructure_composites(module: &mut Module) {
                 _ => None,
             })
             .collect();
-        let mut unused: FxHashSet<_> = reference.keys().map(|x| *x).collect();
+        let mut unused: FxHashSet<_> = reference.keys().copied().collect();
         for inst in function.all_inst_iter_mut() {
             if inst.class.opcode == Op::CompositeExtract && inst.operands.len() == 2 {
                 let mut composite = inst.operands[0].unwrap_id_ref();
@@ -61,7 +61,7 @@ pub fn destructure_composites(module: &mut Module) {
             if inst.result_id.is_none() || !reference.contains_key(&inst.result_id.unwrap()) {
                 for op in inst.operands.iter() {
                     if let Operand::IdRef(id_ref) = op {
-                        unused.remove(&id_ref);
+                        unused.remove(id_ref);
                     }
                 }
             }
@@ -75,7 +75,7 @@ pub fn destructure_composites(module: &mut Module) {
                 if !unused.contains(id) {
                     for op in inst.operands.iter() {
                         if let Operand::IdRef(id_ref) = op {
-                            changed |= unused.remove(&id_ref);
+                            changed |= unused.remove(id_ref);
                         }
                     }
                 }
--- a/crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
+++ b/crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
@@ -1,0 +1,86 @@
+//! Simplify OpCompositeExtract pointing to OpCompositeConstructs / OpCompositeInserts.
+//! Such constructions arise after inlining, when using multi-argument closures
+//! (and other Fn* trait implementations). These composites can frequently be invalid,
+//! containing pointers, OpFunctionArguments, etc. After simplification, components
+//! will become valid targets for OpLoad/OpStore.
+use super::apply_rewrite_rules;
+use rspirv::dr::{Module, Instruction, Operand};
+use rspirv::spirv::Op;
+use rustc_data_structures::fx::{FxHashMap, FxHashSet};
+
+pub fn destructure_composites(module: &mut Module) {
+    for function in module.functions.iter_mut() {
+        let mut rewrite_rules = FxHashMap::default();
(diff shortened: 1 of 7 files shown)


crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
```diff
@@ -0,0 +1,96 @@
...
+                        }
+                    } else {
+                        break None;
+                    }
+                };
+
+                if let Some(origin_id) = origin {
+                    rewrite_rules.insert(
+                        inst.result_id.unwrap(),
+                        rewrite_rules.get(&origin_id).map_or(origin_id, |id| *id),
```
khyperia, line 53: Hmm. I haven't convinced myself that this is correct, but my head is mush right now and I could easily be missing something.

I believe this is to address "transitive" rewrites - something like the following code pattern:

```rust
let a = A { val: x }; // CompositeConstruct
let y = a.val; // CompositeExtract
let b = B { val: y }; // CompositeConstruct
let z = b.val; // CompositeExtract
```

In this code, the `reference` map would contain `a[0] = x` and `b[0] = y`. When processing the first CompositeExtract, `rewrite_rules` would contain the rewrite `y -> x`. When processing the second CompositeExtract, it would first seem to insert `z -> y`, but the statement this comment is on would find that `y` is already in the map, and actually insert the rewrite rule `z -> x`, as it should.

But, imagine the statements are processed in the opposite order - which could happen if it's in a loop, or some other wacky construction of the CFG. In that case, it would first insert the `z -> y` rewrite, and then insert the `y -> x` rewrite. That would mean that `y` would get rewritten to `x`, as intended, but `z` would get rewritten to the now floating-in-the-void `y` variable instead of `x`.

I think the easiest way around this would be to change this line to a plain `rewrite_rules.insert(inst.result_id.unwrap(), origin_id)` statement, and then after the information gathering phase is over, do a, uh, "transitive closure computation" (idk words), something like this (pseudocode):

```
for entry in rewrite_rules:
    while rewrite_rules.contains(entry.value):
        entry.value = rewrite_rules[entry.value]
```

that way the information gathering phase is kept simple, and there's no weird edge cases around which order rules are inserted in, but the transitive closure still happens.

crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
```diff
@@ -0,0 +1,96 @@
...
+            }
+
+            // If the instruction wasn't replaced, modify the unused set
+            if inst.result_id.is_none() || !reference.contains_key(&inst.result_id.unwrap()) {
+                for op in inst.operands.iter() {
+                    if let Operand::IdRef(id_ref) = op {
+                        unused.remove(id_ref);
+                    }
+                }
+            }
```
khyperia, line 67: It's difficult to grok my head around the `unused` logic here (too many double negatives and whatnot, haha) - tbh what might be really nice is to just leave in all the unused stuff as dead code, then let https://github.com/EmbarkStudios/rust-gpu/pull/691 clean it up. That means that this (already kinda complicated and difficult to understand) pass can become simpler, since it doesn't have the responsibility of cleaning up after itself. That's what I did in, for example, mem2reg - it generates too many phis (some of the phis are dead code), but it just leaves that to later optimization passes, since the logic to clean those up already exists in those passes, and there's no need to overcomplicate the mem2reg pass with too many responsibilities.

crates/rustc_codegen_spirv/src/linker/mod.rs
```diff
@@ -231,6 +233,11 @@ pub fn link(sess: &Session, mut inputs: Vec<Module>, opts: &Options) -> Result<L
         }
     }
 
+    if opts.destructure {
+        let _timer = sess.timer("link_destructure");
+        destructure_composites::destructure_composites(&mut output);
+    }
```
khyperia, line 239: Really silly nit, it might be nice to move this to run right after `mem2reg::mem2reg` (within that for loop) and take a `&mut Function` instead, since it doesn't actually need the entire module to work on (it processes functions individually), would help cache locality and performance to not scan over the module twice.

khyperia (commented):
Goodness this is wonderful, I've been wanting to do this since forever (it's https://github.com/EmbarkStudios/rust-gpu/issues/69), but just never got around to doing it.

Unfortunately yeah, this is only one "layer" deep of composites, and so it's not a full solution (structs within structs still need to be dealt with), but it's most definitely better than nothing and I'm more than happy to accept this. AFAIK spirv-opt's version of this pass just runs a one-layer version repeatedly until all layers are peeled away, LLVM's version might too but not 100% sure.

A few small comments, but as a whole this looks really great!

crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
ElectronicRU: I'm pretty sure a loop would go through some OpPhis here?.. But I agree it's a sus place.
I've tried transitive closure computation - it's quite slow by itself (and also requires cloning the keys since you cannot really borrow the map both ways).
I'll try doing BOTH the eager replacement and closure computation, and profile that.

crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
ElectronicRU: True. This was written before #691, and it's required for correctness to prune those here. Launching a full DCE unconditionally each time we do this opt is way more expensive, meanwhile here we have a lot of info needed for this limited DCE. I'm not convinced.

crates/rustc_codegen_spirv/src/linker/mod.rs
ElectronicRU: Yes, but I'd like to preserve each feature flag being its own clause in the function - I think double-scanning the module is not the hugest cost, in my eyes, the cleanliness wins.

crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
khyperia: I disagree - we're _already_ "launching a full DCE unconditionally each time we do this opt". We run this opt once, and we do DCE afterwards. It's not more expensive.

ElectronicRU: Fair enough. Blocked by #691 for now, then.

crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
ElectronicRU: Resolved - included the closure calculation as verbatim as one might.

ElectronicRU:
Should be ready now!

crates/rustc_codegen_spirv/src/linker/mod.rs
```diff
@@ -289,6 +295,11 @@ pub fn link(sess: &Session, mut inputs: Vec<Module>, opts: &Options) -> Result<L
             dce::dce(output);
         }
 
+        {
+            let _timer = sess.timer("link_remove_duplicate_lines_2");
+            duplicates::remove_duplicate_lines(output);
+        }
```
khyperia, line 301: Rather than running this again, please move the original one here and only run it once. The original is where it is now because none of the later passes could modify function contents, but now with #691, DCE can.

khyperia (approved):
LGTM pending one nit!

PR #691, opened 2021-07-22
Aggressively prune no-side-effect instructions during DCE.

ElectronicRU:
Since we're walking all the instructions anyway, it's practically
zero-cost.

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/dce.rs
+++ b/crates/rustc_codegen_spirv/src/linker/dce.rs
@@ -7,8 +7,8 @@
 //! *references* a rooted thing is also rooted, not the other way around - but that's the basic
 //! concept.
 
-use rspirv::dr::{Function, Instruction, Module};
-use rspirv::spirv::{Op, Word};
+use rspirv::dr::{Function, Instruction, Module, Operand};
+use rspirv::spirv::{Op, StorageClass, Word};
 use rustc_data_structures::fx::FxHashSet;
 
 pub fn dce(module: &mut Module) {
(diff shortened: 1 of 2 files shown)


crates/rustc_codegen_spirv/src/linker/dce.rs
```diff
@@ -37,7 +37,13 @@ fn spread_roots(module: &Module, rooted: &mut FxHashSet<Word>) -> bool {
     for func in &module.functions {
         if rooted.contains(&func.def_id().unwrap()) {
             for inst in func.all_inst_iter() {
-                any |= root(inst, rooted);
+                if !instruction_is_pure(inst) {
+                    any |= root(inst, rooted);
+                } else if let Some(id) = inst.result_id {
+                    if rooted.contains(&id) {
+                        any |= root(inst, rooted);
```
khyperia, line 44: hrm, this might be dreadfully inefficient, some profiling might be nice here! (how many times `spread_roots` is called before/after this change). It's probably way more efficient to iterate over instructions backwards instead of forwards.

khyperia (commented):
Neat! I think I'd normally be a little meh about this, since I'm trying (and partially failing) to keep rust-gpu's passes strictly for correctness rather than optimization, and leave the optimization passes to spirv-opt, but if you opt for my comment about `unused` in #690 then this _is_ actually needed for correctness (to nuke the invalid composites).

I think the android CI failure should be fixed by https://github.com/EmbarkStudios/rust-gpu/pull/701, could you rebase?

crates/rustc_codegen_spirv/src/linker/dce.rs
khyperia: Yep, just checked, when running the `multibuilder` example on `main`, `dce` is called three times, and in each of those, `spread_roots` is called `[3, 4, 1]` times. With this PR, it's `[8, 8, 5]`, a significant cost (definitely not zero-cost, especially with the last one being a no-op). But, reversing the iteration order of the function's instructions, it drops back down to `[3, 4, 1]`.

ElectronicRU: Hm. So do the reversing of iteration and write a comment saying why it is so?

khyperia (approved):

PR #692, opened 2021-07-23
Added NonWritable + NonReadable Decorations. Fixes  #689

### LINKED ISSUE #689: Support for NonReadable, NonWritable
I couldn't find any existing issues tracking adding support for specifying these decorations (or other decorations). Would it make sense for these to be exposed as `spirv` attributes, or as another `const` parameter of `Image`, for example? With the latter approach, correct usage of variables with these decorations can be enforced, but it might be complicated slightly by the decorations in SPIR-V being for the variable, not the type itself. I'm also not sure how that approach would work with buffers.

It also does not look to be possible currently to work around this with inline assembly, at least for input arguments, as the global `OpVariable` identifiers do not appear to be visible from `asm!{}`.

charles-r-earp:
Fixes  #689.

This PR allows adding NonWritable and NonReadable decorations to StorageBuffers. This can be added to other types, Images for example. This is an opt in via attributes `#[spirv(non_writable)]` and `#[spirv(non_readable)]`, and does not do any validation. Probably `non_writable` can be inferred from the function declaration but that can be addressed later. 

It looks like this:

    #![cfg_attr(
        target_arch = "spirv",
        no_std,
        feature(register_attr),
        register_attr(spirv)
    )]
    #![deny(warnings)]

    use spirv_std::glam::UVec3;

    #[repr(C)]
    pub struct PushConsts {
        n: u32,
    }

    #[allow(unused)]
    #[spirv(compute(threads(64)))]
    pub fn add(
        #[spirv(global_invocation_id)] global_id: UVec3,
        #[spirv(storage_buffer, descriptor_set = 0, binding = 0, non_writable)] a: &[u32],
        #[spirv(storage_buffer, descriptor_set = 0, binding = 1, non_writable)] b: &[u32],
        #[spirv(storage_buffer, descriptor_set = 0, binding = 2, non_readable)] y: &mut [u32],
        #[spirv(push_constant)] push_consts: &PushConsts,
    ) {
        let gid = global_id.x as usize;
        if global_id.x < push_consts.n {
            y[gid] = a[gid] + b[gid];
        }
    }

As far as validation, this could be done in the validation pass or through the type system, by adding a trait bound to the argument. But the former is probably the most robust way, but it has to do a recursive check against all the access chains as well. It would be nice if immutable arguments were declared non_writable automatically for known types at the very least. For example, `&[T] where T: Copy` should be able to be marked non_writable.

### DIFF
--- a/crates/rustc_codegen_spirv/src/attr.rs
+++ b/crates/rustc_codegen_spirv/src/attr.rs
@@ -85,6 +85,8 @@ pub enum SpirvAttribute {
     Binding(u32),
     Flat,
     Invariant,
+    NonWritable,
+    NonReadable,
     InputAttachmentIndex(u32),
 
     // `fn`/closure attributes:
@@ -120,6 +122,8 @@ pub struct AggregatedSpirvAttributes {
     pub binding: Option<Spanned<u32>>,
     pub flat: Option<Spanned<()>>,
     pub invariant: Option<Spanned<()>>,
+    pub non_writable: Option<Spanned<()>>,
+    pub non_readable: Option<Spanned<()>>,
     pub input_attachment_index: Option<Spanned<u32>>,
 
     // `fn`/closure attributes:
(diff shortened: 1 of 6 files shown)


khyperia:
Unfortunately, as stated in the readme, we're not able to accept major feature contributions right now :( - hopefully we can reopen this later when we're a little more stable!

PR #693, opened 2021-07-26
Fix example-runner-ash

hatoo:
`example-runner-ash` fails to run because it fails to get `sky_shader`.

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', examples\runners\ash\src\main.rs:706:74
error: process didn't exit successfully: `target\debug\example-runner-ash.exe` (exit code: 101)
```

`compile_shaders()` had obtained shader name from `path.file_stem()` but path is
```
[examples\runners\ash\src\main.rs:212] &spv_paths = [
    "C:\\Users\\hato2\\Desktop\\rust-gpu\\target\\spirv-unknown-vulkan1.1\\release\\deps\\sky_shader.spv.dir\\module",
]
```
in my environment (Windows11).

I've fixed `compile_shaders()` to return corrent shader name from path.
Sorry if this fix is incompatioble for other environments.

### DIFF
--- a/examples/runners/ash/src/main.rs
+++ b/examples/runners/ash/src/main.rs
@@ -211,8 +211,20 @@ pub fn compile_shaders() -> Vec<SpvFile> {
         ];
     let mut spv_files = Vec::<SpvFile>::with_capacity(spv_paths.len());
     for path in spv_paths.iter() {
+        let name = path
+            .parent()
+            .unwrap()
+            .file_name()
+            .unwrap()
+            .to_str()
+            .unwrap()
+            .split('.')
+            .next()
+            .unwrap()
+            .to_owned();
+
         spv_files.push(SpvFile {
-            name: path.file_stem().unwrap().to_str().unwrap().to_owned(),
+            name,
             data: read_spv(&mut File::open(path).unwrap()).unwrap(),
         });
     }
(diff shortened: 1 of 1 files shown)


khyperia:
... oh wow, yeah, this code is really incorrect and should not be grabbing the name from the filename - and like, it's excessively weirdly complicated. I think I'd prefer to change the whole function to just something like this:

```rust
fn compile_shaders() -> Vec<SpvFile> {
    let sky_shader_path =
        SpirvBuilder::new("examples/shaders/sky-shader", "spirv-unknown-vulkan1.1")
            .print_metadata(MetadataPrintout::None)
            .build()
            .unwrap()
            .module
            .unwrap_single()
            .to_path_buf();
    let sky_shader = SpvFile {
        name: "sky_shader".to_string(),
        data: read_spv(&mut File::open(sky_shader_path).unwrap()).unwrap(),
    };
    vec![sky_shader]
}
```

and to be honest, I don't know why we're even doing this whole `Vec<SpvFile>` thing instead of just returning a module... the fact that it's been broken for months is a good argument for what we've been considering for a while, and just nuking the whole thing, haha.

hatoo:
Yeah, I agree that the code is complicated.
I copypasted your code. And confirmed to run.

khyperia (approved):

PR #694, opened 2021-07-26
Allow bound and bindless resources at the same time.

VZout:
As the title says, allow bound and bindless resources at the same time.
You can still tell rust-gpu to exclusively use bindless by writing `SpirvBuilder::new(...).bindless(BindingSupport::Exclusive)`

Tagging @Jasper-Bekkers


Jasper-Bekkers:
Could you elaborate a bit on why this would be needed, and what the benefits would be? It seems counter-intuitive to me to mix and match them from a design point of view?

Also if you decide to go forward with this, it's probably wise to do some significant error checking around overlapping descriptor sets.

Notice also that there are some implications for this on Android; lots of devices there only support 4 descriptor sets at a time so this will conflict (since bindless just takes all the descriptor sets for itself).

VZout:
> Could you elaborate a bit on why this would be needed, and what the benefits would be? It seems counter-intuitive to me to mix and match them from a design point of view?
> 
> Also if you decide to go forward with this, it's probably wise to do some significant error checking around overlapping descriptor sets.
> 
> Notice also that there are some implications for this on Android; lots of devices there only support 4 descriptor sets at a time so this will conflict (since bindless just takes all the descriptor sets for itself).

The only reason is to make our internal porting easier. If this gets merged it be removed again in the future just like the other bindless check.

Jasper-Bekkers:
> > Could you elaborate a bit on why this would be needed, and what the benefits would be? It seems counter-intuitive to me to mix and match them from a design point of view?
> > Also if you decide to go forward with this, it's probably wise to do some significant error checking around overlapping descriptor sets.
> > Notice also that there are some implications for this on Android; lots of devices there only support 4 descriptor sets at a time so this will conflict (since bindless just takes all the descriptor sets for itself).
> 
> The only reason is to make our internal porting easier. If this gets merged it should be removed again in the future just like the other bindless check.

Doesn't seem to make a lot of sense to make that the default then - but if it helps with your development go ahead :-)

eddyb (approved):
Rubber-stamping this as I don't want to block anything internal, but I won't merge myself as I haven't gotten around to looking into the bindless infrastructure.

hrydgard:
Seems like a strange CI problem here?

crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
```diff
@@ -126,6 +127,14 @@ impl<'tcx> CodegenCx<'tcx> {
             }
         }
 
+        let mut exclusive_bindless = false;
+        for &feature in &tcx.sess.target_features {
+            if feature == sym.exclusive_bindless {
+                exclusive_bindless = true;
+                break;
+            }
+        }
```
khyperia, line 136: Probably good to merge these for loops. TBH I might prefer an enum here, and do error checking around things like an invalid combination.

khyperia (approved):
LGTM, #701 should have fixed the CI, could you rebase?

PR #698, opened 2021-07-29
Use DefaultMetadataLoader for loading dylib metadata

bjorn3:
If you switch archives from tar to ar archives, it would be possible to use `DefaultMetadataLoader` for everything.

### DIFF
--- a/crates/rustc_codegen_spirv/src/lib.rs
+++ b/crates/rustc_codegen_spirv/src/lib.rs
@@ -269,15 +269,7 @@ impl MetadataLoader for SpirvMetadataLoader {
     }
 
     fn get_dylib_metadata(&self, target: &Target, path: &Path) -> Result<MetadataRef, String> {
-        // HACK(eddyb) this is needed to allow metadata loading for proc macros
-        // (compiled as host dylibs); perhaps it'd be better to use the `object`
-        // crate, like `rustc_codegen_cranelift` does.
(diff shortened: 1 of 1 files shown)


eddyb (approved):

eddyb:
Looks like the Android CI is just broken nowadays, but we should definitely merge this PR whenever CI is unbroken.

bjorn3:
https://github.com/bevyengine/bevy/pull/2533 was the "fix" for bevy. Basically uninstall the broken android build tools version 31 first, so `cargo-apk` can install version 30 I think.

khyperia (approved):
> If you switch archives from tar to ar archives, it would be possible to use `DefaultMetadataLoader` for everything.

What do you mean by "everything"? I guess like, yeah, sure, sounds great, tbh I didn't realize there was a difference between tar and ar archives, haha.

Anyway, #701 should have fixed the CI, could you rebase?

bjorn3:
> What do you mean by "everything"?

For rlibs too and not just for dylibs.

> I guess like, yeah, sure, sounds great, tbh I didn't realize there was a difference between tar and ar archives, haha.

Ar archives are much more limited. For example they don't support directories, symlinks, etc. On the other hand both the gnu and bsd flavor support a symbol table that maps from symbol name to the object file that contains said symbol. This can improve linker speed and is required by most linkers.

> Anyway, #701 should have fixed the CI, could you rebase?

Done

PR #704, opened 2021-08-04
Add sample_bias and gather functions

### LINKED ISSUE #703: Image seems to be missing `sample_bias()`
Yet another sampling function! Simply regular sampling, but with a mip lod bias applied.

Need something corresponding to https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-to-samplebias , can't find anything relevant in https://embarkstudios.github.io/rust-gpu/api/spirv_std/image/struct.Image.html .

### LINKED ISSUE #706: Image texture_gather (yet another sampling mode)
Ran into another image sampling function :)

HLSL: GatherRed/Green/Blue/Alpha
https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/sm5-object-texture2d-gatherred

GLSL: textureGather(texel_coordinate, component)
https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/textureGather.xhtml

SPIR-V: OpImageGather (very similar to OpImageFetch).

This one gathers a single component/channel of the four texels that would have been used for a bilinear filter lookup at the specified coordinate, and returns them in a Vec4. Useful when implementing various custom filters on single channel input.

I can easily work around the lack of this for now by doing four fetches, so no rush.

khyperia:
Fixes #703

Fixes #706

### DIFF
--- a/crates/spirv-std/src/image.rs
+++ b/crates/spirv-std/src/image.rs
@@ -162,6 +162,39 @@ impl<
         }
     }
 
+    /// Sample texels at `coord` from the image using `sampler`, after adding the input bias to the
+    /// implicit level of detail.
+    #[crate::macros::gpu_only]
+    pub fn sample_bias<F, V>(
+        &self,
+        sampler: Sampler,
+        coord: impl ImageCoordinate<F, DIM, ARRAYED>,
+        bias: f32,
+    ) -> V
+    where
+        F: Float,
+        V: Vector<SampledType, 4>,
+    {
+        unsafe {
+            let mut result = Default::default();
+            asm!(
+                "%typeSampledImage = OpTypeSampledImage typeof*{1}",
+                "%image = OpLoad typeof*{1} {1}",
+                "%sampler = OpLoad typeof*{2} {2}",
+                "%coord = OpLoad typeof*{3} {3}",
+                "%sampledImage = OpSampledImage %typeSampledImage %image %sampler",
+                "%result = OpImageSampleImplicitLod typeof*{0} %sampledImage %coord Bias {4}",
+                "OpStore {0} %result",
+                in(reg) &mut result,
+                in(reg) self,
+                in(reg) &sampler,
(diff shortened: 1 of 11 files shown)


crates/spirv-std/src/image.rs
```diff
@@ -136,6 +136,42 @@ impl<
         ACCESS_QUALIFIER,
     >
 {
+    // Note: #[inline] is needed because in vulkan, the component must be a constant expression.
```
eddyb, line 139: You could plausibly add a `const COMPONENT: u32` generic parameter, `#[rustc_legacy_const_generics(3)]`, and remove `component: u32` from the normal parameter list. AFAICT, this would result in the same call syntax working, but it gets automatically converted to the `const` generic.

(this is what `core::arch` uses for constant arguments - not sure how appropriate it is here, but as it stands, the argument isn't even required to be constant)

crates/spirv-std/src/image.rs
```diff
@@ -974,3 +1034,527 @@ impl<
...
+    for Image<
+        SampledType,
+        { Dimensionality::Cube },
+        DEPTH,
+        ARRAYED,
+        { Multisampled::False },
+        SAMPLED,
+        FORMAT,
+        ACCESS_QUALIFIER,
+    >
```
eddyb, line 1558: I *really* wish we could use `_` as a shorthand for generic impls, i.e.:
```rust
impl HasQuerySizeLod for Image<_, { Dimensionality::Cube }, _, _, { Multisampled::False }, _, _, _> {}
```

eddyb (approved):

crates/spirv-std/src/image.rs
khyperia: Yeah, it's a bit of a bizarre case - when compiling for Vulkan, it's required to be constant, but when not, it's free to be a variable. I feel like there's just no good solution here.

PR #705, opened 2021-08-04
Fix `cargo test`

DJMcNab:
Also test a full cargo test in CI

### DIFF
--- a/crates/spirv-std/macros/src/lib.rs
+++ b/crates/spirv-std/macros/src/lib.rs
@@ -85,7 +85,7 @@ use quote::ToTokens;
 ///
 /// The grammar for the macro is as follows:
 ///
-/// ```rust,no_compile
+/// ```rust,ignore
 /// Image!(
 ///     <dimensionality>,
 ///     <type=...|format=...>,
@@ -99,7 +99,7 @@ use quote::ToTokens;
 /// `=true` can be omitted as shorthand - e.g. `sampled` is short for `sampled=true`.
 ///
 /// A basic example looks like this:
-/// ```rust,no_compile
+/// ```rust,ignore
 /// #[spirv(vertex)]
 /// fn main(#[spirv(descriptor_set = 0, binding = 0)] image: &Image!(2D, type=f32, sampled)) {}
 /// ```


crates/spirv-std/macros/Cargo.toml
```diff
@@ -15,4 +15,7 @@ spirv-types = { path = "../shared", version = "0.4.0-alpha.10" }
 heck = "0.3.2"
 proc-macro2 = "1.0.24"
 quote = "1.0.8"
-syn = { version = "1.0.58", features=["full"] }
+syn = { version = "1.0.58", features = ["full"] }
+
+[dev-dependencies]
+spirv-std = { path = ".." }
```
khyperia, line 21: I don't think this is worth the extra build time and complexity to validate a single doctest. (I didn't even know one was allowed to have a circular dev-dependency, seems like a thing to avoid.)

khyperia (changes requested):
Unfortunately, this regresses CI times by 8 minutes, which is unacceptable for the benefit of "running" two doctests which aren't even doctests (we've fought tooth and nail for every minute of CI saved). I would prefer to just disable doctests for spirv-std-macros to make sure `cargo test` doesn't regress.

DJMcNab:
Unfortunately, making sure that cargo test doesn't regress is the root cause of the extra compile time, because it's now trying to test the example runners and the shaders (for example, `cargo test` would regress again if we added a failing test for the collatz example).

That being said, I guess targeted only testing the specific crates which do have tests does make sense - it's just rather unfortunate that it leaves the door open for `cargo test` to silently regress again. I'll aim to move to that version at lunchtime

An alternative could be adding a single extra job which runs all of the tests on one platform - that one wouldn't have to use release mode so it should be relatively faster.

khyperia:
> That being said, I guess targeted only testing the specific crates which do have tests does make sense - it's just rather unfortunate that it leaves the door open for `cargo test` to silently regress again. I'll aim to move to that version at lunchtime

That's why I said I'd prefer disabling tests for those crates that have no tests.

crates/spirv-std/src/lib.rs
```diff
@@ -109,7 +109,6 @@ pub(crate) mod sealed;
 pub mod vector;
 
 pub use self::sampler::Sampler;
-pub use crate::macros::Image;
```
khyperia, line 112: This should still be exported.

crates/spirv-std/src/image.rs
```diff
@@ -6,6 +6,62 @@
...
+///    Accepted values: `true` or `false`. Default: `false`.
+/// - `depth` — Whether it is known that the image is a depth image.
+///    Accepted values: `true` or `false`. Default: `unknown`.
+///
+/// [`ImageFormat`]: spirv_types::image_params::ImageFormat
+///
+/// Keep in mind that `sampled` here is a different concept than the `SampledImage` type:
+/// `sampled=true` means that this image requires a sampler to be able to access, while the
+/// `SampledImage` type bundles that sampler together with the image into a single type (e.g.
+/// `sampler2D` in GLSL, vs. `texture2D`).
```
khyperia, line 64: The text for the original macro definition is bizarrely appended to this text. That should be figured out why before merging.

khyperia (changes requested):
`crates/spirv-std/src/image/params.rs` has a broken doc comment now.

crates/spirv-std/src/image.rs
DJMcNab: Ah, interesting

I only added that doc comment as a last minute thing before I had to go again

I think the best way is just to turn it into a normal comment.

crates/spirv-std/src/lib.rs
DJMcNab: Makes sense, can do

Imo, we should choose one canonical place and stick with it, whether that's inside or outside of the image module of spirv_std

crates/spirv-std/src/image.rs
khyperia: A normal comment would be problematic, as it wouldn't show up in the macro crate's docs.

crates/spirv-std/src/lib.rs
DJMcNab: Documentation of re-exports is not copied into re-exports of the re-export (i.e. no matter how we use this without duplicating the doc comment from `spirv_std::image::Image!`, we'll only get the docs from the macro in spirv_std_macros).

One alternative would be to make a macro `Image` in `spirv_std` which just calls `spirv_std::macros::Image`, which should work. But it would be nicer just not to have this re-re-export at the root.

crates/spirv-std/src/image.rs
DJMcNab: How much do we actually care about the macro crate docs? 

That is, would we be better off serving all of the macros as `spirv_std::whatever`, e.g. by using `#[doc(inline)]` on re-exports?

DJMcNab:
And it looks like we should normalise the positions in `compiletest` to not depend on the positions in `image.rs` crate. I'm going to just re-update the UI test output for now.

crates/spirv-std/src/image.rs
```diff
@@ -6,6 +6,66 @@
...
+/// A macro for creating SPIR-V `OpTypeImage` types. Always produces a
+/// [`spirv_std::image::Image<...>`](struct@Image) type.
+///
+/// The grammar for the macro is as follows:
+///
+/// ```rust
+/// # // This doctest can't use the normal macro, use an alternative version
+/// # macro_rules! Image {
+/// #   ($($tt: tt)*)=> {}
+/// # }
```
khyperia, line 18: This makes no sense to test. We should just ignore this.

crates/spirv-std/src/image.rs
```diff
@@ -6,6 +6,66 @@
...
+///
+/// [`ImageFormat`]: spirv_types::image_params::ImageFormat
+///
+/// Keep in mind that `sampled` here is a different concept than the `SampledImage` type:
+/// `sampled=true` means that this image requires a sampler to be able to access, while the
+/// `SampledImage` type bundles that sampler together with the image into a single type (e.g.
+/// `sampler2D` in GLSL, vs. `texture2D`).
+///
+/// Note: The documentation below results from this being a re-export from the [`spirv-std-macros`](`crate::macros`) crate.
+/// It can be safely ignored
```
khyperia, line 67: This is even more awkward than the original appending of text, IMO

crates/spirv-std/src/lib.rs
```diff
@@ -109,7 +109,14 @@ pub(crate) mod sealed;
...
+/// A macro for creating SPIR-V `OpTypeImage` types and the type itself
+/// See also the re-export of this macro in the [`image`] module, which has more thorough
+/// documentation.
+///
+/// Note: The documentation below results from this being a re-export from the [`spirv-std-macros`](`crate::macros`) crate.
+/// It can be safely ignored
+///
+pub use crate::image::Image;
```
khyperia, line 119: Why do a re-re-export instead of a re-export?

tests/ui/spirv-attr/bad-infer-storage-class.stderr
```diff
@@ -19,7 +19,7 @@ warning: redundant storage class specifier, storage class is inferred from type
 9 |     #[spirv(uniform_constant)] warning: &Image!(2D, type=f32),
   |             ^^^^^^^^^^^^^^^^
 
-error: entry parameter type must be by-reference: `&spirv_std::image::Image<f32, spirv_std::image::Dimensionality::TwoD, spirv_std::image::ImageDepth::Unknown, spirv_std::image::Arrayed::False, spirv_std::image::Multisampled::False, spirv_std::image::Sampled::Unknown, spirv_std::image::ImageFormat::Unknown, core::option::Option::<spirv_std::image::AccessQualifier>::None>`
+error: entry parameter type must be by-reference: `&spirv_std::Image<f32, spirv_std::image::Dimensionality::TwoD, spirv_std::image::ImageDepth::Unknown, spirv_std::image::Arrayed::False, spirv_std::image::Multisampled::False, spirv_std::image::Sampled::Unknown, spirv_std::image::ImageFormat::Unknown, core::option::Option::<spirv_std::image::AccessQualifier>::None>`
```
khyperia, line 22: Why was this changed? It seems like an indication of something gone wrong.

khyperia (changes requested):

crates/spirv-std/src/lib.rs
DJMcNab: It was a re-reexport in an attempt to get the docs properly copied over

Obviously that doesn't work, so it can just be a normal export again.

crates/spirv-std/src/image.rs
DJMcNab: I wanted to avoid the giant *this test is ignored* warning box

But I'm happy to change it back again and just ignore it.

tests/ui/spirv-attr/bad-infer-storage-class.stderr
DJMcNab: This is just because the Image struct is also re-exported at the root, because I was using a re-reexport to try to get the docs.

Since that didn't work anyway, this change should slide back out again when it changes to just a normal re-export.

khyperia:
Honestly at this point it seems like the best plan here is to turn this change into just adding `,ignore` to the two tests, that's it.

DJMcNab:
That's reasonable. I do think this takes us to a better state, but equally it does end up being more complicated. I'll make that change when I get the chance

khyperia (approved):

PR #709, opened 2021-08-09
Implement some float packing/unpacking instructions

### LINKED ISSUE #707: Expose floating point format conversion operations (16 <-> 32 bits)?
HLSL and GLSL both have operations that convert between 16 and 32-bit floating point formats.

In GLSL they look like this, packing/unpacking two 16-bit floats to/from an uint:

```
vec2 unpackHalf2x16(uint v);
uint packHalf2x16(vec2 v);
```

HLSL has scalar (non-pack/unpack) operations only:

```
uint f32tof16(in float value);
float f16tof32(in uint value);
```

I have not looked in detail yet about what actual SPIR-V the GLSL operations become. Possibly there's some capability required to get these natively?

khyperia:
Fixes #707

Wasn't too hot on the GLSL/SPIR-V name of packHalf2x16, it's *really* confusing to me what the heck that means, so tried to go more explicit with the name - at least to me, f32x2_to_f16x2 is much more clear (and in line with the scalar f32_to_f16 version)

Left off f64 conversion insts for now, ark doesn't need them any time soon and they're weird (I think they're `transmute (u32, u32) <-> f64`, not an actual bit-changing conversion?)

### DIFF
--- a/crates/spirv-std/src/float.rs
+++ b/crates/spirv-std/src/float.rs
@@ -16,7 +16,7 @@ unsafe impl Float for f64 {
 /// Converts two f32 values (floats) into two f16 values (halfs). The result is a u32, with the low
 /// 16 bits being the first f16, and the high 16 bits being the second f16.
 #[spirv_std_macros::gpu_only]
-pub fn f32x2_to_f16x2(vec: impl Vector<f32, 2>) -> u32 {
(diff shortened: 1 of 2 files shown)


hrydgard:
The new naming is better. Probably descriptive enough while still avoiding SSE-intrinsic-style acronym salad. The u32 types makes it clear enough that it's about packed values.

hrydgard (approved):

PR #710, opened 2021-08-09
Remove implicit bindless and kernel modes

khyperia:
As discussed in the meeting, remove bindless mode and kernel mode, as they're unfortunately not directions that we want to head in.

Note that "bindless mode" does not refer to actually supporting bindless itself, but rather a strongly opinionated system of features that enforced all rust-gpu programs use bindless, with no option of "normally" bound resources. Bindless is still very much a feature we want to support first-class, but we want to make more incremental changes towards making bindless easier to write, rather than dropping support for non-bindless systems. (We already support bindless in the regular, non-"bindless mode" model, and will continue to support it after this PR)


repi:
Should we clarify in the title and description that it is the _implicit bindless_ mode (or some other name for it?) this is referring to, standard bindless with explicit resources that you manage yourselves is still fully possible and something we use in Ark as well.

khyperia:
> Should we clarify in the title and description that it is the _implicit bindless_ mode (or some other name for it?) this is referring to, standard bindless with explicit resources that you manage yourselves is still fully possible and something we use in Ark as well.

Ah yeah, probably! Edited the description, but not sure how to concisely say that in the title (feel free to edit), I'm too deep in the technical weeds to think of something understandable to folks not in the pit of shenanigans :P

repi:
Thanks fair enough, think we can call the old mode that is removed here "implicit bindless" as it did implicitly create resources for bindless access.

That said - "bindless" is a stupid word as well :)

eddyb (approved):

Jasper-Bekkers:
This removal is really unfortunate and is kind of forcing us to have to maintain a fork of the project :-(

PR #715, opened 2021-08-10
Introduce `remove_unused_params` pass, to run just before zombie removal.

eddyb:
By removing (transitively) dead `fn` parameters, this PR also removes their matching *arguments*, which could be invalid (i.e. zombies).

The main usecase is `#[track_caller]`, which is becoming more prevalent in `core`.
The added test fails before this PR with:
```
error: constant arrays/structs cannot contain pointers to other constants
  --> $DIR/track_caller.rs:16:5
   |
16 |     track_caller_maybe_panic(x);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: Stack:
           track_caller::main
           Unnamed function ID %17
```

This allows updating past https://github.com/rust-lang/rust/pull/86664 (which gets transitively called from `mem::replace`).

Also, I tried using this to remove our hardcoded panic hacks, but it's not compatible with `-C debuginfo=2` (even if the `%x.dbg.spill` stack slots have no debuginfo attached, they're still created and `store`d to, so they get in the way).

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/param_weakening.rs
+++ b/crates/rustc_codegen_spirv/src/linker/param_weakening.rs
@@ -1,0 +1,114 @@
+//! Interprocedural optimizations that "weaken" function parameters, i.e. they
+//! replace parameter types with "simpler" ones, or outright remove parameters,
+//! based on how those parameters are used in the function and/or what arguments
+//! get passed from callers.
+//!
+use crate::linker::ipo::CallGraph;
+use indexmap::IndexMap;
+use rspirv::dr::{Builder, Module, Operand};
+use rspirv::spirv::{Op, Word};
+use rustc_data_structures::fx::FxHashMap;
+use rustc_index::bit_set::BitSet;
+use std::mem;
+
+pub fn remove_unused_params(module: Module) -> Module {
+    let call_graph = CallGraph::collect(&module);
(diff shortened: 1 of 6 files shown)


crates/rustc_codegen_spirv/src/linker/param_weakening.rs
```diff
@@ -0,0 +1,114 @@
...
+
+        // Regenerate the function type from remaining parameters, if necessary.
+        if unused_params.is_some() {
+            let return_type = func.def.as_mut().unwrap().result_type.unwrap();
+            let new_param_types: Vec<_> = func
+                .parameters
+                .iter()
+                .map(|inst| inst.result_type.unwrap())
+                .collect();
+            let new_func_type = builder.type_function(return_type, new_param_types);
```
khyperia, line 107: Hrmpf. `O(n^2)` linear search here for dedup isn't great, but I guess it's fine, usually not many functions.

khyperia (approved):

crates/rustc_codegen_spirv/src/linker/param_weakening.rs
hrydgard: "usually not many" - famous last words ;)

PR #716, opened 2021-08-10
rustup: update to nightly-2021-08-10.

eddyb:
This is finally possible because of:
* https://github.com/rust-lang/rust/pull/87827
* https://github.com/EmbarkStudios/rust-gpu/pull/712
* https://github.com/EmbarkStudios/rust-gpu/pull/715


khyperia (approved):
:tada:

eddyb:
If something broke with android while we weren't updating, I may need to bisect it, ugh.

PR #717, opened 2021-08-10
asm: add support for noreturn option

msiglreith:
`OpUnreachable` will be appended as terminator at the end of the asm block (done automatically by the rustc)

Fixes https://github.com/EmbarkStudios/rust-gpu/issues/570

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
+++ b/crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
@@ -70,8 +70,13 @@ impl<'a, 'tcx> AsmBuilderMethods<'tcx> for Builder<'a, 'tcx> {
         options: InlineAsmOptions,
         _line_spans: &[Span],
     ) {
-        if !options.is_empty() {
-            self.err(&format!("asm flags not supported: {:?}", options));
+        const SUPPORTED_OPTIONS: InlineAsmOptions = InlineAsmOptions::NORETURN;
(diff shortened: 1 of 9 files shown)


khyperia (changes requested):
With `options(noreturn)`, the user should not have to write `%unused = OpLabel`, which is quite a horrible workaround hack.

I believe the proper fix here is to add full understanding of basic blocks to the asm emitter, rather than shoving basic block headers/terminators into `blocks[i].instructions`.

Specifically, the emitter should keep track of whether the current context is within a basic block, keep track of what that block is (if there is a block), validate that no invalidly placed position instructions are present (e.g. OpAdd after a terminator), and assert that at the end of the block, if `noreturn` is not present, there is an active block, but if `noreturn` is present, there is not an active block.

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
```diff
@@ -328,10 +366,32 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
                 }
                 return;
             }
-            _ => {
+
+            op => {
                 self.emit()
                     .insert_into_block(dr::InsertPoint::End, inst)
                     .unwrap();
```
khyperia, line 373: Still would be nice to not shove everything into the same block, including terminators and labels, but whatever I guess!

khyperia (approved):

PR #718, opened 2021-08-10
asm: implicit label after return or abort terminator

### LINKED ISSUE #564: Correctly handle terminators in inline assembly
Currently the following assembly code causes a `ConsumerError(MismatchedTerminator)` ICE.

### Code

```
// build-pass

use spirv_std as _;

#[spirv(any_hit_khr)]
pub fn main() {
    unsafe {
        asm!(r#"OpExtension "SPV_KHR_ray_tracing""#);
        asm!("OpCapability RayTracingKHR", "OpTerminateRayKHR");
    }
}
```

### Error

```
thread 'rustc' panicked at 'called `Result::unwrap()` on an `Err` value: ConsumerError(MismatchedTerminator)', crates/rustc_codegen_spirv/src/link.rs:423:58

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/EmbarkStudios/rust-gpu/issues/new

note: rustc 1.52.0-nightly (61edfd591 2021-03-20) running on x86_64-apple-darwin

note: compiler flags: -Z codegen-backend=/Users/erin.power/src/rust-gpu2/target/release/deps/librustc_codegen_spirv.dylib -Z unstable-options -Z crate-attr=no_std -Z crate-attr=feature(register_attr,asm) -Z crate-attr=register_attr(spirv) -C prefer-dynamic -C overflow-checks=off -C debug-assertions=off -C debuginfo=2 -C embed-bitcode=no --crate-type dylib
```

msiglreith:
Implicitly add an `OpLabel` once a return or abort terminator is found in the asm line stream. This allows to remove the `%unused = OpLabel` workaround.

Internally, it creates an explicit label instruction and skips the begin_block/end_block mechanism of the rspirv builder due to how the cursor works. I hope handling this transparently is fine, asumming there will be no reference to this label.

Depends on #717
Fixes #564

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
+++ b/crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
@@ -70,8 +70,13 @@ impl<'a, 'tcx> AsmBuilderMethods<'tcx> for Builder<'a, 'tcx> {
         options: InlineAsmOptions,
         _line_spans: &[Span],
     ) {
-        if !options.is_empty() {
-            self.err(&format!("asm flags not supported: {:?}", options));
+        const SUPPORTED_OPTIONS: InlineAsmOptions = InlineAsmOptions::NORETURN;
+        let unsupported_options = options & !SUPPORTED_OPTIONS;
+        if !unsupported_options.is_empty() {
+            self.err(&format!(
+                "asm flags not supported: {:?}",
+                unsupported_options
+            ));
         }
         // vec of lines, and each line is vec of tokens
         let mut tokens = vec![vec![]];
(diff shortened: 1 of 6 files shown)


khyperia (changes requested):
This is incorrect, as the user should be able to write multiple blocks in an `asm!` with normal syntax - with this, they would somehow have to know to not write `OpLabel`.

Also, please merge this PR into #717, partial fixes from partial PRs leaving glaring issues are very difficult to coordinate and review. (I didn't see this PR when I reviewed 717)

msiglreith:
Thanks, will look into it again - closing in favor of a single PR #717

PR #721, opened 2021-08-13
link: switch from `tar` to `ar` for `.rlib`s.

### LINKED ISSUE #711: Migrate from tar to ar
As mentioned in #698 (and I guess #701), it'd be nice to use proper ar archives instead of tar.

As this isn't a super urgent task, this would be a great issue for non-Embark folks to work on if you're interested!

eddyb:
Fixes #711 by an almost-`1:1` replacement of the [`tar`](https://docs.rs/tar/0.4.37/tar/) crate with the [`ar`](https://docs.rs/ar/0.8.0/ar/) crate. The documentation for those crates points to more information on the formats, but in short:
* [`ar` dates from 1971](https://en.wikipedia.org/wiki/Ar_(Unix)), holds a flat list of files, and is only used for `.a` static libraries (which contain the `.o` files they were created from, and optionally a symbol table) and `.deb` packages (for some reason)
  * it also has a BSD vs GNU split, but based on the GNU variant being used on Linux and Windows, I chose that, as it likely means better support
* [`tar` dates from 1979](https://en.wikipedia.org/wiki/Tar_(computing)), understands directory trees, and has been used for pretty much any archive purpose on UNIX systems (often compressed as e.g. `.tar.gz` or `.tar.xz`)

While using `tar` (or anything else, like, say, `zip`) doesn't really cause any issues, since we don't need to interop with any tools (for now), switching to `ar` does let us deduplicate some code (sadly for now only the part that finds the `rmeta` file inside the `.rlib`).

For future reference, the `rustc_codegen_cranelift` code for handling archives, which could eventually be moved into `rustc_codegen_ssa` (or in some other crate, more dedicated to linking) can be found in its [`src/archive.rs`](https://github.com/rust-lang/rust/blob/362e0f55eb1f36d279e5c4a58fb0fe5f9a2c579d/compiler/rustc_codegen_cranelift/src/archive.rs) (cc @bjorn3).

**EDIT**: there is some precedent for what we're doing, and also the reason for why "just" having a `.rmeta` file inside the `.rlib` works - on most platforms we actually wrap the `.rmeta` in an `.o`, but [WebAssembly doesn't need that](https://github.com/rust-lang/rust/blob/2fc3c69e5419292e92663a5f1e39203478925661/compiler/rustc_codegen_ssa/src/back/link.rs#L345-L348), which means the default "get `.rmeta` from `.rlib`" implementation in `rustc_codegen_ssa` can handle a plain `.rmeta` file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -78,6 +78,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "595d3cfa7a60d4555cb5067b99f07142a08ea778de5cf993f7b75c7d8fabc486"
 
 [[package]]
+name = "ar"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "450575f58f7bee32816abbff470cbc47797397c2a81e0eaced4b98436daf52e1"
+
+[[package]]
 name = "arrayvec"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2234,6 +2240,7 @@ checksum = "08d43f7aa6b08d49f382cde6a7982047c3426db949b1424bc4b7ec9ae12c6ce2"
 name = "rustc_codegen_spirv"
 version = "0.4.0-alpha.12"
 dependencies = [
+ "ar",
  "bimap",
  "hashbrown 0.11.2",
  "indexmap",
@@ -2249,7 +2256,6 @@ dependencies = [
  "smallvec",
  "spirv-tools",
  "syn",
- "tar",
  "tempfile",
  "topological-sort",
 ]
--- a/crates/rustc_codegen_spirv/Cargo.toml
+++ b/crates/rustc_codegen_spirv/Cargo.toml
@@ -35,6 +35,7 @@ num-traits = { version = "0.2", features = ["libm"] }
 syn = { version = "1", features = ["visit", "visit-mut"] }
 
 # Normal dependencies.
+ar = "0.8.0"
 bimap = "0.6"
 indexmap = "1.6.0"
 rspirv = { git = "https://github.com/gfx-rs/rspirv.git", rev = "c2a11ba7961fa7c675800f50c1903f0e96c633f8" }
@@ -44,7 +45,6 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 smallvec = "1.6.1"
 spirv-tools = { version = "0.6.1", default-features = false }
-tar = "0.4.30"
 topological-sort = "0.1"
 
 [dev-dependencies]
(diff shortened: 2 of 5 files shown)


khyperia (approved):

khyperia:
thanks @bjorn3 for the original pointer on this!

bjorn3:
> For future reference, the rustc_codegen_cranelift code for handling archives, which could eventually be moved into rustc_codegen_ssa (or in some other crate, more dedicated to linking) can be found in its src/archive.rs (cc @bjorn3).

Unfortunately not. The upstream ar crate doesn't handle symnol table generation. I am using a fork that does implement it. It works fine with Linux, FreeBSD and MinGW, but for some reason not on macOS. I hope that the object crate will eventually support writing object files and support macOS too.

eddyb:
Maybe a path forward could be `object` having a wrapper for the `ar` crate, that can read/write the symbol table information? Not sure how valuable `ar` is on its own without that support, but it makes sense on some level that you can have it for the "archive" part, and then `object` can handle "static library in `ar` format".

eddyb:
(disabled auto-merge until I can figure out why it failed on CI on windows - the `HACK` commit is for diagnosing that)

bjorn3:
> Not sure how valuable ar is on its own without that support,

Without symbol table no linker I know of will attempt to read anything from an archive.

> but it makes sense on some level that you can have it for the "archive" part, and then object can handle "static library in ar format".

The symbol table writing and archive writing need to be handled by the same crate. Symbol tables contain offsets to the start of the defining object files, not an index. The symbol table is always the first member of an archive, even for gnu style archives which have a name table before all user defined members. The symbol table format also depends on the archive format. There is a program called ranlib that can create the symbol tables after the fact, but because it is after the fact, it needs to rewrite the archive, which is slow.

khyperia:
> Without symbol table no linker I know of will attempt to read anything from an archive.

I mean, our linker does? :P - tbh though, there's no standard at all for shoving spir-v modules into an archive, and the concept of linking spir-v is only _barely_ included in the spec, and is extremely, extremely rare to do in the spir-v world. We're basically just using tar/ar as "rustc expects us to bundle multiple compilation results into one file, I guess we'll use ar, even though they're not even object files"

eddyb:
Something I forgot to mention is that WASM uses `ar` but doesn't bother wrapping the `.rmeta` file in some kind of WASM binary file, and [there's even a comment about it](https://github.com/rust-lang/rust/blob/2fc3c69e5419292e92663a5f1e39203478925661/compiler/rustc_codegen_ssa/src/back/link.rs#L345-L348) (I'll edit the PR description now).

eddyb:
Windows failure is https://github.com/mdsteele/rust-ar/issues/19 - I will work around it for now, but also try to fix it upstream.

PR #726, opened 2021-08-18
Various fixes and cleanup

khyperia:
While working on other patches that ended up not being applicable, I've gathered these changes unrelated to the irrelevant patch. So, submitting them as a seperate change, since the bigger change isn't going in.

### DIFF
--- a/crates/rustc_codegen_spirv/src/lib.rs
+++ b/crates/rustc_codegen_spirv/src/lib.rs
@@ -621,15 +621,20 @@ fn get_env_dump_dir(env_var: &str) -> Option<PathBuf> {
 /// This is the entrypoint for a hot plugged `rustc_codegen_spirv`
 #[no_mangle]
 pub fn __rustc_codegen_backend() -> Box<dyn CodegenBackend> {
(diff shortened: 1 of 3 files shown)


crates/rustc_codegen_spirv/src/lib.rs
```diff
@@ -621,15 +621,20 @@ fn get_env_dump_dir(env_var: &str) -> Option<PathBuf> {
...
+        // Override rustc's panic hook with our own to override the ICE error
+        // message, and direct people to `rust-gpu`.
+        std::panic::set_hook(Box::new(|panic_info| {
+            rustc_driver::report_ice(
+                panic_info,
+                "https://github.com/EmbarkStudios/rust-gpu/issues/new",
+            );
+            eprintln!("note: `rust-gpu` version {}\n", env!("CARGO_PKG_VERSION"));
+        }));
+    }
```
eddyb, line 637: I think this can be dropped now, or at least after #732 lands.

eddyb (changes requested):

eddyb (approved):

PR #728, opened 2021-08-18
Fix issues with some raytracing functions and add tests

### LINKED ISSUE #727: incorrect implementation of `spirv_std::ray_tracing::RayQuery::get_intersection_object_to_world`
Hi, I don't know much about Ray Query but I believe I found a bug in `spirv_std::ray_tracing::RayQuery::get_intersection_object_to_world`.

### Expected Behaviour

`spirv_std::ray_tracing::RayQuery::get_intersection_object_to_world` should call `OpRayQueryGetIntersectionObjectToWorldKHR` instead of `OpRayQueryGetWorldRayOriginKHR`.

https://github.com/EmbarkStudios/rust-gpu/blob/4dcb50169f7af2d97d0b59a6bb2d1bfc2cb4cfc7/crates/spirv-std/src/ray_tracing.rs#L637

`spirv_std::ray_tracing::RayQuery::get_intersection_object_to_world`  actualy calls `OpRayQueryGetWorldRayOriginKHR` but should call `OpRayQueryGetIntersectionObjectToWorldKHR`.

khyperia:
Fixes #727

Upon adding missing tests (about half of the methods were tested, the other half not, for some reason), I also discovered that get_world_ray_direction/origin were broken, and fixed those too.

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
+++ b/crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
@@ -1,5 +1,5 @@
 use super::Builder;
-use crate::builder_spirv::{BuilderCursor, SpirvValue, SpirvValueExt};
+use crate::builder_spirv::{BuilderCursor, SpirvValue};
 use crate::codegen_cx::CodegenCx;
 use crate::spirv_type::SpirvType;
 use rspirv::dr;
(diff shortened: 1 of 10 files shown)


crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
```diff
@@ -304,8 +304,26 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
...
-                self.err("OpTypeArray in asm! is not supported yet");
-                return;
+                let count = inst.operands[1].unwrap_id_ref();
+                let get_count_ty = || -> Option<Word> {
+                    let emit = self.emit();
+                    let func = &emit.module_ref().functions[emit.selected_function()?];
+                    let insts = &func.blocks[emit.selected_block()?].instructions;
+                    let inst = insts.iter().find(|i| i.result_id == Some(count))?;
+                    inst.result_type
+                };
```
eddyb, line 314: Where is this tested/exercised? I'm worried that the count constant wouldn't end up in the current function, but as a global. Or at least I didn't think constants are valid inside functions.

khyperia: It was required to be implemented for - oh, hold on, I actually deleted that bit of the `asm!`, replaced it with a `typeof*{result}`, I'll just revert this because it was partially hacky (but yeah, the constant does show up in the current function, even though it's not valid there - when rspirv reparses the module for linking, it reshuffles everything to its proper position. whee, hacks.)

eddyb: > when rspirv reparses the module for linking, it reshuffles everything to its proper position

Oh dear that explains things to me that subtly confused me months ago (but not enough for me to ask anyone or try to track it down in the code.

eddyb (approved):

Issue #723, opened 2021-08-17
Compiler panics when using #[spirv(object_to_world)]


hatoo:
Hi, I encountered compiler panic and I report here as output says.

### Expected Behaviour

The compiler doesn't panic.
I also want to know how do we use #[spirv(object_to_world)] correctly.

### Example & Steps To Reproduce

1. Create shader 

lib.rs

```rust 
#![cfg_attr(
    target_arch = "spirv",
    no_std,
    feature(register_attr),
    register_attr(spirv)
)]

use spirv_std::glam::Vec4;
#[cfg(not(target_arch = "spirv"))]
use spirv_std::macros::spirv;

#[spirv(closest_hit)]
pub fn test_closest_hit(#[spirv(object_to_world)] _object_to_world: [Vec4; 3]) {}

```
2. build it

build.rs

```rust
use std::error::Error;

use spirv_builder::{Capability, MetadataPrintout, SpirvBuilder};

fn main() -> Result<(), Box<dyn Error>> {
    SpirvBuilder::new("../shader", "spirv-unknown-vulkan1.2")
        .capability(Capability::RayTracingKHR)
        .extension("SPV_KHR_ray_tracing")
        .print_metadata(MetadataPrintout::Full)
        .build()?;

    Ok(())
}
```

3. get panic

```
> cargo build
    Blocking waiting for file lock on build directory
   Compiling builder v0.1.0 (C:\Users\hato2\Desktop\rust-gpu-issue\builder)
error: failed to run custom build command for `builder v0.1.0 (C:\Users\hato2\Desktop\rust-gpu-issue\builder)`

Caused by:
  process didn't exit successfully: `C:\Users\hato2\Desktop\rust-gpu-issue\target\debug\build\builder-8dcc5829546998ea\build-script-build` (exit code: 1)
  --- stderr
     Compiling core v0.0.0 (C:\Users\hato2\.rustup\toolchains\nightly-2021-08-10-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\core)
     Compiling rustc-std-workspace-core v1.99.0 (C:\Users\hato2\.rustup\toolchains\nightly-2021-08-10-x86_64-pc-windows-msvc\lib\rustlib\src\rust\library\rustc-std-workspace-core)
     Compiling compiler_builtins v0.1.49
     Compiling libm v0.2.1
     Compiling spirv-types v0.4.0-alpha.12
     Compiling bitflags v1.3.2
[...]

PR #734, opened 2021-08-21
Support `#[spirv(object_to_world)]` and `#[spirv(world_to_object)]`

hatoo:
Fixes https://github.com/EmbarkStudios/rust-gpu/issues/723#issuecomment-900967671.

This PR supports `#[spirv(object_to_world)]` and `#[spirv(world_to_object)]`.
This decrares `ObjectToWorldKHR`/`WorldToObjectKHR` builtins as `OpMatrix` type and unpack it to function local variable.

I added `SpirvType::Matrix` variant but some of the other codes left `todo!` because I don't know how to test it when implemented.

I've confirmed that it works in my project.


hatoo:
#738

Issue #82, opened 2020-10-21
repr(simd) for matrix types


Jasper-Bekkers:
This is super low prio because I don't think anybody actually cares about OpTypeMatrix at the moment, but it might be nice to have.

termhn:
related to this and also kinda #121 , what does repr(simd) actually do on these gpu types?

khyperia:
Inside rustc, `repr(simd)` types are called ["Vector"](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_target/abi/enum.Abi.html#variant.Vector), and, we just translate those directly to spir-v [OpTypeVector](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpTypeVector).

https://github.com/EmbarkStudios/rust-gpu/blob/8681464af7e2500f371e5df6a95691cf29ecd16a/rustc_codegen_spirv/src/abi.rs#L336-L343

So, this issue isn't actually about `repr(simd)` for matrix types, because we can't hijack the existing system for that, but rather probably creating a spirv builtin attribute that translates the struct to be OpTypeMatrix, but whatever, close enough. (The attribute would not publicly documented and only used in spirv-std, similar to the current `storage_class`)

Boscop:
> This is super low prio because I don't think anybody actually cares about `OpTypeMatrix` at the moment, but it might be nice to have.

I care about it :) 
In fact for my use case I need `Mat3` and `Mat4` as uniforms that match the native `mat3`/`mat4` types.

khyperia:
I don't think this is on our roadmap any time soon, but is a relatively self-contained nugget of a task that a community member could feasibly do that wouldn't interfere much with anything else, so I'll leave this open for now even though it's technically kind of a `wontfix`.

PR #738, opened 2021-08-29
Implement OpTypeMatrix

hatoo:
Fixes https://github.com/EmbarkStudios/rust-gpu/issues/82.

This PR adds `#[spirv(matrix(ty, m, n))]` attribute. A type with `#[spirv(matrix(..))]` is expressed as `OpTypeMatrix` in SPIR-V.
An example usage is tests/ui/spirv-attr/matrix-type.rs.

~~While `OpTypeVector` is implemented by utilizing `#[repr(simd)]`, this PR implements `OpTypeMatrix` by introducing the new trait `Matrix` because I couldn't find a usable attribute.~~

~~When converting `Ty` to SPIRV type, it uses `OpTypeMatrix` if the `Ty` implements `Matrix`.~~

## TODO

- Make glam matrices `OpMatrix`
Since many operations in glam's matrices cast self to inner type, we need to add attributes to glam matrices and its inner types.
``~~Currently, we can't add the attribute to `glam::core::storage::Column*<V>` because we can't know its vector length at definition.
We may handle it by adding `#[spirv(matrix(m, param(n)))]` syntax and refer nth generics type (e.g. glam::Vec3) and infer vector length from it...~~


hatoo:
I've realized that we can implement this with custom attributes and I think it's better.
I.ve removed `Matrix` trait and added `#[spirv(matrix(..))]`.
I'll revert it if you prefer the `Matrix` trait.

hatoo:
I added 

- #[spirv(matrix(ty, m))]
- #[spirv(matrix(ty))]
- #[spirv(matrix)]

These infer missing parameters from type.

khyperia:
Please only implement `#[spirv(matrix)]` (the inferred one), it's both the preferential design and simplifies a lot of stuff.

hatoo:
Make sense!
I dropped other than `#[spirv(matrix)]`

crates/rustc_codegen_spirv/src/abi.rs
```diff
@@ -843,5 +843,43 @@ fn trans_intrinsic_type<'tcx>(
...
+                count: n as u32,
+            }
+            .def(span, cx);
+
+            Ok(SpirvType::Matrix {
+                element: vector,
+                count: m as u32,
+            }
+            .def(span, cx))
+        }
```
khyperia, line 883: You don't have to (and shouldn't) do this digging into field types for this, and should use the standard type translation tools on the fields instead. Also, this is missing quite a bit of validation (e.g. if fields are different types, or there are no fields - IIRC your code ICEs if there's no fields) - there should be tests for each of the three kinds of error. Something like this should work instead:

```rust
IntrinsicType::Matrix => {
    let field_types = (0..ty.fields.count())
        .map(|i| trans_type_impl(cx, span, ty.field(cx, i), false))
        .collect::<Vec<_>>();
    if field_types.is_empty() {
        cx.tcx
            .sess
            .err("#[spirv(matrix)] type must have at least one field");
        return Err(ErrorReported);
    }
    let elem_type = field_types[0];
    if !field_types.iter().all(|&ty| ty == elem_type) {
        cx.tcx
            .sess
            .err("#[spirv(matrix)] type fields must all be the same type");
        return Err(ErrorReported);
    }
    match cx.lookup_type(elem_type) {
        SpirvType::Vector { .. } => (),
        ty => {
            cx.tcx
                .sess
                .struct_err("#[spirv(matrix)] type fields must all be vectors")
                .note(&format!("field type is {}", ty.debug(elem_type, cx)))
                .emit();
            return Err(ErrorReported);
        }
    }

    Ok(SpirvType::Matrix {
        element: elem_type,
        count: field_types.len() as u32,
    }
    .def(span, cx))
}
```

tests/ui/spirv-attr/matrix-type.rs
```diff
@@ -0,0 +1,43 @@
...
+        Self::IDENTITY
+    }
+}
+
+#[spirv(closest_hit)]
+pub fn main(
+    #[spirv(object_to_world)] _object_to_world: Affine3,
+    #[spirv(world_to_object)] _world_to_object: Affine3,
+) {
+}
```
khyperia, line 43: I believe this isn't actually testing much due to the majority of it being dead code. (Also, nit, the name `Affine3` is a little misleading - it's not an affine transformation - but isn't super important for a test, haha)

I would like to see field accesses/etc. tested as well, I'm nervous about just crossing our fingers and hoping that matricies behave exactly like structs in all ways and no instructions need to be modified to handle matricies specially.

khyperia (changes requested):
I believe `tests/ui/spirv-attr/{invalid-target.rs,multiple.rs}` need to be updated.

crates/rustc_codegen_spirv/src/abi.rs
hatoo: Thanks.
I updated the code and change Matrix length validation since `OpTypeMatrix` requires a length of at least 2.
https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#OpTypeMatrix

tests/ui/spirv-attr/matrix-type.rs
hatoo: Oops, I chose `Affine3` because `glam::f32::Affine3A` has the same layout https://docs.rs/glam/0.18.0/glam/f32/struct.Affine3A.html (Sry, I don't know about math 😨).

I feel adding more tests for field operations in `tests/ui/spirv-attr/* ` is not appropriate because it may should contain only about attr tests.
Is it OK? or is there any better place?

khyperia: Oh! whoops, misread it as 4 Vec4s, not 4 Vec3s, 4 Vec3s definitely _is_ an affine transform, haha, sorry

Yeah, not totally sure about where to put tests, anywhere is probably fine

hatoo: I added tests `to matrix-type.rs` I think this is enough but are there suggestions for more cases?

hatoo:
Thanks for pointed out.
I updated `tests/ui/spirv-attr/multiple.rs` but failed to update `tests/ui/spirv-attr/invalid-target.rs`...

Because of the error
```
error: failed to decode compiler output as json: `EOF while parsing a string at line 1 column 239`
```

khyperia:
> I simply replaced `sampler, block, sampled_image, generic_image_type, // struct-only` to `sampler, block, sampled_image, generic_image_type, matrix, // struct-only`.
> 
> But got the error
> 
> ```
> error: failed to decode compiler output as json: `EOF while parsing a string at line 1 column 239`
> ```

Looks like compiletest is just straight up broken if output is too long - it injects [this into the output](https://github.com/rust-lang/rust/blob/dbb0fe9d803f7fa8688b33306266a6b054c2c3f4/src/tools/compiletest/src/runtest.rs#L4008) and then later tries to [parse the output as json](https://github.com/rust-lang/rust/blob/dbb0fe9d803f7fa8688b33306266a6b054c2c3f4/src/tools/compiletest/src/runtest.rs#L3247), tripping over its own trimmed marker.

Feel free to just skip updating `invalid-target.rs` for now. (or, split it into multiple files if you feel like it so the output is shorter, but, I can do that after this PR is merged)

hatoo:
I can't reproduce the CI failure.
I've tested on Windows 11 and Ubuntu20.04 (both WSL and native).

khyperia:
I'm also unable to reproduce, even with CI's exact version of spirv-tools. hmm. I'll try rerunning tests, I guess?

khyperia:
Was able to reproduce locally, the full error was cut off in CI:

```
error: line 70: OpEntryPoint Entry Point <id> '4[%4]'s callgraph contains function <id> 4[%4], which cannot be used with the current execution model:
[VUID-StandaloneSpirv-None-04644] in Vulkan evironment, Output Storage Class must not be used in GLCompute, RayGenerationKHR, IntersectionKHR, AnyHitKHR, ClosestHitKHR, MissKHR, or CallableKHR execution models

  %4 = OpFunction %void None %28
  ```
  
  the issue seems to be an output (`&mut`) var being used in `closest_hit` functions

**Comment by hatoo** [comment]
Thank you!
I updated test cases.

**Comment by khyperia** [comment]
could you add a `// build-fail` test that triggers the three new errors? (also remove the accidentally commited `module` file) - other than that, looks pretty good!

**Comment by hatoo** [comment]
I don't still understand why
```
#[spirv(fragment)]
pub fn main_add(in1: Affine3, in2: Affine3, out: &mut glam::Vec3) {
    *out = in1.x + in1.y + in1.z + in1.w + in2.x + in2.y + in2.z + in2.w;
}
```
failed but reverted to previous code for now.

khyperia:
Ah, I think that failure is due to a similar thing as https://github.com/EmbarkStudios/rust-gpu/pull/513, which, might become more pressing to fix once people actually start using matrices. TL;DR matricies take up a handful of binding slots, so incrementing binding slot by 1 per parameter doesn't work.

hatoo:
I added failing tests.

I added two test files because it seems not able to test #[(spirv(matrix))] for empty struct with others in the same file.

Error messages feel unkind because it doesn't show type definition.
But we can't get a span of type definition in the function, right?

hatoo:
Thanks, I choose to use `def_id_for_spirv_type_adt` and it will returns never None because of `trans_type_impl`.

khyperia (approved):
:tada:

hatoo:
Woo! 🎉 🎉 🎉 
I really appreciate your support and thank you for your time! 🙏

PR #742, opened 2021-08-31
Add a spirv-builder option to include all debug info

### LINKED ISSUE #736: Add second option for adding OpNames to whole module
Right now, OpNames for the whole module is enabled by setting debuginfo to be enabled in cargo (passed on to rustc). This isn't very discoverable, and while docs should definitely be improved to make it be more discoverable, it'd be nice to add _another_ option to enable OpNames for the whole module in spirv-builder - similar to how `name_variables` exists today. (both the spirv-builder and cargo config option would do the same thing)

(maybe change `name_variables` into a more generic sounding option, and turn it into an `enum {None, Variables, Everything}` or something, idk)

khyperia:
This is in addition to the cargo debug information option doing the same thing. Fixes #736

### DIFF
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -125,13 +125,7 @@ impl fmt::Display for SpirvBuilderError {
 
 impl Error for SpirvBuilderError {}
 
-pub enum MemoryModel {
-    Simple,
-    Vulkan,
-    GLSL450,
-}
-
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub enum MetadataPrintout {
     /// Print no cargo metadata.
     None,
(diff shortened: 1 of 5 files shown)


crates/spirv-builder/src/lib.rs
```diff
@@ -399,8 +402,12 @@ fn invoke_rustc(builder: &SpirvBuilder) -> Result<PathBuf, SpirvBuilderError> {
...
     }
-    if builder.name_variables {
-        llvm_args.push("--name-variables");
+    if builder.spirv_metadata != SpirvMetadata::None {
+        if builder.spirv_metadata == SpirvMetadata::Full {
+            llvm_args.push("--spirv-metadata=full");
+        } else {
+            llvm_args.push("--spirv-metadata=name-variables");
+        }
     }
```
eddyb, line 411: This could probably be a `match`.

eddyb (approved):
Only a nitpick, feel free to merge as-is.

PR #757, opened 2021-09-27
Add an `read_clock_khr` function that calls `OpReadClockKHR`

expenses:
Reading from the shader clock lets us create heatmaps for ray tracing as seen here: https://developer.nvidia.com/blog/profiling-dxr-shaders-with-timer-instrumentation/.

heatmap

### DIFF
--- a/crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
+++ b/crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
@@ -745,7 +745,7 @@ pub fn instruction_signatures(op: Op) -> Option<&'static [InstSig<'static>]> {
         // SPV_AMD_shader_fragment_mask
         Op::FragmentMaskFetchAMD | Op::FragmentFetchAMD => reserved!(SPV_AMD_shader_fragment_mask),
         // SPV_KHR_shader_clock
-        Op::ReadClockKHR => reserved!(SPV_KHR_shader_clock),
+        Op::ReadClockKHR => {}
         // SPV_NV_mesh_shader
         Op::WritePackedPrimitiveIndices4x8NV => reserved!(SPV_NV_mesh_shader),
         // SPV_NV_ray_tracing
--- a/crates/spirv-std/src/arch.rs
+++ b/crates/spirv-std/src/arch.rs
@@ -150,3 +150,25 @@ pub unsafe fn vector_insert_dynamic<T: Scalar, V: Vector<T, N>, const N: usize>(
 pub fn kill() -> ! {
     unsafe { asm!("OpKill", options(noreturn)) }
 }
+
+/// Read from the shader clock.
+///
+/// Requires the `SPV_KHR_shader_clock` extension and the `ShaderClockKHR` capability.
+///
+/// See:
+/// <https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_shader_clock.html>
+#[spirv_std_macros::gpu_only]
+#[doc(alias = "OpReadClockKHR")]
+pub unsafe fn read_clock_khr() -> u64 {
+    let mut result: u64 = 0;
+
+    asm! {
+        "%uint = OpTypeInt 32 0",
+        "%uint_3 = OpConstant %uint 3",
+        "%result = OpReadClockKHR typeof*{result} %uint_3",
+        "OpStore {result} %result",
+        result = in(reg) &mut result,
+    };
+
+    result
+}


crates/spirv-std/src/arch.rs
```diff
@@ -150,3 +150,25 @@ pub unsafe fn vector_insert_dynamic<T: Scalar, V: Vector<T, N>, const N: usize>(
...
+/// See:
+/// <https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_shader_clock.html>
+#[spirv_std_macros::gpu_only]
+#[doc(alias = "OpReadClockKHR")]
+pub unsafe fn read_clock_khr() -> u64 {
+    let mut result: u64 = 0;
+
+    asm! {
+        "%uint = OpTypeInt 32 0",
+        "%uint_3 = OpConstant %uint 3",
```
expenses, line 167: I'm not entirely sure how to do this 'scope' field. It has to be an `OpConstant` so I'm not sure if that's something we can use `in(reg)` for. 

With glsl:

```glsl
#version 450
#extension GL_EXT_shader_explicit_arithmetic_types_int64 : enable
#extension GL_ARB_shader_clock : enable

void main() {
	uint64_t clock_value = clockARB();
}
```

becomes:

```spirv
; SPIR-V
; Version: 1.0
; Generator: Google Shaderc over Glslang; 10
; Bound: 12
; Schema: 0
               OpCapability Shader
               OpCapability Int64
               OpCapability ShaderClockKHR
               OpExtension "SPV_KHR_shader_clock"
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %main "main"
               OpExecutionMode %main OriginUpperLeft
               OpSource GLSL 450
               OpSourceExtension "GL_ARB_shader_clock"
               OpSourceExtension "GL_EXT_shader_explicit_arithmetic_types_int64"
               OpSourceExtension "GL_GOOGLE_cpp_style_line_directive"
               OpSourceExtension "GL_GOOGLE_include_directive"
               OpName %main "main"
               OpName %clock_value "clock_value"
       %void = OpTypeVoid
          %3 = OpTypeFunction %void
      %ulong = OpTypeInt 64 0
%_ptr_Function_ulong = OpTypePointer Function %ulong
       %uint = OpTypeInt 32 0
     %uint_3 = OpConstant %uint 3
       %main = OpFunction %void None %3
          %5 = OpLabel
%clock_value = OpVariable %_ptr_Function_ulong Function
         %11 = OpReadClockKHR %ulong %uint_3
               OpStore %clock_value %11
               OpReturn
               OpFunctionEnd
```

so I just copied that constant.

msiglreith: Using `const` in combination with the `Scope` enum from `memory` would be an option:
```rust
let mut result: u64;

asm! {
    "%uint = OpTypeInt 32 0",
    "%scope = OpConstant %uint {scope}",
    "{result} = OpReadClockKHR typeof*{result} %scope",
    result = out(reg) result,
    scope = const Scope::Subgroup as _,
};

result
```

expenses: Ooh, yeah that makes sense!

crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
```diff
@@ -745,7 +745,7 @@ pub fn instruction_signatures(op: Op) -> Option<&'static [InstSig<'static>]> {
         // SPV_AMD_shader_fragment_mask
         Op::FragmentMaskFetchAMD | Op::FragmentFetchAMD => reserved!(SPV_AMD_shader_fragment_mask),
         // SPV_KHR_shader_clock
-        Op::ReadClockKHR => reserved!(SPV_KHR_shader_clock),
+        Op::ReadClockKHR => {}
```
khyperia, line 748: ping @eddyb to review this bit, I forget how spirv_type_constraints works

crates/spirv-std/src/arch.rs
```diff
@@ -150,3 +150,26 @@ pub unsafe fn vector_insert_dynamic<T: Scalar, V: Vector<T, N>, const N: usize>(
...
+pub unsafe fn read_clock_khr() -> u64 {
+    let mut result: u64 = 0;
+
+    asm! {
+        "%uint = OpTypeInt 32 0",
+        "%scope = OpConstant %uint {scope}",
+        "%result = OpReadClockKHR typeof*{result} %scope",
+        "OpStore {result} %result",
+        result = in(reg) &mut result,
+        scope = const crate::memory::Scope::Subgroup as _,
```
khyperia, line 171: Shouldn't this be a const generic, similar to barriers?

crates/spirv-std/src/arch.rs
```diff
@@ -150,3 +150,26 @@ pub unsafe fn vector_insert_dynamic<T: Scalar, V: Vector<T, N>, const N: usize>(
...
+#[doc(alias = "OpReadClockKHR")]
+pub unsafe fn read_clock_khr() -> u64 {
+    let mut result: u64 = 0;
+
+    asm! {
+        "%uint = OpTypeInt 32 0",
+        "%scope = OpConstant %uint {scope}",
+        "%result = OpReadClockKHR typeof*{result} %scope",
+        "OpStore {result} %result",
+        result = in(reg) &mut result,
```
khyperia, line 170: I believe this can be `out(reg) result`, with `{result} = OpReadClockKHR ...`

crates/spirv-std/src/arch.rs
```diff
@@ -150,3 +150,26 @@ pub unsafe fn vector_insert_dynamic<T: Scalar, V: Vector<T, N>, const N: usize>(
...
+
+/// Read from the shader clock.
+///
+/// Requires the `SPV_KHR_shader_clock` extension and the `ShaderClockKHR` capability.
+///
+/// See:
+/// <https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_shader_clock.html>
+#[spirv_std_macros::gpu_only]
+#[doc(alias = "OpReadClockKHR")]
+pub unsafe fn read_clock_khr() -> u64 {
```
khyperia, line 162: I think this should have a `#[cfg(target_feature = "Int64")]`. It would be nice to also include the u32-tuple version, to allow this to be used where the Int64 capability is not available.

khyperia (changes requested):

khyperia:
Please also add compiletests

crates/spirv-std/src/arch.rs
expenses: Yep, using a const generic here makes sense.

https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_shader_clock.html#_description says:

> The two valid SPIR-V scopes for `OpReadClockKHR` are `Subgroup` and `Device`.

so we probably want to have a new enum with only those two scopes, to limit invalid usage.

expenses: We can't use a scope as a const generic yet though:

```rust
error: constant expression depends on a generic parameter
   --> crates\spirv-std\src\arch.rs:189:23
    |
189 |         scope = const SCOPE as u32,
    |                       ^^^^^^^^^^^^
    |
    = note: this may fail depending on what value the parameter takes
                ^^^^^^^^^^^^
 ```

and if we add `#![feature(const_evaluatable_checked)]` to the crate we get:

```rust
error: unconstrained generic constant
   --> crates\spirv-std\src\arch.rs:189:23
    |
189 |         scope = const SCOPE as u32,
    |                       ^^^^^^^^^^^^
    |
    = help: try adding a `where` bound using this expression: `where [(); SCOPE as u32]:`
```

khyperia: Check how barriers do this (barrier.rs), they do this exact thing.

(and for some context, arch.rs is not intended to be ergonomic or safe to use, it's supposed to be the building blocks of a good abstraction)

expenses: That's fair. Hopefully when const generics can implemented a little further we'll be able to ensure (some) correctness though.

tests/ui/arch/read_clock_khr.rs
```diff
@@ -0,0 +1,16 @@
+// build-pass
+// compile-flags: -Ctarget-feature=+Int64,+ShaderClockKHR,+ext:SPV_KHR_shader_clock
```
khyperia, line 2: Ah, shoot, right, I'm silly and completely missed this the first time around, there should also be a `#[cfg(and(target_feature = "ShaderClockKHR", target_feature = "ext:SPV_KHR_shader_clock"))]` put on too, to both functions, sorry!

expenses:
Hmm, `compiletests` doesn't seem to like those `#[cfg]` flags.

khyperia:
... oh, ugh, right, since we build spirv-std without those flags, but the test itself with those flags (something that never happens, and cannot happen, in the real world - can only happen here since we're directly invoking rustc instead of going through cargo). uuuuh, hmm, going to try to think of a way to resolve it...

crates/rustc_codegen_spirv/src/spirv_type_constraints.rs
eddyb: Looking at [the extension](https://github.com/KhronosGroup/SPIRV-Registry/blob/004f312872a014d6b7cf33e33fbd235c33a6be27/extensions/KHR/SPV_KHR_shader_clock.asciidoc#instructions), looks like the instruction doesn't have multiple types that are related, so nothing to here.

However, I'd suggest this change:
```suggestion
        Op::ReadClockKHR => {
            // NOTE(eddyb) we actually use these despite not being in the standard yet.
            // reserved!(SPV_KHR_shader_clock)
        }
```

We're ending up with a lot of these, I wish the standard contained the current status of the extensions instead of just describing the instructions as "reserved" (which is why I initially avoided handling them here).

tomachinz:
nanosecond timing will be useful.

_Timestamp values in this time domain are in units of nanoseconds and are comparable with platform timestamp values captured using the POSIX clock_gettime API. This extension advertises the SPIR-V ShaderClockKHR capability for Vulkan, which allows a shader to query a real-time or monotonically incrementing counter at the subgroup level or across the device level. The two valid SPIR-V scopes for OpReadClockKHR are Subgroup and Device._

khyperia:
This is a horrible no-good very bad solution, and I'm really hoping I can think of something better, but right now my only thought to fix it is [adding ShaderClockKHR and the ext to here](https://github.com/EmbarkStudios/rust-gpu/blob/6780432b37f337fb7dd5f2a9b07cf06992198109/tests/src/main.rs#L303).

expenses:
> 
> 
> This is a horrible no-good very bad solution, and I'm really hoping I can think of something better, but right now my only thought to fix it is [adding ShaderClockKHR and the ext to here](https://github.com/EmbarkStudios/rust-gpu/blob/6780432b37f337fb7dd5f2a9b07cf06992198109/tests/src/main.rs#L303).

That doesn't look too too bad to me :)

Another think that I thought off that could to have a seperate feature flag just for the compiletests:

```rust
#[cfg(any(
    all(
        target_feature = "Int64",
        target_feature = "ShaderClockKHR",
        target_feature = "ext:SPV_KHR_shader_clock"
    ),
    target_feature = "compiletests"
)]
```

That way we don't have to shove a whole lot of features into that line you linked.

PR #765, opened 2021-10-12
Update dependencies

khyperia:


khyperia:
I said this in discord already, but to add context where it's relevant, this is currently blocked on https://github.com/EmbarkStudios/ash-molten/pull/56 getting merged - I'd rather not reference a git fork. Either ash-molten will do another release once that's merged, or we can reference ash-molten git main.

This is because [ash 0.32 has cargo-deny issues](https://rustsec.org/advisories/RUSTSEC-2021-0090), so, need to update to ash 0.33, but ash-molten doesn't have support for ash 0.33 yet (that PR adds it).

Issue #151, opened 2020-10-26
Support Vulkan Debug Printf validation layer


termhn:
It would be cool to be able to pipe `print!()`, `dbg!()` and friends (or introduce similar gpu-specific versions) through to https://vulkan.lunarg.com/doc/sdk/1.2.148.1/windows/debug_printf.html

nipunG314:
Is anyone currently working on this? If not, I would like to take a stab at it.

bjorn3:
I don't know how easy it would be to make it work with rust-gpu, but would it make sense to use [`defmt`](https://github.com/knurling-rs/defmt) for better performance due to less copying on the gpu side? Or is it expected that the amount of prints is very low.

nipunG314:
The article linked mentions above that, by default, that the shader will return a value to the validation layer every time the debug statement is encountered in the shader code. The article mentions that steps can be taken to prevent this from happening, but doesn't highlight any of them. We'll probably have to look into that.

If we don't limit the number of messages, the performance will be very slow, especially in the fragment shader.

termhn:
> Is anyone currently working on this? If not, I would like to take a stab at it.

Nobody is working on this atm, no. Go for it :)

> The article linked mentions above that, by default, that the shader will return a value to the validation layer every time the debug statement is encountered in the shader code. The article mentions that steps can be taken to prevent this from happening, but doesn't highlight any of them. We'll probably have to look into that.
> 
> If we don't limit the number of messages, the performance will be very slow, especially in the fragment shader.

yeah, perf is always going to be bad if you're printing from a shader though so I wouldn't worry too much about micro perf opts for this path atm.

khyperia:
`format_args!` and related infrastructure (`println!`, etc.) may be difficult to re-use. As far as I understand it, arguments are stored into a [`fmt::ArgumentV1` struct](https://github.com/rust-lang/rust/blob/07e968b640e8ff76fa8be4b48b70ab80ea577800/library/core/src/fmt/mod.rs#L237-L240), which contains a pointer to a `fmt::Opaque`, as well as a function pointer that takes a `fmt::Opaque` pointer, casts it to the right type, and formats it.

This is problematic for multiple reasons:

1) Casting a pointer to an arbitrary type to a pointer to a `fmt::Opaque` is not supported in spir-v (nor GPU hardware) (so no transmute on pointers, etc.)
2) Function pointers are not supported in spir-v (so no vtables too, etc.)
[...]

PR #768, opened 2021-10-16
Add `debug_printf!` and `debug_printfln!` macros that uses the DebugPrintf extension

expenses:
See https://github.com/EmbarkStudios/rust-gpu/issues/151.

This will probably require some work before being merged. These macros and different from the core rust `print` and `println` macros as the type of variables needs to be specified in the format string. There's probably a way to change this as mentioned here: https://github.com/EmbarkStudios/rust-gpu/issues/151#issuecomment-717584254 but for now I'm doing the simpler thing.

I have a branch here: https://github.com/EmbarkStudios/rust-gpu/compare/main...expenses:printf-working that modifies the sky shader and ash runner for testing. It requires 'Debug Printf' to be checked in the Vulkan Configuration app.

One important caveat is that using DebugPrintf seems to SEGFAULT (on Windows I get a (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)) the Vulkan program if you build a single module with multiple entry points. This is truely bizarre, no clue what's going on here. The only difference between working code and segfaulting code is this: https://github.com/expenses/rust-gpu/compare/expenses:printf-not-working...printf-working.

At first I thought this was a driver issue but I'm getting this on both an integrated intel gpu and an external amd gpu.

### DIFF
--- a/crates/spirv-std/macros/src/lib.rs
+++ b/crates/spirv-std/macros/src/lib.rs
@@ -430,7 +430,7 @@ fn debug_printf_inner(input: DebugPrintfInput) -> TokenStream {
 
     fn map_specifier_to_type(
         specifier: char,
-        chars: &mut impl Iterator<Item = char>,
+        chars: &mut std::str::Chars<'_>,
     ) -> Option<proc_macro2::TokenStream> {
         let mut peekable = chars.peekable();
 


expenses:
Fun fact: you can print emojis using these macros

expenses:
> we could map each format argument to the rust type - `%f` to `f32` for example - and wrap each one in `assert_is_type::<T>` before using them. It's a bit gross but it would work.

I did a test of this and it actually works quite nicely:

```rust
error[E0308]: mismatched types
  --> $DIR/debug_printf.rs:41:31
   |
41 |         debug_printfln!("%f", vec);
   |                               ^^^ expected `f32`, found struct `Vec2`
```

khyperia:
> Just tested this - it just prints `r` without a `%`.

what in the absolute heckity goodness, lmao, that's wild

anyway, yeah, probably fine to merge as-is and work on full validation later if you'd prefer that PR model!

as for the path test failure, I thiiink there might be a way to work around that with something like [this](https://github.com/EmbarkStudios/rust-gpu/blob/28313a2029aec5de7283e167ea2d67d2aa350f62/tests/ui/image/gather_err.rs#L2), although that might have to be tweaked slightly

PR #778, opened 2021-10-27
No longer require the workaround for the collatz example

DJMcNab:
Follow up from #776, which works exactly as required <3


khyperia:
Now that https://github.com/EmbarkStudios/rust-gpu/pull/809 is in, could you rebase to main and remove the `Int8` cap?

examples/runners/wgpu/builder/src/main.rs
```diff
@@ -26,14 +18,10 @@ fn build_shader(
 }
 
 fn main() -> Result<(), Box<dyn Error>> {
-    build_shader("../../../shaders/sky-shader", true, &[])?;
-    build_shader("../../../shaders/simplest-shader", false, &[])?;
+    build_shader("../../../shaders/sky-shader", true)?;
+    build_shader("../../../shaders/simplest-shader", false)?;
     // We need the int8 capability for using `Option`
```
khyperia, line 23: please remove this comment~

PR #792, opened 2021-11-03
compiletests: unlock `// {only,ignore}-*` usage.

eddyb:
~~*Opening as draft because https://github.com/Manishearth/compiletest-rs/pull/249 isn't merged yet.*~~

This enables the test from #789 to correctly run only on `vulkan1.1` targets, and be ignored on other targets (tested both with the default and `--target-env opengl4.5`).

We should also be able to use `// ignore-*` now, without `compiletest-rs` panicking.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -347,9 +347,8 @@ dependencies = [
 
 [[package]]
 name = "compiletest_rs"
-version = "0.7.0"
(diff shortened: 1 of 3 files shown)


eddyb:
https://github.com/Manishearth/compiletest-rs/pull/249 is now merged, this just needs rubber-stamping.

hrydgard (approved):

PR #800, opened 2021-11-15
Remove unused dependencies in rust-gpu

bnjbvr:


VZout (approved):

examples/runners/ash/Cargo.toml
```diff
@@ -16,16 +16,10 @@ use-compiled-tools = ["spirv-builder/use-compiled-tools"]
...
-serde_json = "1.0"
 structopt = "0.3.20"
 cfg-if = "1.0.0"
 shared = { path = "../../shaders/shared" }
 spirv-builder = { path = "../../../crates/spirv-builder", default-features = false }
 
-# TODO: Remove this once no longer needed, only needed to make cargo-deny happy for some reason.
-# https://rustsec.org/advisories/RUSTSEC-2021-0119
-nix = "0.20.2"
-
```
eddyb, line 29: What makes this possible? I don't understand what else changed to allow removing this.

bnjbvr: Unclear, but it seems that the Cargo.lock from this patch references only `nix 0.20.2` and `nix 0.22.2`, which are both said to include the patch according to https://rustsec.org/advisories/RUSTSEC-2021-0119.

khyperia: Ah. Figured it out. This should be kept, and not removed.

@bnjbvr some additional context as to why this was included is here https://github.com/EmbarkStudios/rust-gpu/pull/765#discussion_r727076669 . Long story short, cargo had the choice between `bitflags 1.3.2, nix 0.20.0, nix 0.22.0` and `bitflags 1.2.1, nix 0.20.2, nix 0.22.2`, and cargo was choosing the first combination (which had the vuln). Adding this explicit reference forces it to choose the latter.

However, once the reference has been forced, it's stored in `Cargo.lock`, and won't be reverted - subsequently removing the reference (as in this PR) keeps the choice of `bitflags 1.2.1, nix 0.20.2, nix 0.22.2`. But, if you run `cargo update`, it reverts back to the first choice, and will fail the vuln check:

```
(on this branch)> cargo update
    Updating bitflags v1.2.1 -> v1.3.2
    Removing nix v0.20.2
    Removing nix v0.22.2
      Adding nix v0.20.0
      Adding nix v0.22.0
```

khyperia (changes requested):
(see previous comment)

khyperia (approved):
neat, thanks!

PR #794, opened 2021-11-09
Updated readme, added list of related projects

ColeSeverson:
This adds a list for related projects, projects related to running Rust on GPUs, below the contribution section in the main README.MD
The list is ordered by last update to the repo, with ongoing and in-maintainance projects marked as such. 

Fix #20

PR #802, opened 2021-11-19
Remove projects that aren't compiling Rust code to GPUs

khyperia:
https://github.com/EmbarkStudios/rust-gpu/pull/794 included some projects that aren't actually compilers, but rather host-side bindings of graphics/compute APIs:

- The `ocl` crate is bindings to the OpenCL API - GPU source is passed in via OpenCL source string.
- The `emu` crate is a wrapper around wgpu to make it feel like CUDA - source is passed in via e.g. GLSL string.
- `RustaCUDA` is a wrapper around CUDA - source is passed in via binary blobs, e.g. compiled by `nvcc`.


repi (approved):
good catch! that PR got merged too early (thought it needed all reviews).

should indeed only list projects that are related to running Rust code in some way or another on GPUs

PR #803, opened 2021-11-20
asm: support `OpCompositeExtract` inference for `OpTypeStruct` fields.

eddyb:
The motivation for this was unchecked indexing via `asm!`, on `&[T]` (as can be seen in the test) - all the pieces were there, except for the `OpCompositeExtract` to get the data pointer part of the `&[T]`.


khyperia (approved):
nice!

PR #805, opened 2021-11-24
Add an `IndexUnchecked` trait that uses `asm!`

expenses:
After https://github.com/EmbarkStudios/rust-gpu/pull/803 it became possible to write `asm!` functions that index into arrays and `RuntimeArray`s without bounds checking. This PR exposes these functions as a trait. I decided on `index_unchecked*` instead of `get_unchecked*` to avoid the name aliasing. It could still be pretty confusing though.

### DIFF
--- a/crates/spirv-std/src/arch.rs
+++ b/crates/spirv-std/src/arch.rs
@@ -246,7 +246,7 @@ pub fn signed_max<T: SignedInteger>(a: T, b: T) -> T {
     unsafe { call_glsl_op_with_ints::<_, 42>(a, b) }
 }
 
-/// Index into an array or RuntiemArray without bounds checking.
+/// Index into an array or `RuntimeArray` without bounds checking.
 pub trait IndexUnchecked<T> {
     /// Returns a reference to the element at `index`. The equivalent of `get_unchecked`.
     unsafe fn index_unchecked(&self, index: usize) -> &T;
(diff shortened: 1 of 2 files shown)


crates/spirv-std/src/arch.rs
```diff
@@ -245,3 +245,69 @@ pub fn signed_min<T: SignedInteger>(a: T, b: T) -> T {
 pub fn signed_max<T: SignedInteger>(a: T, b: T) -> T {
     unsafe { call_glsl_op_with_ints::<_, 42>(a, b) }
 }
+
+/// Index into an array or `RuntimeArray` without bounds checking.
```
khyperia, line 249: Nit: `RuntimeArray` is not implemented here, so it shouldn't be mentioned.

Also, it'd be nice to explain _why_ this exists when `get_unchecked` exists as well.

khyperia (changes requested):

khyperia:
Also, it'd be nice to not use `gpu_only` here, because it's not actually gpu_only

crates/spirv-std/src/arch.rs
expenses: Oh yeah I guess we shouldn't mention the `RuntimeArray` struct, but what do you call a `&[T]` when used in SPIR-V as a storage buffer?

khyperia: I'd probably call it a slice, as that's what Rust calls it.

expenses: Yeah the terminology is tricky here. It is a slice, but it's specifically a `storage_buffer` accessible as a slice. I guess. You can't create slices from arrays (yet) in rust-gpu and you also can't sub-slice a slice either.

khyperia (approved):

PR #807, opened 2021-11-24
Remove unused attributes from storage class

### LINKED ISSUE #568: Remove unneeded Storage Class attributes
With recent changes to how we specify the storage class of a variable, we've managed to eliminate the need for a few of the storage class attributes in cases where we can successfully infer all possible valid uses. Mainly the `input`, `output`, and `uniform_constant` attributes at the time of this writing are no longer needed and should be removed now that they can't be used for anything useful.

andrewleverette:
resolves #568

### DIFF
--- a/crates/rustc_codegen_spirv/src/symbols.rs
+++ b/crates/rustc_codegen_spirv/src/symbols.rs
@@ -138,15 +138,9 @@ const BUILTINS: &[(&str, BuiltIn)] = {
 const STORAGE_CLASSES: &[(&str, StorageClass)] = {
     use StorageClass::*;
     &[
-        ("uniform_constant", UniformConstant),
-        ("input", Input),
         ("uniform", Uniform),
-        ("output", Output),
         ("workgroup", Workgroup),
         ("cross_workgroup", CrossWorkgroup),
-        ("private", Private),
-        ("function", Function),
-        ("generic", Generic),
         ("push_constant", PushConstant),
         ("atomic_counter", AtomicCounter),
         ("image", Image),


khyperia (changes requested):
I believe a few error messages and tests need to be updated with the removal of uniform_constant

andrewleverette:
Sorry for the late reply, I'm just now catching up after the holiday. I ran the test suite before I submitted the pull request, but I've only just realized the CI test stage is running tests differently. I'm new to the project so I'm not sure what needs to be changed. If you could clarify or provide some context for the necessary changes, I would definitely appreciate it. I'm still happy to work on this task.

expenses:
> Sorry for the late reply, I'm just now catching up after the holiday. I ran the test suite before I submitted the pull request, but I've only just realized the CI test stage is running tests differently. I'm new to the project so I'm not sure what needs to be changed. If you could clarify or provide some context for the necessary changes, I would definitely appreciate it. I'm still happy to work on this task.

As you removed all the attributes that are used in the test: https://github.com/EmbarkStudios/rust-gpu/blob/main/tests/ui/spirv-attr/invalid-storage-class.rs, the failure message of the test is now different.

You could either run the test command `cargo run -p compiletests --release --no-default-features --features "use-installed-tools" -- --target-env vulkan1.1,vulkan1.2,spv1.3` with the `--bless` flag, or probably just remove the test as it doesn't make sense anymore.

andrewleverette:
I removed the test case that you mentioned. I did run the command with the `--bless'` to retest but it looks like it changed several files. Let me know if this should be reverted and just delete the test case.

tests/ui/arch/all_memory_barrier.stderr
```diff
@@ -1,7 +1,14 @@
+error: error:0:0 - failed spawn executable: No such file or directory (os error 2)
```
expenses, line 1: We don't want this error to be here :P Do you have `spirv-val` installed and in your $PATH?

khyperia:
Additionally, the errors themselves needs to be updated, as they can no longer happen.

PR #809, opened 2021-11-26
Generate bools as bools instead of u8

khyperia:
As lengthily discussed in discord.

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -782,22 +782,24 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
         result
     }
 
+    // rustc has the concept of an immediate vs. memory type - bools are compiled to LLVM bools as
+    // immediates, but if they're behind a pointer, they're compiled to u8. The reason for this is
+    // because LLVM is bad at bools behind pointers (something something u1 bitmasking on load).
+    //
+    // SPIR-V allows bools behind *some* pointers, and disallows others - specifically, it allows
+    // bools behind the storage classes Workgroup, CrossWorkgroup, Private, Function, Input, and
+    // Output. In other words, "stuff the CPU can't see, bools are OK, stuff the CPU can't see, no
+    // bools". So, we always compile bools to bools, even if they're behind a pointer, and error if
+    // bools are in an interface (the user should choose u8, u32, or something else instead). That
+    // means that immediate types and memory types are the same, and no conversion needs to happen
+    // here.
     fn from_immediate(&mut self, val: Self::Value) -> Self::Value {
-        if self.lookup_type(val.ty) == SpirvType::Bool {
-            let i8 = SpirvType::Integer(8, false).def(self.span(), self);
-            self.zext(val, i8)
-        } else {
-            val
-        }
+        val
     }
 
     // silly clippy, we can't rename this!
     #[allow(clippy::wrong_self_convention)]
-    fn to_immediate_scalar(&mut self, val: Self::Value, scalar: Scalar) -> Self::Value {
-        if scalar.is_bool() {
-            let bool = SpirvType::Bool.def(self.span(), self);
-            return self.trunc(val, bool);
-        }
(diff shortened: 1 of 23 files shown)


crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -782,22 +782,24 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
...
     }
 
+    // rustc has the concept of an immediate vs. memory type - bools are compiled to LLVM bools as
+    // immediates, but if they're behind a pointer, they're compiled to u8. The reason for this is
+    // because LLVM is bad at bools behind pointers (something something u1 bitmasking on load).
+    //
+    // SPIR-V allows bools behind *some* pointers, and disallows others - specifically, it allows
+    // bools behind the storage classes Workgroup, CrossWorkgroup, Private, Function, Input, and
+    // Output. In other words, "stuff the CPU can't see, bools are OK, stuff the CPU can't see, no
+    // bools". So, we always compile bools to bools, even if they're behind a pointer, and error if
```
expenses, line 792: Should this be:

```suggestion
    // Output. In other words, "stuff the CPU can't see, bools are OK, stuff the CPU can see, no
    // bools". So, we always compile bools to bools, even if they're behind a pointer, and error if
```

?

expenses: The sentence doesn't quite parse for me

crates/spirv-std/shared/src/image_params.rs
```diff
@@ -126,7 +123,7 @@ pub enum Sampled {
     No = 2,
 }
 
-#[cfg(any(not(target_arch = "spirv"), target_feature = "Int8"))]
+#[cfg(not(target_arch = "spirv"))]
```
DJMcNab, line 126: I don't think this needs a `cfg` at all, like the others

DJMcNab (approved):
Haven't fully checked the code, but it passes a visual inspection, and this seems much nicer.

crates/spirv-std/shared/src/image_params.rs
khyperia: No, it needs a `cfg`.

DJMcNab: Oh yeah, sorry

My eyes glazed over whilst reading it
I see now that the cfg was saying that it's only GPU compatible.

tests/ui/lang/core/ops/range-contains.rs
```diff
@@ -11,6 +11,6 @@ fn has_two_decimal_digits(x: u32) -> bool {
 }
 
 #[spirv(fragment)]
-pub fn main(i: u32, o: &mut bool) {
-    *o = has_two_decimal_digits(i);
+pub fn main(i: u32, o: &mut u32) {
+    *o = if has_two_decimal_digits(i) { 1 } else { 0 };
```
eddyb, line 15: Hmm I wonder if `as u32` works.

crates/spirv-std/shared/src/image_params.rs
eddyb: > but now is not usable on GPU at all.

Correct, `Option<bool>` breaks with this PR, because it expects `bool` to be an `u8` - even if we forcefully disable the niche, it still turns the `bool` into `u8` as part of a scalar pair (the same `(is_some: bool, MaybeUninit<T>)`-like scalar pair representation of `Option<T>` which has allowed `for i in 0..n` to work, so we can't just mess with it in general, yet).

eddyb (approved):

tests/ui/lang/core/ops/range-contains.rs
khyperia: oh, cursed, didn't know that existed! yes it does work.

crates/spirv-std/shared/src/image_params.rs
ElectronicRU: Hm - with the bitcast smarts I introduced in #819 this turned out to be rather easy to amend -- will probably add that to the PR, too.

PR #811, opened 2021-11-29
Refactoring & improvements to algorithmic complexity of inline.rs.

ElectronicRU:
inline.rs was a sore spot for me reading the codebase because of how many TODOs and quadratic algorithms it had sitting there.

Looking at it, I was able to see some nicer algorithms trying to poke out. Also a little bit more uniformity for doing the same stuff (like finding variable split points), less cloning, and more caching :)

Tests aren't affected at all, which I'd consider a success.

P.S. Idle hands shave yaks, or so they say.


ElectronicRU:
I've tried to explain what exactly I am doing in each commit, and the rationale behind it. This is a second draft, but if some parts seem unclear or I missed something simple, be sure to tell.

crates/rustc_codegen_spirv/src/linker/inline.rs
```diff
@@ -13,124 +13,158 @@ use rustc_data_structures::fx::{FxHashMap, FxHashSet};
...
 
 pub fn inline(sess: &Session, module: &mut Module) -> super::Result<()> {
+    let (disallowed_argument_types, disallowed_return_types) =
+        compute_disallowed_argument_and_return_types(module);
+    let mut to_delete: Vec<_> = module
+        .functions
+        .iter()
+        .map(|f| should_inline(&disallowed_argument_types, &disallowed_return_types, f))
+        .collect();
     // This algorithm gets real sad if there's recursion - but, good news, SPIR-V bans recursion
```
khyperia, line 26: Making this comment a bit more explicit instead of referring to the extremely tiny and easy to miss `?` would be nice.

crates/rustc_codegen_spirv/src/linker/inline.rs
```diff
@@ -13,124 +13,158 @@ use rustc_data_structures::fx::{FxHashMap, FxHashSet};
...
-    if module_has_recursion(sess, module) {
-        return Err(rustc_errors::ErrorReported);
-    }
+    let postorder = compute_function_postorder(sess, module, &mut to_delete)?;
     let functions = module
```
khyperia, line 28: Renaming this would be nice, it's no longer a map to functions, but rather to their indices.

crates/rustc_codegen_spirv/src/linker/inline.rs
```diff
@@ -364,27 +386,28 @@ impl Inliner<'_, '_> {
...
-        let num_variables = callee_header
-            .iter()
-            .position(|inst| inst.class.opcode != Op::Variable)
-            .unwrap_or(callee_header.len());
-        caller.blocks[block_idx]
-            .instructions
-            .append(&mut callee_header.split_off(num_variables));
+        let num_variables = callee_header.partition_point(|inst| inst.class.opcode == Op::Variable);
+        // Rather than fuse blocks, generate a new jump here. Branch fusing will take care of
+        // it, and we maintain the invariant that current block has finished processing.
```
khyperia, line 394: Clarifying this comment would be nice - "Rather than fuse blocks" is in reference to the old state of the code, and if someone doesn't know the old state, it's kind of vague. Maybe something like "Rather than fuse the first block of the inlined function into the current block of the caller, ..." or something like that.

crates/rustc_codegen_spirv/src/linker/inline.rs
```diff
@@ -260,42 +297,27 @@ impl Inliner<'_, '_> {
...
 
-    fn inline_block(&mut self, caller: &mut Function, block_idx: usize) -> bool {
+    /// Inlines one block and returns whether inlining actually occurred.
+    /// After calling this, blocks[block_idx] is finished processing.
+    fn inline_block(
+        &mut self,
+        caller: &mut Function,
+        functions: &[Function],
+        block_idx: usize,
+    ) -> bool {
```
khyperia, line 320: Does this still need to return a bool?

crates/rustc_codegen_spirv/src/linker/inline.rs
```diff
@@ -394,10 +417,17 @@ impl Inliner<'_, '_> {
             post_call_block_insts.insert(
                 0,
                 Instruction::new(
-                    Op::Load,
+                    Op::Phi,
```
khyperia, line 420: The structurizer does not support phis, and so this cannot be done. mem2reg should take care of the load/store.

crates/rustc_codegen_spirv/src/linker/inline.rs
```diff
@@ -343,18 +360,23 @@ impl Inliner<'_, '_> {
...
-            None
-        };
         let return_jump = self.id();
         // Rewrite OpReturns of the callee.
-        let mut inlined_blocks = get_inlined_blocks(callee, return_variable, return_jump);
+        let (mut inlined_blocks, phi_pairs) = get_inlined_blocks(callee, return_jump);
         // Clone the IDs of the callee, because otherwise they'd be defined multiple times if the
         // fn is inlined multiple times.
         self.add_clone_id_rules(&mut rewrite_rules, &inlined_blocks);
+        // If any of the OpReturns were invalid, return will also be invalid.
```
khyperia, line 369: Please add tests around cases like this, I don't think we have any IIRC.

khyperia (changes requested):
Thanks! This is a whole heck of a lot, and my burnt-out brainthinker can barely cope with reviewing it, but hopefully we can get this through. (I'd definitely like @eddyb to take a look here, too)

khyperia (approved):
LGTM, just want to wait a bit to see if eddyb wants to review!

ElectronicRU:
Updated the MR and resolved conflicts, @eddyb do you wish to take a look?

PR #817, opened 2021-12-04
Use the safe shader initialisation path of wgpu

DJMcNab:
`wgpu` can handle our example shaders safely now, except for the mouse shader, which needed a small change (e04d2d9a0996c5f97491c0b393645c4f05f5452d) to work around naga (?) not handling for loops properly.

This needs further investigation.

### DIFF
--- a/examples/runners/wgpu/src/compute.rs
+++ b/examples/runners/wgpu/src/compute.rs
@@ -22,7 +22,7 @@ pub fn start(options: &Options) {
 
 pub async fn start_internal(
     _options: &Options,
-    shader_binary: wgpu::ShaderModuleDescriptorSpirV<'static>,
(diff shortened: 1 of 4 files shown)


khyperia:
The reason we're not using naga today is because we've had conversations with the authors, and come to the mutual conclusion that, today, naga is not aimed towards the use case of translating built rust-gpu SPIR-V shaders (they are, quite understandably, focused on WGSL/etc.). As naga has been in various states of brokenness for rust-gpu shaders, unbreaking/rebreaking over time, we don't want to block ourselves on the fluctuating state of naga. That might change in the future as naga develops, but that's a future conversation to have. So, I'd like to keep our wgpu example bypassing naga for now.

Issue #78, opened 2020-10-21
Rust Language Support


XAMPPRocky:
This is a tracking issue for adding support for using Rust language features and the `core` library with `rustc_codegen_spirv`.

## Language Support
- [x] Closures
- [ ] Enums
  - [x] Data-less enums. `enum Foo { A, B }`
  - [ ] Algebraic data types.  `enum Option<T> { Some(T), None }`
- [x] Loops
  - [x] `loop`
  - [x] `while`
  - [x] `for`
- [x] `if`
- [x] `match`
- [x] [Proc macros]
- [ ] Try (`?`) operator
- [ ] [128 bit integers](https://github.com/EmbarkStudios/rust-gpu/issues/196)
- [ ] [`union`](https://github.com/EmbarkStudios/rust-gpu/issues/415)
- [ ] Compiles on stable
  - Blocked on stabilisation of `-Z codegen-backend`, also see https://github.com/rust-lang/rust/issues/77933

[proc macros]: https://github.com/EmbarkStudios/rust-gpu/issues/192

## Compiler & Library Support
- [ ] Debug builds
  - Blocked on #81, #186 
- [ ] Iterators
- [ ] [`Option<T>`](https://github.com/EmbarkStudios/rust-gpu/issues/234)

## Features that don't work
- Dynamic Dispatch
- Pointer casts
- [Unsized slices](https://github.com/EmbarkStudios/rust-gpu/issues/378#issue-786964025) (`&[T]`)

## About Tracking Issues
Tracking issues are meant for providing status updates on the progress of a feature, it's not meant for bug reports or discussions on a feature. Please instead file a new issue or join the `#rust-gpu` channel on Discord.

jarble:
Will closures be supported as well?

XAMPPRocky:
@jarble Closures are currently supported.

eddyb:
Should we update `README.md` and/or `docs` to indicate that e.g. `for` loops are now supported?

mitchmindtree:
Hi folks! Is it possible that rust-gpu may one day support trait objects? I'm currently implementing ray tracing in one weekend with rust-gpu and there are a couple cases like `Hittable` and `Material`

khyperia:
> Hi folks! Is it possible that rust-gpu may one day support trait objects?

[...]

PR #819, opened 2021-12-07
Enums MVP.

ElectronicRU:
Supported cases:
- non-repr-C enums without non-null optimization;
- repr-C enums that are dataless / have the same layout for every case
- enums with non-null optimization that don't use pointers (e.g. Option<NonZeroUsize>).

Future work directions:
- Option<&_> and such - currently emits zombies/invalid SPIR-V, likely requires adaptations in mem2reg
- repr(C) enums that can be represented as (discriminant, union) - some of the groundwork is already there


joeftiger:
Would this mean that one can use things like `Option<T>` or alge raic enums, e.g.? 
As long as `type` is not a pointer?

I didn't get to play around with `rust-gpu` that much yet, my understanding was that so far only data-less enums work (#78).

ElectronicRU:
@joeftiger yes. With pointers, it can generate incorrect SPIR-V, because operations on pointers are heavily restricted and amending that would require basically a full-program analysis that I haven't come up with.

Keavon:
This has been open for nearly a year. Is it still ready and just waiting for a review?

djdisodo:
i've roughly merged this into latest version [here](https://github.com/djdisodo/rust-gpu/tree/c04bacf4df57c21a493a12844ebaebf76916e15a)

i compiled this code
```rust
struct Nah {
    a: u32
}

enum Foo {
    A(Nah),
    B,
}

#[spirv(compute(threads(64)))]
pub fn main(
    #[spirv(global_invocation_id)] id: UVec3,
    #[spirv(storage_buffer, descriptor_set = 0, binding = 1)] output: &mut [u32],
) {
    let aa = Foo::A(Nah { a: 0xffffffff });

    if let Foo::A(a) = aa {
        output[id.x as usize] = a.a;
    } else {
        output[id.x as usize] = 99;
    }
}
```
it fails to compile

however if i revert `providers.layout_of` changes it works

need further research

ElectronicRU:
@djdisodo do you mean the `deoverlay_enum_fields` change etc?
@Keavon there were some unresolved issues with this code, and ultimately, not much desire from Embark to do this. No-one wanted to do a massive code review, I guess.

@eddyb there seems to still be interest in this feature - I can probably bring it up to date, write some more tests, etc. Unfortunately I never did summon the courage required to make it work with pointers.

Keavon:
@ElectronicRU that would be awesome if you could return to this and get it landed now that the project seems to be picking back up on Embark's end. Our project https://github.com/GraphiteEditor/Graphite needs it for our node graph compositing shaders, which are written in Rust, to work with (some degree of) generic types in our node graph system.

brynnbrancamp:
This is fundamental Rust, we need some form of this feature; if not, this.

Lucky4Luuk:
What is the status on this feature? I'm running into issues with Option that I feel are related to this, specifically with Option<T> where T is a struct.

ElectronicRU:
@Lucky4Luuk sorry - life & work take up a lot of time, I don't have any to work on this and won't for the foreseeable future :( Someone else would need to push it through for now, sorry.

reinismu:
Noticed rust-gpu project and got so excited today, but after few hours encountered unsupported `Option<T>`. Gave it a try to merge this PR in main, but a lot of things have changed and I don't have the know how :/

Would really appropriate if someone picked this up again :pray:

ElectronicRU:
@reinismu could be easier to re-do at this point :) btw the repository has moved.

PR #822, opened 2021-12-09
Add `RayQuery::confirm_intersection`

expenses:
I spent like 20 minutes looking for this one before realising that it wasn't implemented :sweat_smile:

You need this if the ray query flags do not contain `RayFlags::OPAQUE`, as otherwise a `while ray.proceed() {}` will just skip through all the candidate intersections without confirming any of them. Essentially, you need it for alpha clipping and stuff like that. See here for an example: https://github.com/expenses/transmission-renderer/blob/c5dbd648e513def2719f3971fd5d9c8ce992ba63/shader/src/lib.rs#L736-L766

### DIFF
--- a/crates/spirv-std/src/ray_tracing.rs
+++ b/crates/spirv-std/src/ray_tracing.rs
@@ -288,14 +288,28 @@ impl RayQuery {
     }
 
     /// Terminates further execution of a ray query; further calls to
-    /// [`Self::proceed`] will return `false`.The value returned by any prior
+    /// [`Self::proceed`] will return `false`. The value returned by any prior
     /// execution of [`Self::proceed`] with the same ray query object must have
-    /// been true. Refer to the client API specification for more details.
+    /// been true.
     #[spirv_std_macros::gpu_only]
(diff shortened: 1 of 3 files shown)


crates/spirv-std/src/ray_tracing.rs
```diff
@@ -288,16 +288,30 @@ impl RayQuery {
     }
 
     /// Terminates further execution of a ray query; further calls to
-    /// [`Self::proceed`] will return `false`.The value returned by any prior
+    /// [`Self::proceed`] will return `false`. The value returned by any prior
     /// execution of [`Self::proceed`] with the same ray query object must have
-    /// been true. Refer to the client API specification for more details.
```
expenses, line 293: I removed this like just because it wasn't consistent with the rest of the docs and seems to have just been copied from https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_ray_query.asciidoc.

khyperia (approved):

PR #824, opened 2021-12-13
A whole host of minor tweaks and clippy fixes

khyperia:


termhn (approved):

crates/rustc_codegen_spirv/src/builder/byte_addressable_buffer.rs
```diff
@@ -43,7 +42,7 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
                 u32_ptr,
                 None,
                 array.def(self),
-                IntoIter::new([actual_index.def(self)]),
+                IntoIterator::into_iter([actual_index.def(self)]),
```
eddyb, line 45: `in_bounds_access_chain` takes `impl IntoIterator`, so I think this can just pass the array now.

crates/rustc_codegen_spirv/src/builder/byte_addressable_buffer.rs
```diff
@@ -222,7 +221,7 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
```
eddyb, line 224: `in_bounds_access_chain` takes `impl IntoIterator`, so I think this can just pass the array now.

eddyb (changes requested):

eddyb:
The test failure looks like just needs a `--bless` (line numbers, in some diagnostic, changed).

examples/runners/wgpu/src/compute.rs
```diff
@@ -61,7 +61,7 @@ pub async fn start_internal(
         .clone()
         // Not sure which endianness is correct to use here
         .map(u32::to_ne_bytes)
-        .flat_map(core::array::IntoIter::new)
+        .flat_map(IntoIterator::into_iter)
```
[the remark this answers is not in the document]
khyperia: Replaced with `.map(u32::to_ne_bytes).flatten()`, clippy suggested `.flat_map(u32::to_ne_bytes)`, sure I guess!

eddyb (approved):

PR #826, opened 2021-12-15
Fix some ICEs with bools

khyperia:
https://github.com/EmbarkStudios/rust-gpu/pull/809 missed some cases where `to/from_immediate` was inlined into local code, causing some ICEs

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -927,20 +927,16 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
             let b_offset = a.value.size(self).align_to(b.value.align(self).abi);
 
             let pair_ty = place.layout.spirv_type(self.span(), self);
-            let mut load = |i, scalar: &Scalar, align| {
+            let mut load = |i, _: &Scalar, align| {
                 let llptr = self.struct_gep(pair_ty, place.llval, i as u64);
-                let load = self.load(
+                // WARN! This should go through to_immediate, but because we only have a Scalar,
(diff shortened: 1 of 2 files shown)


crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -927,20 +927,16 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
             let b_offset = a.value.size(self).align_to(b.value.align(self).abi);
 
             let pair_ty = place.layout.spirv_type(self.span(), self);
-            let mut load = |i, scalar: &Scalar, align| {
+            let mut load = |i, _: &Scalar, align| {
                 let llptr = self.struct_gep(pair_ty, place.llval, i as u64);
-                let load = self.load(
+                // WARN! This should go through to_immediate, but because we only have a Scalar,
+                // not a Ty, we can't call to_immediate here (even though we implement it as a
+                // no-op), so "inline" the no-op here and do nothing.
```
eddyb, line 934: Isn't that why `to_immediate_scalar` exists, and `to_immediate` merely proxies to it? `rustc_codegen_ssa` calls `to_immediate_scalar` in a few places where `to_immediate` isn't available (e.g. for each half of a `ScalarPair`).

eddyb (changes requested):

crates/rustc_codegen_spirv/src/builder/builder_methods.rs
khyperia: ahah, yes, I totally remembered `to_immediate_scalar` existed, mhmm, definitely. I'm going to go with the excuse that "I don't think it existed when I originally wrote this code, and I just copied the comment/behavior forward with this new change" and hope that's right :sparkles:

eddyb (approved):

PR #829, opened 2021-12-16
Update deps

khyperia:
there's an angey yellow badge on the readme frontpage, so update stuff!

(although sadly, it says `nix` is out of date, but we can't update nor remove that dependency because it [fixes a wild issue](https://github.com/EmbarkStudios/rust-gpu/pull/765#discussion_r727076669) with cargo update)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -174,9 +174,9 @@ checksum = "8f1e260c3a9040a7c19a12468758f4c16f31a81a1fe087482be9570ec864bb6c"
 
 [[package]]
 name = "bytemuck"
-version = "1.7.2"
+version = "1.7.3"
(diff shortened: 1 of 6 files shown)


termhn (approved):
Not sure if the glam issue you mentioned affects this but if not then 👍

khyperia:
Yep, that's the CI failure :c - might revert just the glam update here and get everything else in, we'll see how my investigation tomorrow goes and if it's a relatively easy fix or not.

eddyb (approved):

khyperia:
Yeah, reverting the glam update since the fix won't potentially land until I'm on vacation, and I'd like to make the process of updating glam here as easy as possible if someone other than me wants to work on it over vacation (although I guess I'm only gone a week). So, this should pass and merge all the other pending updates.

Issue #827, opened 2021-12-15
DCE shouldn't remove unused descriptors


expenses:
I'm performing reflection on generated SPIR-V modules to create descriptor set layouts. Unfortunately, this is a little fiddly, because whenever I comment out some code to debug something, if it means that a descriptor is no longer used, this descriptor will not be output in the final module and the generated descriptor set layouts will be wrong.

In my opinion, this sort of Dead Code Elimination should not be performed, especially with `SpirvMetadata::NameVariables` or `NameVariables::Full`. Given that `SpirvMetadata::NameVariables` has the comment 'Only include OpNames for public interface variables (uniforms and the like), to allow shader reflection.', I think we can say that this DCE is unwanted behaviour.

khyperia:
We already looked into this quite a bit in https://github.com/EmbarkStudios/rust-gpu/issues/722, and tl;dr, we cannot fix this for SPIR-V versions <=1.3. Use >=1.4 if keeping unused entry parameters is needed.

expenses:
Oh, nice! That should probably be documented somewhere.

expenses:
Actually the DCE still seems to happen even when building for `spirv-unknown-spv1.5`, at least on https://github.com/EmbarkStudios/rust-gpu/commit/da7c3ed71f27d67e05c0d2a4d3c839db14daf1f8.

khyperia:
Ah, that would be a `spirv-opt` issue then, it's removing it - this doesn't include `samp`, but removing the `spirv-opt` invocation makes it appear: https://shader-playground.timjones.io/05a24c0565ef5fd9b8dc51457c8ad135

expenses:
Ah, right! Looks like `spirv_tools::opt::Options::preserve_bindings` should be set to true then.

PR #830, opened 2021-12-20
Add a preserve_bindings opt to the builder and codegen opts

expenses:
~~This resolves https://github.com/EmbarkStudios/rust-gpu/issues/827 and makes reflection a lot easier. I think having it set to true unconditionally is the right move but two alternatives are:~~

~~* Setting it to `cg_args.spirv_metadata != SpirvMetadata::None`. This is pretty unclear behaviour but it does keep things the same in the default case.~~
* Adding a new `CodegenArgs::opt_preserve_bindings` argument. Adds more complexity and is easily missed. Doesn't really 'just work'.

### DIFF
--- a/crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/mod.rs
@@ -251,6 +251,9 @@ pub struct CodegenArgs {
     pub uniform_buffer_standard_layout: bool,
     pub scalar_block_layout: bool,
     pub skip_block_layout: bool,
+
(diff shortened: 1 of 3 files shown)


khyperia:
I think the best option here is the second alternative: adding a new option to spirv-builder. Removing unused bindings is the default of other shader compilers, as well as a lot of tooling, and is usually the desired behavior AFAIK. Adding an option for alternate use cases seems like the way to go.

expenses:
@khyperia done.

khyperia (approved):
thanks!

PR #831, opened 2021-12-20
Update to glam 0.20

emilk:
When it comes to dependencies, it takes all the running you can do to keep in the same place.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,9 +39,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.51"
+version = "1.0.52"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8b26702f315f53b6071259e15dd9d64528213b44d61de1ec926eca7715d62203"
+checksum = "84450d0b4a8bd1ba4144ce8ce718fbc5d071358b1e5384bace6536b3d1f2d5b3"
 
 [[package]]
 name = "ar"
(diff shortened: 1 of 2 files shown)


hrydgard:
glam is broken for rust-gpu in the 0.20.x releases before 0.20.2, not sure if worth trying to specify the range more exactly?

repi:
Oh good catch, then specifying the exact minimum patch version should solve that.  And it will only allow version of that or above. hopefully works out

emilk:
So ideally we would allow all version in the `0.17`-`0.19` range PLUS `0.20.2`, but I don't think cargo supports specifying multiple allowed ranges like that: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

The alternative is changing it to just `0.20.2`, but that means dropping support for glam `0.17`-`0.19`.

repi:
Oh ouch, that does indeed sound like a limitation of cargos version comparison, and a fair workaround with the patch version.

hrydgard (approved):

khyperia (approved):
Yay! CI tests with `Cargo.lock`, which was still at 0.19 - so I quick pushed a `cargo update` to make sure everything's tidy with 0.20.

khyperia:
ugh, `cargo fetch --locked` stalled for an hour and a half in the lint CI step, I cancelled that but github has no option for "just rerun these things" and had to rerun everything. Now, the ubuntu CI step is also stalled for an hour and a half on `cargo fetch --locked`. I'm just going to force-merge because everything has been green, just not all on the same CI run.

PR #835, opened 2021-12-27
Fuse OpPtrAccessChain with previous OpAccessChain

### LINKED ISSUE #834: Support fixed size array on stack
Code like `let mut block_indexs = [0usize; LEVEL_COUNT];` generate `Cannot offset a pointer to an arbitrary element`. 

According to the conversation here https://discord.com/channels/750717012564770887/750717499737243679/924984414432264214 , SPIR-V support fixed size array on stack, so it can be supported.

khyperia:
Fixes #834

The issue was that memset was getting called with a pointer to an element, rather than pointer to array. That pointer was presumably obtained through a pointercast implemented via "oh that's a field, lets get a pointer to the field", i.e. OpAccessChain. That means when the pointer to the element was offset, it generated code that looks like:

```
%element = OpAccessChain %arr [%const_0]
%result = OpPtrAccessChain %element %offset []
```

this is silly and dumb, we should recognize this as

```
%result = OpAccessChain %arr [%offset]
```

so, do so.

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/mod.rs
+++ b/crates/rustc_codegen_spirv/src/builder/mod.rs
@@ -13,7 +13,7 @@ use crate::abi::ConvSpirvType;
 use crate::builder_spirv::{BuilderCursor, SpirvValue, SpirvValueExt};
 use crate::codegen_cx::CodegenCx;
 use crate::spirv_type::SpirvType;
-use rspirv::spirv::Word;
+use rspirv::spirv::{self, Word};
 use rustc_codegen_ssa::mir::operand::OperandValue;
 use rustc_codegen_ssa::mir::place::PlaceRef;
 use rustc_codegen_ssa::traits::{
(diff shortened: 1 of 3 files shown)


crates/rustc_codegen_spirv/src/builder/mod.rs
```diff
@@ -98,8 +98,29 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
         self.current_span.unwrap_or(DUMMY_SP)
     }
 
+    // Given an ID, check if it's defined by an OpAccessChain, and if it is, return its ptr/indices
+    fn find_access_chain(&self, id: spirv::Word) -> Option<(spirv::Word, Vec<spirv::Word>)> {
+        let emit = self.emit();
+        let module = emit.module_ref();
+        let func = &module.functions[emit.selected_function().unwrap()];
+        let ptr_def_inst = func.all_inst_iter().find(|inst| inst.result_id == Some(id));
```
eddyb, line 106: Nice usecase for a def-use graph during codegen itself, heh. In my WIP stuff it would look like this:
```rust
fn find_access_chain(&self, ptr_value: Value) -> Option<(Value, &[Value])> {
    if let Value::DataInstOutput(inst) = ptr_value {
        let ptr_def_inst = &self.func.data_insts[inst];
        if ptr_def_inst.kind == ... {
            return Some((ptr_def_inst.inputs[0], &ptr_def_inst.inputs[1..]));
        }
    }
    None
}
```

eddyb (approved):

crates/rustc_codegen_spirv/src/builder/mod.rs
```diff
@@ -134,42 +155,58 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
...
+            // Transform the following:
+            // OpAccessChain original_ptr [a, b, c]
+            // OpPtrAccessChain ptr base [d, e, f]
+            // into
+            // OpAccessChain original_ptr [a, b, c + base, d, e, f]
+            // to remove the need for OpPtrAccessChain
+            let last = original_indices.last_mut().unwrap();
+            *last = self
+                .add(last.with_type(indices[0].ty), indices[0])
+                .def(self);
```
eddyb, line 170: Just looked at the CI failure, this doesn't do constant-folding, you may need to change this other thing:
https://github.com/EmbarkStudios/rust-gpu/blob/3067848d1c7df819a0036eff4ee1ad60a4a3bc0b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs#L644-L657

Something like this:
```rust
simple_op! {
    add, i_add,
    fold_const { 
        int(a, b) => a + b
    }
}
```

However, I think the existing one should be `.wrapping_mul` and this one should be `.wrapping_add`, not `*` and `+` (either that or [the `fold_const` handling](https://github.com/EmbarkStudios/rust-gpu/blob/3067848d1c7df819a0036eff4ee1ad60a4a3bc0b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs#L51-L55) should be using `Wrapping` around the `u128` values).

khyperia: Ah yep! Was curious about what rust's behavior here is, looks like there's a deny-by-default warning ([that people want to make into an error](https://github.com/rust-lang/rust/issues/71800), but looks like that missed the 2021 train) for constant overflow - mmmaybe we want to also error instead of overflow, but who cares right now tbh :P

PR #837, opened 2022-01-07
Fix ByteAddressableBuffer PassMode::Pair

khyperia:


tests/ui/byte_addressable_buffer/empty_struct.rs
```diff
@@ -0,0 +1,25 @@
...
+pub struct EmptyStruct {}
+
+#[spirv(fragment)]
+pub fn load(
+    #[spirv(descriptor_set = 0, binding = 0, storage_buffer)] buf: &mut [u32],
+    #[spirv(flat)] out: &mut EmptyStruct,
+) {
+    unsafe {
+        let buf = ByteAddressableBuffer::new(buf);
+        *out = buf.load(5);
```
hrydgard, line 14: 5? Aren't only 4-byte-aligned addresses valid to use in these?

khyperia: in HLSL, it's documented to be UB to not have a 4-byte-aligned number, but in rust-gpu, I figured hey, UB is gross, let's just document to what HLSL compilers actually do, which is round down to the nearest multiple of 4. So, this is equivalent to passing in 4. (This code is never ran, and it doesn't really matter - all the tests in `byte_addressable_buffer` are copypasted from each other, and I just chose a random nonzero number without thinking much)

hrydgard: OK! Makes sense - although that this is the intention should maybe be in a comment somewhere, possibly just copy-pasted from what you just wrote?

eddyb (approved):

PR #844, opened 2022-01-23
abi: explicitly unpack `ScalarPair` newtypes.

### LINKED ISSUE #836: Cannot cast between pointer types
### Expected Behaviour

The main branch doesn't throw this error, but after some refactoring in that branch it throws that error

(the main branch already merged the deref-type branch, but it fixes the problem by duplicating all the code in two build targets)

repo: https://github.com/molikto/graphics-playground/tree/deref-type

run: `cargo run -p bosky --release`

result in rust-gpu error:

```
     --> C:\Users\molikto\GitHub\graphics-playground\crates\common-no-std\src\math\svo.rs:180:36
      |
  180 |                 let target_block = self.mem[target_block_index];
      |                                    ^^^^^^^^
      |
      = note: from: *struct common_no_std::svo::Svo<&[u16], 2_u16, 8_usize> { *[u16], u32 }
      = note: to: *struct &[u16] { *[u16], u32 }
```

## System Info

 - Rust: rustc 1.57.0 (f1edd0429 2021-11-29)
 - OS: Windows 10
 - SPIR-V: SPIRV-Tools v2021.4-dev v2021.3-86-g21e3f681

eddyb:
I've left good description of the new codepath in a comment in the source, reproduced here:

> Unlike `Abi::Scalar`'s simpler newtype-unpacking behavior, `Abi::ScalarPair` can be composed in two ways:
> * two `Abi::Scalar` fields (and any number of ZST fields), gets handled the same as a `struct { a, b }`, further below
> * an `Abi::ScalarPair` field (and any number of ZST fields), which requires more work to allow taking a reference to that field, and there are two potential approaches:
>   1. wrapping that field's SPIR-V type in a single-field `OpTypeStruct` - this has the disadvantage that GEPs would have to inject an extra `0` field index, and other field-related operations would also need additional work
>   2. reusing that field's SPIR-V type, instead of defining a new one, offering the `(a, b)` shape `rustc_codegen_ssa` expects, while letting noop pointercasts access the sole `Abi::ScalarPair` field - this is the approach taken here

The user-facing summary is that `MyNewtype<(u32, u32)>` or `MyNewtype<&[T]>` (given `struct MyNewtype<T>(T);`) used to allow reading the single field `MyNewtype` *by value* (it would read the two scalars separately), but you couldn't *take a reference* to the field, as the types would be different `OpTypeStruct`s (with the same fields).

On the implementation side, the choice was to keep the existing type identities for anything that itself has the two scalars as separate fields, and only reuse SPIR-V types in the "newtype of something that *itself* is `ScalarPair`" case, to avoid accidentally allowing more type conflation than necessary.

Fixes #836.

### DIFF
--- a/crates/rustc_codegen_spirv/src/abi.rs
+++ b/crates/rustc_codegen_spirv/src/abi.rs
@@ -370,6 +370,40 @@ impl<'tcx> ConvSpirvType<'tcx> for TyAndLayout<'tcx> {
             .def_with_name(cx, span, TyLayoutNameKey::from(*self)),
             Abi::Scalar(ref scalar) => trans_scalar(cx, span, *self, scalar, Size::ZERO),
             Abi::ScalarPair(ref a, ref b) => {
+                // NOTE(eddyb) unlike `Abi::Scalar`'s simpler newtype-unpacking
+                // behavior, `Abi::ScalarPair` can be composed in two ways:
+                // * two `Abi::Scalar` fields (and any number of ZST fields),
+                //   gets handled the same as a `struct { a, b }`, further below
+                // * an `Abi::ScalarPair` field (and any number of ZST fields),
+                //   which requires more work to allow taking a reference to
+                //   that field, and there are two potential approaches:
+                //   1. wrapping that field's SPIR-V type in a single-field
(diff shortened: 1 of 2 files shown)


hrydgard (approved):
Looks good!

PR #845, opened 2022-01-29
in multimodule mode, rewrite entry name to "main"

molikto:
Feel free to close it if this is not desired.

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/mod.rs
+++ b/crates/rustc_codegen_spirv/src/linker/mod.rs
@@ -309,7 +309,9 @@ pub fn link(sess: &Session, mut inputs: Vec<Module>, opts: &Options) -> Result<L
             .map(|entry| {
                 let mut module = output.clone();
                 module.entry_points.clear();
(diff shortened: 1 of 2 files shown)


expenses:
This definitely can make loading modules earlier. It might be better as a config option though?

hrydgard:
I see how this can make sense. But yeah, it does need to be an option, so please make it one and we'll be open to merging this.

molikto:
Personally, I don't want to make it an option because it seems to be feature bloat to me.

I have the assumption that multi-module feature is just for compatibility reasons, so the entry rewrite seems always desired (or at least I don't see when it causes inconvenience). But I am just very new to graphics programming, my assumption might be wrong.

hrydgard:
Your assumption is unfortunately wrong.. It turns out that multimodule is pretty much required today, as there are too many ecosystem bugs with keeping multiple shaders in a single SPIR-V binary. But we still like to use more specific entry point names than `main`. So merging this as-is would break our code, which we are not willing to do.

Issue #841, opened 2022-01-14
stop relying on spirv-tools


dvc94ch:
currently the spirv spec says that subgroups are restricted to kernels. however glsl supports subgroups in regular compute shaders. seems like the glsl spec team forgot to tell the spirv spec team to update their specs. this leads to spirv-tools complaining about an invalid decorator in the opt/validation phases which in turn leads to rust-gpu not being competitive with glsl for no real reason. removing the dependency on spirv-tools would allow supporting "defacto" standards instead of just what was bothered to be written down in a spec.

- [0] https://github.com/KhronosGroup/GLSL/blob/master/extensions/khr/GL_KHR_shader_subgroup.txt

khyperia:
> currently the spirv spec says that subgroups are restricted to kernels

can you show where it says that? not seeing where it says that.

dvc94ch:
section 3.21 builtin says that kernel, groupnonuniform and sugroupballotkhr are enabling capabilities for `subgroup_size`, `num_subgroups`, `subgroup_id`, `subgroup_local_invocation_id`, to my understanding implying that they aren't available in the `GLCompute` mode, the `Kernel` mode being something to do with opencl.

khyperia:
That's an `or` between each capability, not an `and` - you don't need `Kernel` to use `SubgroupSize` and friends, just `GroupNonUniform`. They're available in compute shader SPIR-V and rust-gpu, and pass spirv-val.

dvc94ch:
so how do I enable it? I currently get this:

```
error: Operand 3 of Decorate requires one of these capabilities: Kernel GroupNonUniform SubgroupBallotKHR 
         OpDecorate %subgroup_local_invocation_id BuiltIn SubgroupLocalInvocationId
  |
  = note: module `/home/dvc/cloudpeer/xuda/target/shader/target/spirv-unknown-vulkan1.2/release/deps/shader.spv.dir/module`

warning: an unknown error occurred
  |
  = note: spirv-opt failed, leaving as unoptimized
  = note: module `/home/dvc/cloudpeer/xuda/target/shader/target/spirv-unknown-vulkan1.2/release/deps/shader.spv.dir/module`

error: error:0:0 - Operand 3 of Decorate requires one of these capabilities: Kernel GroupNonUniform SubgroupBallotKHR 
         OpDecorate %subgroup_local_invocation_id BuiltIn SubgroupLocalInvocationId
  |
  = note: spirv-val failed
[...]

PR #848, opened 2022-02-07
Enable builtins.

dvc94ch:


eddyb:
Can you try adding tests for these? I thought they couldn't work for now.

dvc94ch:
They seem to work fine. Any reason why you think they shouldn't?

dvc94ch:
Is there an example of an existing builtin test somewhere?

expenses:
See: https://github.com/EmbarkStudios/rust-gpu/pull/789 (I followed the git blame)

> In discord, it was pointed out that `#[spirv(workgroup_size)]` is broken - the decoration applies to a constant, not a variable, and it specifies the workgroup size, it's not provided with the workgroup size. (A subsequent discussion also showed that WorkgroupSize in SPIR-V is just plain broken with multi-entry-point shaders - but the LocalSize(Id) execution mode is a replacement for its behavior)

dvc94ch:
ok `workgroup_size` might be broken, realized I wasn't actually using it. but I think we went over this in https://github.com/EmbarkStudios/rust-gpu/issues/841, `subgroup_local_invocation_id` should be supported as it has the following enabling capabilities: Kernel, GroupNonUniform, SubgroupBallotKHR.

dvc94ch:
Let me know if you want me to drop the example.

eddyb (approved):
[Looking at the Vulkan spec](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#interfaces-builtin-variables-sgli) (since SPIR-V fails to document it), looks like `SubgroupLocalInvocationId` does indeed go on a variable.

(Whereas [for `WorkgroupSize` it mentions "constants"](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#interfaces-builtin-variables-warpidnv) - though technically ambiguous/a misuse of "variable")

PR #852, opened 2022-02-25
Enable Vulkan for WGPU runner

andrusha:
Metal doesn't support SPIR-V shader passthrough, so on macOS using WGPU runner you get following error with any example:

```
Features SPIRV_SHADER_PASSTHROUGH are required but not enabled on the device
```

Vulkan supports shader passthrough, but requires compatibility layer of MoltenVK to work, which needs to be installed on the host machine. Currently this feature is only available on master.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,15 +57,6 @@ checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
 name = "ash"
-version = "0.33.3+1.2.191"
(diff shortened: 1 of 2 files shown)


hrydgard (approved):

termhn (approved):

VZout (approved):

hrydgard:
Would merge this but there are CI failures. Gotta commit the lock file too.

(I can do it for you, but thought I'd ask first).

hrydgard:
The last CI failure is unrelated (due to a vuln caught by cargo-deny) so I'll merge and "fix in post".

Issue #847, opened 2022-02-05
`#[inline(never)]` ignored when using borrowed push constant as argument.


SiebenCorgie:
### Expected Behaviour
The function
```rust
#[inline(never)]
fn sdf(coord: Vec2, con: &PushConst) -> f32{
    let result = (coord - Vec2::splat(200.0)).length() - 150.0;
    result - con.color[0]
}
```

is expected to be emitted as a `NoInline` tagged spirV function, but it gets inlined.

### Fix

This can be fixed by cloning the constant, which results in the function actually being emitted like this:
```
%19 = OpFunction  %25  DontInline %34
%52 = OpFunctionParameter  %32
%53 = OpFunctionParameter  %18
%54 = OpLabel
OpLine %5 30 4	
```

eddyb:
We did land this:
* https://github.com/EmbarkStudios/rust-gpu/pull/875

Is that good enough? Sadly we can't fix the underlying pointer issue without doing what you did automatically. Maybe we will, eventually, but for big enough data we'd need heuristics to avoid accidentally copying kilobytes of buffers pointlessly.

SiebenCorgie:
Yeah that looks like a good fix for now, thanks!

PR #857, opened 2022-04-05
Inline globals

molikto:
* warn on `inline(never)` that are not respected
* inline global variables so that we can inline less functions caused by illegal arg/return type

fix https://github.com/EmbarkStudios/rust-gpu/issues/847

## tests

My shader which compile to a single file of 2500 lines (dis glsl) now respect the `inline(never)` and result in 1400 lines of glsl

TODO: some compiletests produce different result??


eddyb:
> some compiletests produce different result??

For the SPIR-V disasm checks, that makes sense - you're removing some inlining requirements.

<hr/>

I've tried looking at this PR, and while I believe the general idea makes sense (i.e. as a form of "parameter weakening", all call sites agreeing on a quasi-constant value for some parameter does allow removing the parameter from the signature), I can't easily reason about the specifics of the code in this PR at this time.

<hr/>

At the very least, I feel like it should go into the `param_weakening` module:
https://github.com/EmbarkStudios/rust-gpu/blob/5ac500d5b96145d72da6fe2a1059bd4e2a34b5e8/crates/rustc_codegen_spirv/src/linker/param_weakening.rs#L1-L4

And use the `CallGraph` abstraction to correctly visit the module in the right order for the right operations. Ideally we would have a framework for computing general function parameter weakenings from either callee uses (like the existing `remove_unused_params`) or call sites, and correctly propagating them through any number of nested calls.

<hr/>

I'm leaving this comment now as-is because I'm worried I might not be able to meaningfully review this any time soon, or at least not without doing some of the generalization work, to be able to convince myself this transform won't have subtle failure modes.

AdrianEddy:
Any update on this? I have a shader which compiles to 6000 lines of GLSL or WGSL code, while hand-crafted one was only 300 lines, and it would really help to not inline the functions, as my only parameter to the inlined functions is the single uniform struct reference

Issue #858, opened 2022-04-10
Optimize `spirv-builder` clean build wall time


repi:
We should look into improve our clean build times of when building `spriv-builder` & `rustc_codegen_spirv` as that is the main scenario in CI and also useful for users if our crates rebuild fast.

Right now it looks like we have a couple of quite unfortunate serial dependencies that significantly increases the wall time. Here is the current results with `cargo build --release -p spirv-builder -Z timings` on my AMD 5950x (32 vCPU):

![image](https://user-images.githubusercontent.com/1262692/162613624-66cd6fec-6455-4569-976d-33ca9f3ee80c.png)

## Potential optimizations

1. Could `spirv-builder` be the one depending on `spirv-tools` instead of `rustc_codegen_spirv`?
    - I.e. move the responsibility of running spirv-opt/spirv-val and final linking to the builder. 
    - Does that sound like it could be doable @eddyb ?
2. Remove/replace `serde`, quite crazy it takes 20 seconds to compile
    - Think we only have two use cases, our `ModuleResult` output and then the internal decorations that is using more complex serde implementation (`CustomDecoration`).
    - have no idea how `serde`+`serde_json` is taking such a crazy long time to build here though, can't see any other crates enabling tons of types that serde derive macros are used on or similar
    - `ModuleResult` would be trivial to replace serde with `nanoserde`, could work for `CustomDecoration` also?
3. Remove/replace `sanitize-filename` dependency that compiles and uses `regex` crate for super trivial things

If we do get rid of serde and manage to make sure `rustc_codegen_spirv` doesn't have to wait for the `spirv-tools-sys` build, we could get get a 15+ second wall time improvement here as `rustc_codegen_spirv` could start as soon as frontend section of `rspirv` has been built:
      -  ![image](https://user-images.githubusercontent.com/1262692/162613819-e0b53cb2-3cec-4b42-8241-98f00ee40486.png)

repi:
Did an experiment to switch to `nanoserde` and it seems to work fine, there are some contraints and type/format output changes that one have to figure out (no `PathBuf` support), and does get rid of all the long serde compilation, 20 sec -> 1.7 sec.

Wall time benefit was just ~1 second though, but would enable real gains if one also finds a way not have `rustc_codegen_spirv` stall and wait for `spirv-tools-sys` build.

![image](https://user-images.githubusercontent.com/1262692/162616546-87b6103e-4718-43d6-a507-9fa80602e419.png)

[...]

PR #859, opened 2022-04-10
Switch `rustc_codegen_spirv` to static lib

repi:
Was set as a `dylib` but not really sure why. Having it as a static library enables dependencies, like `spriv-builder`, to start building directly after the frontend section of the build of the crate is done. Rather than having to wait for the codegen as well.

Saw clean ´spirv-builder` build times go from 46.3 -> 45.8 seconds with this. Have seen bigger benefits when combined with other optimizations as well (example: https://github.com/EmbarkStudios/rust-gpu/issues/858#issuecomment-1094259795)

Part of #858

### DIFF
--- a/crates/rustc_codegen_spirv/Cargo.toml
+++ b/crates/rustc_codegen_spirv/Cargo.toml
@@ -9,9 +9,6 @@ repository = "https://github.com/EmbarkStudios/rust-gpu"
 #categories = []
 #readme = "README.md"
 
-[lib]
-crate-type = ["dylib"]
-
 [features]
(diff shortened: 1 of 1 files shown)


repi:
Oh interesting CI failure, so we do need this as `dylib` for the tests?

repi:
That must have indeed been the case, as can't reproduce this building without being a `dylib` on a separate machine after properly clean build.

eddyb:
> I did do a `cargo clean` before hand and examples did work

Heh, I don't trust tools to do this properly (for whatever reason), and I `rm -rf target/` instead.

PR #860, opened 2022-04-10
Experiment to switch to `nanoserde` instead of `serde`

repi:
Part of #858

### DIFF
--- a/crates/rustc_codegen_spirv-types/src/compile_result.rs
+++ b/crates/rustc_codegen_spirv-types/src/compile_result.rs
@@ -1,17 +1,15 @@
-use serde::{Deserialize, Serialize};
-use std::collections::BTreeMap;
+use nanoserde::{DeJson, SerJson};
+use std::collections::HashMap;
 use std::fmt::Write;
-use std::path::{Path, PathBuf};
 
-#[derive(Debug, Serialize, Deserialize)]
-#[serde(untagged)]
+#[derive(Debug, DeJson, SerJson)]
 pub enum ModuleResult {
-    SingleModule(PathBuf),
-    MultiModule(BTreeMap<String, PathBuf>),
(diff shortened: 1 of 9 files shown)


crates/rustc_codegen_spirv-types/src/compile_result.rs
```diff
@@ -1,17 +1,15 @@
...
-use std::path::{Path, PathBuf};
 
-#[derive(Debug, Serialize, Deserialize)]
-#[serde(untagged)]
+#[derive(Debug, DeJson, SerJson)]
 pub enum ModuleResult {
-    SingleModule(PathBuf),
-    MultiModule(BTreeMap<String, PathBuf>),
+    SingleModule(String),
+    MultiModule(HashMap<String, String>),
```
Jasper-Bekkers, line 8: Check if using a HashMap here doesn't impact buildtime determinism.

hrydgard: That's a good point. I think it's likely best to just stick to BTreeMap.

repi: `nanoserde` doesn't support `BTreeMap` unfortunately, that's why I changed it. 

but yes logically it should be a `BTreeMap` here. though the builds are rather non-deterministic anyway so probably quite fine. but should comment it

joeftiger: `nanoserde` has a closed issue on this here: https://github.com/not-fl3/nanoserde/issues/7
Apparently the owner of the library prefers to use a proxy in this case

> [...] instead of adding all possible containers into the library. "

Might be worthwhile to just open a PR over there if the additional buildtime determinism is worth it.

PR #861, opened 2022-04-10
Upgrade from our standard lints v4 -> v6

repi:
And switches to enable them in root `.cargo/config.toml` so they are enabled for all crates in the repository. Also resolves some warnings as errors when running clippy on some of the crates that we had.

The code fixups for this was mostly from the [`clippy::single_match_else`](https://rust-lang.github.io/rust-clippy/master/#single_match_else) lint. In our other code bases we've been seeing some good improvements from the style that that enforces as avoids deeper indentation, though in this codebase the both cases look quite similar but probably doesn't hurt to just use our default.


.cargo/config
```diff
@@ -1,2 +1,85 @@
...
+    "-Wclippy::verbose_file_reads",
+    "-Wclippy::zero_sized_map_values",
+    "-Wfuture_incompatible",
+    "-Wnonstandard_style",
+    "-Wrust_2018_idioms",
+    # END - Embark standard lints v6 for Rust 1.55+
+
+    # repo specific lints
+    "-Aunsafe_code",
+]
```
eddyb, line 85: Missing newline at the end of file.

eddyb (changes requested):
LGTM, just one tiny nit.

.cargo/config
repi: hah fixed

repi:
Fixed your nit @eddyb :) 

CI fails, but currently all CI fails on Mac due to #865, so just going to merge this one

PR #862, opened 2022-04-11
rustup: update to nightly-2022-04-11.

eddyb:
As you can see by the commits, I've tried doing this in steps, but it still was quite messy.
<sup>(If we needed any proof that we should rustup more often, heh)</sup>


repi (approved):
Nice!

Tested this in Ark and compiled fine without any changes and visuals look like they should look like :)

Issue #866, opened 2022-04-17
Upgrade CI to use latest prebuilt spirv-tools on Windows & Linux


repi:
We currently use a very old version from 2021-08-05 which doesn't fail our tests, but latest does (#865). So need to upgrade our CI to use latest available pre-built spirv-tools.

But can't find any instructions of where to find these magic pre-built URLs and a google search doesn't reveal it either. These URLs:

> https://storage.googleapis.com/spirv-tools/artifacts/prod/graphics_shader_compiler/spirv-tools/linux-clang-release/continuous/1530/20210805-040049

@Jake-Shadle do you remember what site  #162 found what the latest prebuilds are? Ideally we shouldn't be using specific dates/commits either but stable versioned spirv-tools releases instead. Unfortunate that SPIRV-Tools repo still doesn't have any official prebuilds.

Jake-Shadle:
They're from this [page](https://github.com/KhronosGroup/SPIRV-Tools/blob/master/docs/downloads.md), they're just the direct CI artifacts, which is terrible, but... :shrug:

repi:
Thanks, will test to upgrade to it and add link in the CI definitions to the page for next time

eddyb:
Should we try to keep `spirv-tools-rs` in sync? For example, I couldn't reproduce the validation failures when using the version compiled through `spirv-tools-rs`.

repi:
Think we should, but `spirv-tools-rs` also is on a frozen fork of SPRIV-Tools that we need to update also. 

@Jake-Shadle could you do that?

Jake-Shadle:
Sure.

Issue #865, opened 2022-04-17
CI tests fails with latest spirv-val


repi:
The Mac builds in CI use latest available `spirv-val` and that started failing the tests now. Possibly from the [SPIRV-Tools v2022.2 release](https://github.com/KhronosGroup/SPIRV-Tools/releases/tag/v2022.2) 10 days ago

https://github.com/EmbarkStudios/rust-gpu/runs/6053175509?check_suite_focus=true

```
test [ui] ui/image/sample_depth_reference/sample_gradient.rs ... FAILED
normalized stderr:
error: error:0:0 - [VUID-StandaloneSpirv-OpImage-04777] In Vulkan, OpImage*Dref* instructions must not use images with a 3D Dim
  |
  = note: spirv-val failed
  = note: module `$TEST_BUILD_DIR/image/sample_depth_reference/sample_lod.stage-id.spv.dir/module`

error: aborting due to previous error
```

Linux and Windows tests do not fail as they use a fixed frozen spirv-val version that it downloads, though we should upgrade that also. Filed:
- #866

eddyb:
Fixed in the commit I added to #867 (see https://github.com/EmbarkStudios/rust-gpu/pull/867#pullrequestreview-945526022).

PR #867, opened 2022-04-17
Upgrade SPIRV-Tools prebuild to 2022-04-14

repi:
For Linux and Windows, Mac CI uses latest installed from `brew`

Also added comment about how/where to get the download link for next upgrade.

Resolves: #866

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,7 +22,7 @@ jobs:
             target: aarch64-linux-android
     runs-on: ${{ matrix.os }}
     env:
-      spirv_tools_version: "20210805"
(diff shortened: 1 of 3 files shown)


repi:
Confirmed that we get the new validation error (#865) not just on Mac but on Linux and Windows with this now also. So that is high priority to fix. Could potentially just merge this still so we get the same failure on all platforms @eddyb ?

eddyb (approved):
Pushed a commit adjusting the tests to not use `OpImage*Dref*` on cubemaps on Vulkan, in accordance with [VUID-StandaloneSpirv-OpImage-04777](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-StandaloneSpirv-OpImage-04777):
> `OpImage*Dref*` instructions **must** not consume an image whose `Dim` is 3D

If it works, we should have CI green again.

PR #868, opened 2022-04-17
examples: fix pipeline layout in compute runner

msiglreith:
(Remove the outdated comment which refered to the prior empty compute shader and doesn't apply anymore as the current example takes one buffer descriptor.)

The motivation behind `min_binding_size` is not totally clear to me but seems to involve the buffer size bound to the descriptor. Two options appear valid here:
- `None`: defers the wgpu check to runtime and it will internally validate that the bound buffer matches the shader interface.
- `Some(n * sizeof::<u32>())`: It needs to be able to hold at least one element ([u32] buffer) -> `Some(NonZeroU64::new(4).unwrap())` at least

Both approaches work fine locally but the `None` case seems less of a hassle

### DIFF
--- a/examples/runners/wgpu/src/compute.rs
+++ b/examples/runners/wgpu/src/compute.rs
@@ -2,7 +2,7 @@ use wgpu::util::DeviceExt;
 
 use super::Options;
 use futures::future::join;
-use std::{convert::TryInto, future::Future, num::NonZeroU64, time::Duration};
+use std::{convert::TryInto, future::Future, time::Duration};
 
 fn block_on<T>(future: impl Future<Output = T>) -> T {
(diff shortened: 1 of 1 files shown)


VZout (approved):
Thanks!

hrydgard (approved):

fu5ha (approved):

hrydgard:
Still seems to be some CI issue, a couple tests fail.

msiglreith:
The test errors seem unrelated: https://github.com/EmbarkStudios/rust-gpu/issues/865

hrydgard:
Oh right. I'll go ahead and merge this then, let's get those solved separately.

Issue #873, opened 2022-05-16
Spirv-val failed on shader


molikto:
### Expected Behaviour

spriv-val should not fail.

### Example & Steps To Reproduce

This shader fails with `spirv-val` error

But either:

1. move out of the closure
2. remove the rayquery
3. remove the `index_unchecked` line

Make it compile successfully.

```
#![feature(asm_experimental_arch)]
#![cfg_attr(
    target_arch = "spirv",
    no_std,
    feature(register_attr),
    register_attr(spirv)
)]

use spirv_std::arch::IndexUnchecked;
#[cfg(not(target_arch = "spirv"))]
use spirv_std::macros::spirv;
use spirv_std::ray_tracing::{AccelerationStructure, CommittedIntersection, RayFlags, RayQuery};
use spirv_std::{glam::Vec3, ray_query};

pub fn some_fun(c: impl Fn() -> bool) {
    let mut i = 0;
    while i < 10 {
        if c() {
            return;
        }
        i += 1;
    }
}
#[spirv(intersection)]
pub fn voxels_gbuffer_is(
    #[spirv(storage_buffer, descriptor_set = 1, binding = 1)] vertices: &[u32],
    #[spirv(descriptor_set = 1, binding = 3)] acceleration_structure: &AccelerationStructure,
) {
    some_fun(|| {
        unsafe {
            ray_query!(let mut handle);
            handle.initialize(
                acceleration_structure,
                RayFlags::TERMINATE_ON_FIRST_HIT,
                0xFF,
                Vec3::new(0.0, 0.0, 0.0),
                0.0,
                Vec3::new(0.0, 0.0, 0.0),
                10000.0,
            );
            unsafe {
                // Indices of the triangle
                let ind = vertices.index_unchecked(0);
            };
            true
        }
    });
}
```

## System Info

 - Rust: rustc 1.62.0-nightly (1f7fb6413 2022-04-10)
 - OS: Windows 10  Version	10.0.19044 Build 19044
 - GPU: RTX 3070 Ti
 - SPIR-V: SPIRV-Tools v2022.2-dev v2022.1-6-g45dd184c

## Backtrace

<details><summary>Backtrace</summary>
<p>

```
error: error:0:0 - OpLoad Pointer <id> '58[%58]' is not a logical pointer.
         %61 = OpLoad %20 %58
  |
  = note: spirv-val failed
```

</p>
</details>

oisyn:
This is the relevant error:
```
[...]

PR #874, opened 2022-05-17
fix: capturing deep composites

molikto:
try fixing https://github.com/EmbarkStudios/rust-gpu/issues/873

The tests I added still errors.

But the wired thing is, if I compile the tests using my own setup in my own repo, it compiles successfully, but in the tests it fails with a different error. Also this change fixes my original problem.

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
+++ b/crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
@@ -14,29 +14,31 @@ pub fn destructure_composites(function: &mut Function) {
         .all_inst_iter()
         .filter_map(|inst| match inst.class.opcode {
             Op::CompositeConstruct => Some((inst.result_id.unwrap(), inst.clone())),
-            Op::CompositeInsert if inst.operands.len() == 3 => {
+            Op::CompositeInsert => {
                 Some((inst.result_id.unwrap(), inst.clone()))
             }
             _ => None,
         })
         .collect();
     for inst in function.all_inst_iter_mut() {
-        if inst.class.opcode == Op::CompositeExtract && inst.operands.len() == 2 {
+        if inst.class.opcode == Op::CompositeExtract {
             let mut composite = inst.operands[0].unwrap_id_ref();
-            let index = inst.operands[1].unwrap_literal_int32();
+            let index: Vec<u32> = inst.operands[1..].iter().map(|i| i.unwrap_literal_int32()).collect();
 
             let origin = loop {
                 if let Some(inst) = reference.get(&composite) {
                     match inst.class.opcode {
                         Op::CompositeInsert => {
-                            let insert_index = inst.operands[2].unwrap_literal_int32();
+                            let insert_index: Vec<u32> = inst.operands[2..].iter().map(|i| i.unwrap_literal_int32()).collect();
                             if insert_index == index {
                                 break Some(inst.operands[0].unwrap_id_ref());
                             }
                             composite = inst.operands[1].unwrap_id_ref();
                         }
                         Op::CompositeConstruct => {
-                            break inst.operands.get(index as usize).map(|o| o.unwrap_id_ref());
+                            if index.len() ==1 {
+                                break inst.operands.get(index[0] as usize).map(|o| o.unwrap_id_ref());
+                            }
                         }
                         _ => unreachable!(),
                     }
--- a/tests/ui/lang/control_flow/closure_capture_invalid_01.rs
+++ b/tests/ui/lang/control_flow/closure_capture_invalid_01.rs
@@ -1,0 +1,32 @@
+// build-pass
+// compile-flags: -Ctarget-feature=+RayTracingKHR,+RayQueryKHR,+ext:SPV_KHR_ray_tracing,+ext:SPV_KHR_ray_query
+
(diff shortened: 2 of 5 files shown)


crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
```diff
@@ -14,29 +14,31 @@ pub fn destructure_composites(function: &mut Function) {
...
             }
             _ => None,
         })
         .collect();
     for inst in function.all_inst_iter_mut() {
-        if inst.class.opcode == Op::CompositeExtract && inst.operands.len() == 2 {
+        if inst.class.opcode == Op::CompositeExtract {
             let mut composite = inst.operands[0].unwrap_id_ref();
-            let index = inst.operands[1].unwrap_literal_int32();
+            let index: Vec<u32> = inst.operands[1..].iter().map(|i| i.unwrap_literal_int32()).collect();
```
eddyb, line 26:
```suggestion
            let indices = &inst.operands[1..];
```

You only need to do the `.unwrap_literal_int32()` on `indices[0]` in the `break` statement.

crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
```diff
@@ -14,29 +14,31 @@ pub fn destructure_composites(function: &mut Function) {
...
 
             let origin = loop {
                 if let Some(inst) = reference.get(&composite) {
                     match inst.class.opcode {
                         Op::CompositeInsert => {
-                            let insert_index = inst.operands[2].unwrap_literal_int32();
+                            let insert_index: Vec<u32> = inst.operands[2..].iter().map(|i| i.unwrap_literal_int32()).collect();
```
eddyb, line 32:
```suggestion
                            let insert_indices = &inst.operands[2..];
```

crates/rustc_codegen_spirv/src/linker/destructure_composites.rs
```diff
@@ -14,29 +14,31 @@ pub fn destructure_composites(function: &mut Function) {
...
                             if insert_index == index {
                                 break Some(inst.operands[0].unwrap_id_ref());
                             }
                             composite = inst.operands[1].unwrap_id_ref();
                         }
                         Op::CompositeConstruct => {
-                            break inst.operands.get(index as usize).map(|o| o.unwrap_id_ref());
+                            if index.len() ==1 {
```
eddyb, line 39:
```suggestion
                            if index.len() == 1 {
```

tests/ui/lang/control_flow/closure_capture_invalid_01.rs
```diff
@@ -0,0 +1,32 @@
+// build-pass
```
eddyb, line 1:
```suggestion
// build-fail
```
(because you have a `.stderr` file and presumably compilation fails)

tests/ui/lang/control_flow/closure_capture_invalid_02.rs
```diff
@@ -0,0 +1,43 @@
+// build-pass
```
eddyb, line 1:
```suggestion
// build-fail
```
(because you have a `.stderr` file and presumably compilation fails)

eddyb (changes requested):
LGTM modulo comments, but I am a bit confused: is it actually possible for us to end up with `OpCompositeInsert`s with multiple indices? Do we have a pass that *generates* those?
(I'm asking because AFAIK, `rustc_codegen_ssa` sure doesn't have the ability to emit anything similar, at most it does single-level and usually only for `ScalarPair` anyway)

molikto:
> LGTM modulo comments, but I am a bit confused: is it actually possible for us to end up with `OpCompositeInsert`s with multiple indices? Do we have a pass that _generates_ those? (I'm asking because AFAIK, `rustc_codegen_ssa` sure doesn't have the ability to emit anything similar, at most it does single-level and usually only for `ScalarPair` anyway)

I suspect it's if a closure captures an storage buffer (`&[u32]`) and then another acceleration structure, the closure capturing generate a top level composite, then the array is another composite (len and pointer)

eddyb:
> I suspect it's if a closure captures an storage buffer (`&[u32]`) and then another acceleration structure, the closure capturing generate a top level composite, then the array is another composite (len and pointer)

That's true at the *type* level, but is there something that generates the equivalent of `.field.subfield` extracts/inserts *in a single* SPIR-V instruction? (as opposed to what `rustc_codegen_ssa` might generate for such independent operations, of one instruction for `.field` and one for `.subfield` - tho it usually only does pointer offsets, not even by-value extract/insert)

molikto:
> > I suspect it's if a closure captures an storage buffer (`&[u32]`) and then another acceleration structure, the closure capturing generate a top level composite, then the array is another composite (len and pointer)
> 
> That's true at the _type_ level, but is there something that generates the equivalent of `.field.subfield` extracts/inserts _in a single_ SPIR-V instruction? (as opposed to what `rustc_codegen_ssa` might generate for such independent operations, of one instruction for `.field` and one for `.subfield` - tho it usually only does pointer offsets, not even by-value extract/insert)

it seems when I simplify my reproduction the error is gone, but my original shader does have multi index CompositeInsert. So the testcase isn't correct. Will try to get a new testcase if possible

PR #875, opened 2022-05-17
warn when inline(never) is not respected

molikto:


crates/rustc_codegen_spirv/src/linker/inline.rs
```diff
@@ -35,15 +35,29 @@ pub fn inline(sess: &Session, module: &mut Module) -> super::Result<()> {
...
         }
     });
+    if !inlined_dont_inlines.is_empty() {
+        let names = get_names(module);
+        for f in inlined_dont_inlines {
+            sess.warn(&format!(
+                    "function `{}` has `dont_inline` attribute, but need to be inlined because it has illegal argument or return types",
+                    get_name(&names, f)
+                ));
+        }
```
eddyb, line 58:
```suggestion
        for f in inlined_dont_inlines {
            sess.warn(&format!(
                "`#[inline(never)]` function `{}` needs to be inlined \
                 because it has illegal argument or return types",
                get_name(&names, f)
            ));
        }
```

(I'm not too fond of "illegal" but I can't think of a good substitute, maybe "unsupported" but that doesn't really have the "needs legalization" feel to it so maybe "illegal" is the right choice)

eddyb (changes requested):
Looks great, only had nits about Rust formatting and the warning phrasing (most importantly I think we should say `#[inline(never)]` since that's what would be in the source code).

tests/ui/lang/consts/nested-ref.stderr
```diff
@@ -0,0 +1,6 @@
+warning: function `nested_ref::deep_load` has `dont_inline` attribute, but need to be inlined because it has illegal argument or return types
+
+warning: function `nested_ref::deep_transpose` has `dont_inline` attribute, but need to be inlined because it has illegal argument or return types
+
+warning: 2 warnings emitted
+
```
eddyb, line 6: You'll need to `git pull` the new commit locally then run `cargo compiletest --bless` again to update this `.stderr` file.

PR #877, opened 2022-06-01
Add low-level atomic instructions to `spirv_std::arch`

BeastLe9enD:
Hey, I added some atomic instructions to `spirv_std::arch`. 
There was already a discussion to add them, but it wasn't done at the time because the idea was to implement atomic using `core::sync::atomic` (see #442).
But I think it's ok to add the intrinsics in advance and then build the API on top of them later 🙂

### DIFF
--- a/crates/spirv-std/src/arch/atomic.rs
+++ b/crates/spirv-std/src/arch/atomic.rs
@@ -409,7 +409,7 @@ pub unsafe fn atomic_xor<I: Integer, const SCOPE: u32, const SEMANTICS: u32>(
 #[spirv_std_macros::gpu_only]
 #[doc(alias = "OpAtomicFMinEXT")]
 #[inline]
-pub unsafe fn atomic_f_min_ext<F: Float, const SCOPE: u32, const SEMANTICS: u32>(
+pub unsafe fn atomic_f_min<F: Float, const SCOPE: u32, const SEMANTICS: u32>(
     ptr: &mut F,
     value: F,
 ) -> F {
@@ -436,7 +436,7 @@ pub unsafe fn atomic_f_min_ext<F: Float, const SCOPE: u32, const SEMANTICS: u32>
 #[spirv_std_macros::gpu_only]
 #[doc(alias = "OpAtomicFMaxEXT")]
 #[inline]
-pub unsafe fn atomic_f_max_ext<F: Float, const SCOPE: u32, const SEMANTICS: u32>(
+pub unsafe fn atomic_f_max<F: Float, const SCOPE: u32, const SEMANTICS: u32>(
     ptr: &mut F,
     value: F,
 ) -> F {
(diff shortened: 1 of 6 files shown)


crates/spirv-std/src/arch/atomic.rs
```diff
@@ -0,0 +1,487 @@
...
+    }
+
+    old
+}
+
+/// Atomically min a float and return the old value.
+#[spirv_std_macros::gpu_only]
+#[doc(alias = "OpAtomicFMinEXT")]
+#[inline]
+pub unsafe fn atomic_f_min_ext<F: Float, const SCOPE: u32, const SEMANTICS: u32>(
```
expenses, line 412: I'm just quicky skimming over this on my phone, but I think these would probably be fine with out the `_ext` on the end of the function name. Everything else looks good though! 👍

BeastLe9enD: Okay, I removed the ext suffix 🙂

BeastLe9enD:
I saw that an `atomics` branch already existed, so I took the documentation from there and added the `Number` trait as well.
Also, in the long run, it would be better to use the `Scope` and `Semantics` enum types, but that would require `adt_const_params` and `generic_const_exprs`, which are not stable yet, so I decided to stick with `u32` for now

crates/spirv-std/src/number.rs
```diff
@@ -0,0 +1,15 @@
+//! Traits and helper functions related to numbers.
+
+/// Abstract trait representing a SPIR-V integer type.
```
eddyb, line 3:
```suggestion
/// Abstract trait representing a SPIR-V integer or floating-point type.
```

eddyb (changes requested):
Only noticed one comment nit, but otherwise LGTM.

If @expenses and/or @hrydgard are fine with landing this, feel free to merge or ping me to do so.

eddyb:
@BeastLe9enD I just tried pushing a commit fixing the test, but either the `main` branch has different rules, or you opted out of the feature, either way, you need to make this change:
```patch
diff --git a/tests/ui/arch/debug_printf_type_checking.stderr b/tests/ui/arch/debug_printf_type_checking.stderr
index f719d0aba..d422023e6 100644
--- a/tests/ui/arch/debug_printf_type_checking.stderr
+++ b/tests/ui/arch/debug_printf_type_checking.stderr
@@ -100,9 +100,9 @@ error[E0277]: the trait bound `{float}: Vector<f32, 2_usize>` is not satisfied
               <IVec3 as Vector<i32, 3_usize>>
             and 9 others
 note: required by a bound in `debug_printf_assert_is_vector`
-   --> $SPIRV_STD_SRC/lib.rs:144:8
+   --> $SPIRV_STD_SRC/lib.rs:145:8
     |
-144 |     V: crate::vector::Vector<TY, SIZE>,
+145 |     V: crate::vector::Vector<TY, SIZE>,
     |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `debug_printf_assert_is_vector`

 error[E0308]: mismatched types
```
(line numbers in `spirv-std` changed, the test isn't broken, only differs in diagnostic details)

Or `cargo compiletest --bless` if that's easier for you (i.e. if it won't require a rebuild etc.).

<hr/>

Sorry this fell through the cracks, I don't remember hearing from anyone after posting the above review (https://github.com/EmbarkStudios/rust-gpu/pull/877#pullrequestreview-995641915).

eddyb (approved):

PR #892, opened 2022-08-09
Atomic unit tests

wesselfr:
This PR is aimed to adding unit-tests for the atomics added in #877
In addition, this PR is also adding atomic flags.


eddyb:
There's something wrong going on with the commits, your git config might be set up to have your name and email as `=` (or something weirder happened).

Also, you seem to have commits from #877, but it doesn't look like you need a rebase, but rather a rebase already left most of these commits much smaller, only the delta from #877?

wesselfr:
Something went wrong during the rebase, applied my changes again on: https://github.com/EmbarkStudios/rust-gpu/pull/898

PR #894, opened 2022-08-10
Added version tags to some crates referencing rustc_codegen_spirv

oisyn:
Some lingering dependencies to `rustc_codegen_spirv` that don't use the version number. `cargo-release` complains about this.

### DIFF
--- a/tests/Cargo.toml
+++ b/tests/Cargo.toml
@@ -14,5 +14,5 @@ use-compiled-tools = ["rustc_codegen_spirv/use-compiled-tools"]
 
 [dependencies]
 compiletest = { version = "0.7.0", package = "compiletest_rs" }
-rustc_codegen_spirv = { path = "../crates/rustc_codegen_spirv", default-features = false }
+rustc_codegen_spirv = { path = "../crates/rustc_codegen_spirv", version = "0.4.0-alpha.12", default-features = false }
 structopt = "0.3.21"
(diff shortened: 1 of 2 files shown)


repi (approved):

tests/Cargo.toml
```diff
@@ -14,5 +14,5 @@ use-compiled-tools = ["rustc_codegen_spirv/use-compiled-tools"]
 
 [dependencies]
 compiletest = { version = "0.7.0", package = "compiletest_rs" }
-rustc_codegen_spirv = { path = "../crates/rustc_codegen_spirv", default-features = false }
+rustc_codegen_spirv = { path = "../crates/rustc_codegen_spirv", version = "0.4.0-alpha.12", default-features = false }
```
repi, line 17: technically one doesn't need to specify the version here as this is a crate only used in the workspace, not a published crate. so it uses the `path` here instead and ignores the `version`

oisyn: Yeah I was wondering about that. When doing a dry run with `cargo-release` it complained that its version was missing. We do use version info for the other creates that we reference by path, though.

repi: it is fine to leave it here with the version

oisyn: Well that's interesting, so `cargo-release` complained about the missing version tags, but now that they are there it does not update them together with the crate versions themselves 🙄

PR #898, opened 2022-08-11
Atomic unit tests

wesselfr:
This PR is aimed to adding unit-tests for the atomics added in https://github.com/EmbarkStudios/rust-gpu/pull/877
In addition, this PR is also adding atomic flags.

### DIFF
--- a/tests/ui/arch/atomics/atomic_flag_clear.rs
+++ b/tests/ui/arch/atomics/atomic_flag_clear.rs
@@ -1,0 +1,25 @@
+// build-pass
+
+#![allow(incomplete_features)]
+#![feature(generic_const_exprs)]
+use spirv_std::{
+    arch::IndexUnchecked,
+    memory::{Scope, Semantics},
+};
+
+#[cfg(target_arch = "spirv")]
+use core::arch::asm;
+
+#[spirv(compute(threads(64)))]
+pub fn main(#[spirv(descriptor_set = 0, binding = 0, storage_buffer)] buffer: &mut [u32]) {
+    // Ensure kernel capabilities.
+    unsafe { asm!("OpCapability Kernel") };
+
+    let old = unsafe {
+        spirv_std::arch::atomic_flag_clear::<
+            _,
+            { Scope::Workgroup as u32 },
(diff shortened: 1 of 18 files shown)


tests/ui/arch/atomics/atomic_xor.rs
```diff
@@ -0,0 +1,19 @@
+// build-pass
+
+#![allow(incomplete_features)]
+#![feature(generic_const_exprs)]
```
eddyb, line 4: Do all of these tests need `#![feature(generic_const_exprs)]`?

tests/ui/arch/atomics/atomic_flag_clear.rs
```diff
@@ -0,0 +1,25 @@
...
+    memory::{Scope, Semantics},
+};
+
+#[cfg(target_arch = "spirv")]
+use core::arch::asm;
+
+#[spirv(compute(threads(64)))]
+pub fn main(#[spirv(descriptor_set = 0, binding = 0, storage_buffer)] buffer: &mut [u32]) {
+    // Ensure kernel capabilities.
+    unsafe { asm!("OpCapability Kernel") };
```
eddyb, line 16: Do we have other tests that do this? My understanding was that we've given up on `Kernel` support (could be wrong though).

eddyb (changes requested):
LGTM, modulo the `Kernel` capabilities and double-checking the `generic_const_exprs` feature-gate.

tests/ui/arch/atomics/atomic_flag_clear.rs
wesselfr: `Kernel` capability is required for the atomic flag operations. Other test that require specific capabilities are the atomic floats, which I've omitted for the reason that the capability is vendor specific (Nvidia only if I'm correct). The same can be done for the atomic flags tests if required

tests/ui/arch/atomics/atomic_xor.rs
wesselfr: Seems like these are redundant, they have been removed

PR #899, opened 2022-08-11
Updated versions to 0.4.0-alpha.13

oisyn:


repi (approved):
Think for all the examples we could just set `version = 0.0.0` and `publish = false` as they are not published nor versioned and just part of the workspace.

but this also does work

oisyn:
They currently already are `publish=false` but I'll set the version to 0.0.0 in the next version update

PR #909, opened 2022-08-17
Updated to nightly-2022-04-18

oisyn:
This also fixes #907.

PR #910, opened 2022-08-17
Use the same official SPIRV-Tools builds for macOS CI too.

eddyb:
See https://github.com/EmbarkStudios/rust-gpu/pull/909#issuecomment-1218042649 for some context - `brew install` is picking up `2022.3` while we're still only supporting `2022.2` currently (we *should* upgrade but it shouldn't block landing other PRs).

I've tried to organize and add more documentation (props to @oisyn for showing me [Google Cloud Storage XML API docs](https://cloud.google.com/storage/docs/xml-api/get-bucket-list), it was really hard to get access to historical build versions without that) - so that next time it should be much easier to update all 3 OSes.

However, I didn't touch the Lint builder (should it share an env var? how?) and also there's some Linux/macOS unification that can be done since they're both doing the same thing (assuming that even works).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,32 +22,43 @@ jobs:
             target: aarch64-linux-android
     runs-on: ${{ matrix.os }}
     env:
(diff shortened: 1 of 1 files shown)


repi (approved):
sweet, definitely a good idea to have exact same version on all platforms in CI

PR #912, opened 2022-08-22
Updated nightly to 2022-05-16

oisyn:
(edit: while this is awaiting a review, I'll just pile on extra updates until we can merge into main)

`2022-05-02` and `2022-04-25` compiled and tested without any changes
`2022-05-09` had some API changes, and I fixed a linter warning
`2022-05-16` more API changes, blessed a test that failed because of changed line numbers


eddyb (approved):
Looking great!

I was holding on the review to see how far you would get today, but I should leave this review anyway so you can land it as-is if you hit a roadblock with later nightlies.

(Also, you might want to set up format-on-save if you're using `rustfmt`, or otherwise run `cargo fmt` manually - and that reminds me that I should probably make RA run `cargo clippy` instead of `cargo check`, so that clippy failures are also found)

PR #914, opened 2022-08-26
Updated to nightly-2022-06-13

oisyn:


crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
```diff
@@ -802,7 +802,7 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
         id_map: &mut FxHashMap<&'a str, Word>,
         defined_ids: &mut FxHashSet<Word>,
         token: Token<'a, 'cx, 'tcx>,
-    ) -> Option<OutRegister<'a>> {
+    ) -> Option<OutRegister<'tcx>> {
```
eddyb, line 805: Can you also change the definition of `OutRegister` to not say `'a` at all (but `'tcx` instead)?

eddyb (changes requested):

PR #915, opened 2022-08-29
Update to nightly-2022-07-04

oisyn:
The update to `07-04` gave a dead code warning on `fn call_glsl_op_with_ints()` in `crates\spirv-std\src\arch.rs:191`. I seem to be able to silence the warning, but I'm a bit confused by I'm getting it in the first place, since it's obviously being used by the public functions below it.

### DIFF
--- a/crates/spirv-std/src/arch.rs
+++ b/crates/spirv-std/src/arch.rs
@@ -188,6 +188,7 @@ pub unsafe fn read_clock_uvec2_khr<V: Vector<u32, 2>, const SCOPE: u32>() -> V {
 }
 
 #[spirv_std_macros::gpu_only]
+#[allow(dead_code)]
 unsafe fn call_glsl_op_with_ints<T: Integer, const OP: u32>(a: T, b: T) -> T {
     let mut result = T::default();
     asm!(
--- a/crates/spirv-std/src/arch.rs
+++ b/crates/spirv-std/src/arch.rs
@@ -3,13 +3,13 @@
 //! This module is intended as a low level abstraction over SPIR-V instructions.
 //! These functions will typically map to a single instruction, and will perform
 //! no additional safety checks beyond type-checking.
+#[cfg(target_arch = "spirv")]
+use crate::integer::Integer;
 use crate::{
     integer::{SignedInteger, UnsignedInteger},
     scalar::Scalar,
     vector::Vector,
 };
-#[cfg(target_arch = "spirv")]
-use crate::integer::Integer;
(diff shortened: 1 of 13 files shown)


crates/spirv-std/src/arch.rs
```diff
@@ -188,6 +188,7 @@ pub unsafe fn read_clock_uvec2_khr<V: Vector<u32, 2>, const SCOPE: u32>() -> V {
 }
 
 #[spirv_std_macros::gpu_only]
+#[allow(dead_code)]
```
eddyb, line 191: ~~This *seems* like a bug? You could use `#[expect(dead_code)]` instead, so that we get an error if it does get fixed, e.g.: https://godbolt.org/z/sdTb3neqz (you won't need the `#[deny(unfulfilled_lint_expectations)]` since we have `#![deny(warnings)]` IIRC)~~

Or maybe it's only linting when compiled for the CPU, because of `#[spirv_std_macros::gpu_only]` (which replaces the body of the function with an unconditional panic).

In that case, this might be fixable by keeping `#[spirv_std_macros::gpu_only]` on `call_glsl_op_with_ints` but removing it from all callers of `call_glsl_op_with_ints` (it's redundant anyway, they just dispatch to `call_glsl_op_with_ints`).

If we want to keep the `#[spirv_std_macros::gpu_only]` on *only* the exported functions (e.g. if the panic mentions the function name), then I think we can use `#[cfg(target_arch = "spirv")]` instead of `#[spirv_std_macros::gpu_only]` on `call_glsl_op_with_ints`.

In fact that sounds simpler, I am going to suggest it now.

eddyb, line 191:
```suggestion
#[cfg(target_arch = "spirv")]
```

(more details in the other comment, but this might just work, and it's a clean local change)

eddyb (changes requested):
LGTM, modulo picking a solution for the "unused" warning (left a couple comments, you can just try the diff suggestion on CI if you want).

crates/spirv-std/src/arch.rs
oisyn: > (which replaces the body of the function with an unconditional panic).
Aah, I thought the function would be removed altogether. So the warning makes sense. But why do we start hearing about it now?

oisyn:
Oh, that's why 😂. Yeah that's kinda annoying.

oisyn:
Oh for crying out loud

eddyb (approved):

PR #919, opened 2022-09-21
Added toolchain check to rustc_codegen_spirv

oisyn:
Check can be disabled by specifying `skip-toolchain-check` feature on either `rustc_codegen_spirv` or `spirv_builder`

### DIFF
--- a/crates/rustc_codegen_spirv/build.rs
+++ b/crates/rustc_codegen_spirv/build.rs
@@ -1,59 +1,75 @@
 //! This custom build script merely checks whether we're compiling with the appropriate Rust toolchain
 
-#[allow(unused_imports)]
-use colored::*;
-use std::process::Command;
+#![allow(clippy::string_add)]
 
-/// The required toolchain version for this package. Update accordingly.
-#[cfg(not(feature = "skip-toolchain-check"))]
-const REQUIRED_TOOLCHAIN: &str = "nightly-2022-08-29";
+use std::error::Error;
+use std::process::{Command, ExitCode};
 
-/// The required toolchain commit hash for this package. Update accordingly.
-#[cfg(not(feature = "skip-toolchain-check"))]
-const REQUIRED_COMMIT_HASH: &str = "ce36e88256f09078519f8bc6b21e4dc88f88f523";
+/// Current `rust-toolchain` file
+/// Unfortunately, directly including the actual workspace `rust-toolchain` doesn't work together with
+/// `cargo publish`. We need to figure out a way to do this properly, but let's hardcode it for now :/
+//const REQUIRED_RUST_TOOLCHAIN: &str = include_str!("../../rust-toolchain");
+const REQUIRED_RUST_TOOLCHAIN: &str = r#"[toolchain]
+channel = "nightly-2022-08-29"
+components = ["rust-src", "rustc-dev", "llvm-tools-preview"]
+# commit_hash = ce36e88256f09078519f8bc6b21e4dc88f88f523"#;
 
-#[cfg(not(feature = "skip-toolchain-check"))]
-fn get_rustc_commit_hash() -> Option<String> {
-    let rustc = std::env::var("RUSTC").unwrap_or(String::from("rustc"));
-    Command::new(rustc)
-        .arg("-vV")
-        .output()
-        .ok()
-        .and_then(|output| String::from_utf8(output.stdout).ok())
-        .map(|s| {
-            for l in s.lines() {
-                match l.strip_prefix("commit-hash: ") {
-                    Some(hash) => {
-                        return Some(hash.to_string());
-                    }
-                    _ => {}
-                }
-            }
-            None
-        })
-        .unwrap_or(None)
+fn get_rustc_commit_hash() -> Result<String, Box<dyn Error>> {
+    let rustc = std::env::var("RUSTC").unwrap_or_else(|_| String::from("rustc"));
+    String::from_utf8(Command::new(rustc).arg("-vV").output()?.stdout)?
+        .lines()
+        .find_map(|l| l.strip_prefix("commit-hash: "))
+        .map(|s| s.to_string())
+        .ok_or_else(|| Box::<dyn Error>::from("`commit-hash` not found in `rustc -vV` output"))
 }
-#[cfg(not(feature = "skip-toolchain-check"))]
-fn check_toolchain_version() {
-    std::env::set_var("CLICOLOR_FORCE", "1"); // make sure our coloring gets through to cargo
 
-    let current_hash = get_rustc_commit_hash().unwrap_or(String::from("<unknown>"));
-    if current_hash != REQUIRED_COMMIT_HASH {
-        eprintln!(
-            "{}: {} (found {}). Make sure you specify {} in your project's {} file",
-            "error".bright_red(),
-            "Wrong toolchain detected".bold(),
-            current_hash.bright_cyan(),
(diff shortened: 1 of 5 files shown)


crates/rustc_codegen_spirv/build.rs
```diff
@@ -0,0 +1,59 @@
+//! This custom build script merely checks whether we're compiling with the appropriate Rust toolchain
+
+#[allow(unused_imports)]
+use colored::*;
+use std::process::Command;
+
+/// The required toolchain version for this package. Update accordingly.
+#[cfg(not(feature = "skip-toolchain-check"))]
+const REQUIRED_TOOLCHAIN: &str = "nightly-2022-08-29";
```
eddyb, line 9: This not being checked against the top-level `rust-toolchain` file seems potentially fragile.

You could look at `env::get("RUSTUP_TOOLCHAIN")` and maybe error if *it is set* and it's a different value.

`include_str!("../../../rust-toolchain")` might also work great, in terms of both:
* not having to hardcode the date here (potentially the hash could also be moved there as a comment)
* being able to *print* the file to the user as something they can verbatim paste into the project to get things working (it's not just the date, but also the necessary components)

That should work at least for git dependencies (since the whole repo gets downloaded somewher in `~/.cargo/`), but to make sure this can work *in general* (crates.io etc.), you would have to use `cargo publish -p rustc_codegen_spirv --dry-run` (or `cargo package -p rustc_codegen_spirv` directly, that seems less likely to risk actually publishing).

crates/rustc_codegen_spirv/build.rs
```diff
@@ -0,0 +1,59 @@
...
+        .unwrap_or(None)
+}
+#[cfg(not(feature = "skip-toolchain-check"))]
+fn check_toolchain_version() {
+    std::env::set_var("CLICOLOR_FORCE", "1"); // make sure our coloring gets through to cargo
+
+    let current_hash = get_rustc_commit_hash().unwrap_or(String::from("<unknown>"));
+    if current_hash != REQUIRED_COMMIT_HASH {
+        eprintln!(
+            "{}: {} (found {}). Make sure you specify {} in your project's {} file",
```
eddyb, line 43:
```suggestion
            "{}: {} (found `{}`). Make sure you specify `{}` in your project's `{}` file",
```
(it's standard in `rustc` error messages to clearly separate natural language phrasing from data/code/etc. being embedded)

Also, you may want to also show the expected hash and whatnot, but also it's probably good to do the `RUSTUP_TOOLCHAIN` check I mentioned above: since that would result in something much more readable for the user (especially if they're just on older pinned nightly!).

crates/rustc_codegen_spirv/build.rs
```diff
@@ -0,0 +1,59 @@
...
+            "error".bright_red(),
+            "Wrong toolchain detected".bold(),
```
eddyb, line 45:
```suggestion
            "wrong toolchain detected".bold(),
```
(Rust error messages are never capitalized, tho if you see any in `rustc_codegen_spirv` that are, they need to be fixed)

crates/rustc_codegen_spirv/build.rs
```diff
@@ -0,0 +1,59 @@
...
+            current_hash.bright_cyan(),
+            format!("channel=\"{}\"", REQUIRED_TOOLCHAIN).bright_cyan(),
```
eddyb, line 47: AFAIK this is not a valid file on its own without `[toolchain]`, and even then there's the issue of the other setting (`components = ["rust-src", "rustc-dev", "llvm-tools-preview"]`), this is why I wrote above that it would be really cool if we can just have this build script reuse the `rust-toolchain` file itself.

eddyb (changes requested):
Cool idea! Most of my comments are nitpicks, especially the more aesthetic ones, the main worry around these things is making sure they're not so fragile that the next time we bump the nightly we forget to update something etc.

crates/rustc_codegen_spirv/build.rs
```diff
@@ -0,0 +1,59 @@
...
+                    Some(hash) => {
+                        return Some(hash.to_string());
+                    }
+                    _ => {}
+                }
+            }
+            None
+        })
+        .unwrap_or(None)
+}
```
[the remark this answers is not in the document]
oisyn: I had considered it but I didn't think it was appropriate. The thing I dislike most here is the hardcoded constants. I was hoping I could put them in Cargo.toml but there doesn't seem to be a way. I could of course just put them in some file and read that.

crates/rustc_codegen_spirv/build.rs
```diff
@@ -0,0 +1,59 @@
...
+#[cfg(not(feature = "skip-toolchain-check"))]
+const REQUIRED_COMMIT_HASH: &str = "ce36e88256f09078519f8bc6b21e4dc88f88f523";
+
+#[cfg(not(feature = "skip-toolchain-check"))]
+fn get_rustc_commit_hash() -> Option<String> {
+    let rustc = std::env::var("RUSTC").unwrap_or(String::from("rustc"));
+    Command::new(rustc)
+        .arg("-vV")
+        .output()
+        .ok()
```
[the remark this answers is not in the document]
oisyn: Loving these nitpicks btw. So many ways to solve the same problem, it's hard to see the trees through the forest.

crates/rustc_codegen_spirv/build.rs
oisyn: I had considered this (roughly), I mainly thought this wouldn't work through crates.io. I can certainly test.

crates/rustc_codegen_spirv/build.rs
oisyn: Oh you mean in terms of "hey, this should be what your rust-toolchain file should look like: <exact contents>"?

crates/rustc_codegen_spirv/build.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    let current_hash = get_rustc_commit_hash().unwrap_or(String::from("<unknown>"));
+    if current_hash != REQUIRED_COMMIT_HASH {
+        eprintln!(
+            "{}: {} (found {}). Make sure you specify {} in your project's {} file",
+            "error".bright_red(),
+            "Wrong toolchain detected".bold(),
+            current_hash.bright_cyan(),
+            format!("channel=\"{}\"", REQUIRED_TOOLCHAIN).bright_cyan(),
+            "rust_toolchain".bright_cyan()
```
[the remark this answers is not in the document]
oisyn: I'm assuming you meant `conhost`, the old terminal emulator, as `cmd.exe` itself is just a console process. Virtual terminal emulation is the default on windows 10. I did some digging, and it seems that ENABLE_VIRTUAL_TERMINAL_PROCESSING did not exist prior to win10, so neither did ANSI coloring support. Cargo depends on `termcolor`, which tries to set that flag and uses the console API to do coloring otherwise, so I would suspect it breaks down before win10.

Perhaps I should switch to `termcolor` also. That way we at least get clean uncolored text if ANSI coloring is not supported.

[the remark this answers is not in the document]
oisyn: Heh we have to anyway since `colored` is essentially denied 🤔

crates/rustc_codegen_spirv/build.rs
oisyn: Indeed, `cargo package` fails on `include_str!("../../rust-toolchain")` :/

repi (approved):

PR #920, opened 2022-09-22
Version bump to 0.4.0-alpha.15.

oisyn:


repi (approved):

repi:
Get this in after #919 gets in right so we have the Rust version check?

oisyn:
Yes exactly. Well I guess the order doesn't matter, as long as I tag the later commit and publish after 🙂

PR #923, opened 2022-09-29
Optimise CI with faster Linux machines

repi:
Build on 16 vCPU machine instead of the default 4 vCPU, using GitHub Action's beta of GitHub-hosted larger runner.

They have support for Windows machines also, but got a some weird compiletest failures on that so disabled that for now (any ideas what could cause that failure?):
- https://github.com/EmbarkStudios/rust-gpu/actions/runs/3160216480/jobs/5144407978

Mac agents remain the main bottleneck for PRs though, and that they do not have hosted agents for. We could potentially disable the Mac builds if we want to improve the PR iteration time. 

But at least with this change we reduce the Linux & Android build times significantly: 

- Android build: 10m 54s -> 5m 13s
- Linux build: total build: 11m 36s -> 4m 43s

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ jobs:
         include:
           - os: ubuntu-20.04-16core
             target: x86_64-unknown-linux-gnu
-          - os: windows-latest
+          - os: windows-2022-16core
             target: x86_64-pc-windows-msvc
           - os: macOS-latest
             target: x86_64-apple-darwin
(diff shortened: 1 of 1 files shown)


oisyn:
What was up with the Windows builder? I noticed it was pending for quite some time.

repi:
Initially it was a configuration problem on my side with the Windows runner,  but after fixing that it runs but fails to build the compile tests: https://github.com/EmbarkStudios/rust-gpu/actions/runs/3160216480/jobs/5144407978

don't understand that error

oisyn:
Weird. I did find one thing that seems similar: https://github.com/rust-lang/rust/issues/53332. Not sure why it's being triggered by this platform.

Unrelated, two tasks are blocking this PR from being merged. More config issues?

repi:
Ah yes need to update in the repo branch protection settings which builds are required as they got new names with this. Will fix

repi:
there!

PR #924, opened 2022-10-04
Avoid rebuilds between `example-runner-{ash,wgpu}`.

eddyb:
The goal is simple: repeatedly switching between these two commands *should build nothing*:
```
cargo run --release -p example-runner-ash
cargo run --release -p example-runner-wgpu
```

The issues fixed in this PR were of two kinds (and I made a commit for each):
* non-build-script usage of `spirv-builder` needs to set environment variables to make it think it *is* running in a build script, in order to get the separate target directory, and avoid tainting the original `target` dir
* the ongoing whack-a-mole with features of (transitive) dependencies of `rustc_codegen_spirv`:
  * because `rustc_codegen_spirv` is a dylib, Cargo enforces that *there must be only one*, which will be replaced on every built if necessary (and we wouldn't want to build it twice *anyway*)
  * this started in #546
    * that PR actually made a mistake in putting `syn` in `[dependencies]`, since it's a transitive dependency of proc macros, so `[build-dependencies]` has to be used instead - however, at the time we were using `spirv-builder` as a build dep itself, so it's possible that may have obfuscated the issue
  * this PR adds features `once_cell/default`, `regex/perf`, `smallvec/union` and `syn/extra-traits` to the hacky list (and whatever `syn/full` adds to what we had before I moved it to `[build-dependencies]`)
  * we should really automate testing for this but I can't think of a *nice* way to do it

Ideally after running these 3 commands once, they should not output anything *the second time*:
```rust
cargo build --release -p spirv-builder 2>&1 | grep Compiling
cargo build --release -p example-runner-ash 2>&1 | grep Compiling
cargo build --release -p example-runner-wgpu 2>&1 | grep Compiling
```
(but that approach feels a bit too messy to easily test in CI)

---

I also looked at the [`cargo-hakari`](https://docs.rs/cargo-hakari) tooling, which seems great *if we start using it as intended*, but trying to only use it for verification leads to false positives because we can't input only the "scenarios" we care about (such as specifying a crate to build with `-p`, on a specific platform), and it includes things like "`example-runner-wgpu` without any `[target.*.dependencies]`" that I can't seem to exclude.


examples/runners/ash/src/main.rs
```diff
@@ -197,6 +197,15 @@ pub fn main() {
 }
 
 pub fn compile_shaders() -> Vec<SpvFile> {
+    // Hack: spirv_builder builds into a custom directory if running under cargo, to not
```
fu5ha, line 200: Since this is in example code, maybe make a note that this is not required for normal use and is there to make developing rust-gpu itself easier.

examples/runners/ash/build.rs
```diff
@@ -0,0 +1,8 @@
+use std::env;
+
+fn main() {
+    // While OUT_DIR is set for both build.rs and compiling the crate, PROFILE is only set in
+    // build.rs. So, export it to crate compilation as well.
```
fu5ha, line 5: Same as other comment

fu5ha:
Ah, well, unfortunate that bot merged instantly lol. Whatever, can just fix that in another commit it's pretty trivial

examples/runners/ash/src/main.rs
eddyb: This is not my comment, I just replicated it from the `wgpu` runner:
https://github.com/EmbarkStudios/rust-gpu/blob/6dc23afaf80f5917f11ccc8ea66f59b10aef98d4/examples/runners/wgpu/src/lib.rs#L96-L103

(one hint is `// Hack:`, I always write `// HACK:` or `// HACK(eddyb):`, heh)

PR #925, opened 2022-10-04
Update wgpu (and ash, to avoid duplicate deps).

eddyb:
For updating `wgpu`, these backwards-incompatible changes required fixes:
* https://github.com/gfx-rs/wgpu/pull/2698
* https://github.com/gfx-rs/wgpu/pull/2786
* https://github.com/rust-windowing/winit/pull/2376
* https://github.com/gfx-rs/wgpu/pull/3236

The original hope was that with newer `wgpu` we'll get newer `naga` which might fix the bug breaking the mouse shader (likely https://github.com/gfx-rs/naga/issues/1977 - at least with SPIR-T) - but alas, no such luck
(however, replacing the `for` loop with a `while` loop, and turning off SPIR-T with `RUSTGPU_CODEGEN_ARGS=--no-spirt` works around `naga` bugs, and is enough for testing).

---

I've also took the liberty of tweaking a couple dependency versions and (selectively) running `cargo update -p ...` to reduce duplicates in `Cargo.lock`, hence the larger amount of changes than might be expected (and the dependency on https://github.com/EmbarkStudios/ash-molten/pull/71).

---

~~One thing I have not figured out yet is why the mouse shader (with the locally applied workaround) doesn't render new frames except on input events *only with X11* (Wayland is fine) - there may be a "power saving" setting that has changed default in `wgpu`/`winit`?~~

**EDIT**: turns out that was us using `winit` as if we were a GUI application, in "only redraw after events" mode (not sure my workaround here is the best approach, but it works, and enabling VSync makes it acceptable).

### DIFF
--- a/examples/runners/wgpu/src/graphics.rs
+++ b/examples/runners/wgpu/src/graphics.rs
@@ -4,7 +4,7 @@ use super::Options;
 use shared::ShaderConstants;
 use winit::{
     event::{ElementState, Event, KeyboardInput, MouseButton, VirtualKeyCode, WindowEvent},
-    event_loop::{ControlFlow, EventLoop},
+    event_loop::{ControlFlow, EventLoop, EventLoopBuilder},
     window::Window,
 };
 
@@ -37,25 +37,37 @@ async fn run(
     window: Window,
     shader_binary: wgpu::ShaderModuleDescriptor<'static>,
 ) {
-    let instance = wgpu::Instance::new(wgpu::Backends::VULKAN | wgpu::Backends::METAL);
+    let backends = wgpu::util::backend_bits_from_env()
+        .unwrap_or(wgpu::Backends::VULKAN | wgpu::Backends::METAL);
+    let instance = wgpu::Instance::new(wgpu::InstanceDescriptor {
+        backends,
+        dx12_shader_compiler: wgpu::util::dx12_shader_compiler_from_env().unwrap_or_default(),
+    });
 
-    // Wait for Resumed event on Android; the surface is only needed early to
-    // find an adapter that can render to this surface.
-    let mut surface = if cfg!(target_os = "android") {
-        None
+    // HACK(eddyb) marker error type for lazily-created surfaces (e.g. on Android).
+    struct SurfaceCreationPending {
+        preferred_format: wgpu::TextureFormat,
+    }
+
+    // Wait for Resumed event on Android; the surface is only otherwise needed
+    // early to find an adapter that can render to this surface.
+    let initial_surface = if cfg!(target_os = "android") {
+        Err(SurfaceCreationPending {
+            preferred_format: wgpu::TextureFormat::Rgba8UnormSrgb,
+        })
     } else {
-        Some(unsafe { instance.create_surface(&window) })
+        Ok(unsafe { instance.create_surface(&window) }
+            .expect("Failed to create surface from window"))
     };
 
-    let adapter = instance
-        .request_adapter(&wgpu::RequestAdapterOptions {
-            power_preference: wgpu::PowerPreference::default(),
-            force_fallback_adapter: false,
-            // Request an adapter which can render to our surface
-            compatible_surface: surface.as_ref(),
-        })
-        .await
-        .expect("Failed to find an appropriate adapter");
+    let adapter = wgpu::util::initialize_adapter_from_env_or_default(
+        &instance,
+        backends,
+        // Request an adapter which can render to our surface
+        initial_surface.as_ref().ok(),
+    )
+    .await
+    .expect("Failed to find an appropriate adapter");
 
     let features = wgpu::Features::PUSH_CONSTANTS;
     let limits = wgpu::Limits {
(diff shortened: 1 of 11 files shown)


eddyb:
This is incredible, the `cargo-deny` failure is because of something needed *for font rendering*.

Why would we need to render text? Well, we don't, except... Look at what `winit` needs to support Wayland:
* https://github.com/rust-windowing/winit/issues/1802
  * (this issue was closed effectively as "wontfix" and does remain a problem)

So according to that, there's this "SCTK" (`smithay-client-toolkit`) which among other things handles CSDs on Wayland (where CSD is "Client-Side Decoration", i.e. letting an app have a custom titlebar).

But why would `winit` *default* to CSD? Surely _**custom**_ window decorations are *optional*, right?
Well, [GNOME appears to have forced *the custom mode* on everyone](https://github.com/electron/electron/issues/27522#issuecomment-768542543).

Not just that but [some of them even posted to KDE venues proposing KDE *also do the same*](https://forum.kde.org/viewtopic.php?f=305&t=168320) (i.e. remove perfectly good WM-side window decorations and force everyone to use custom "CSD" ones).

My only hope is that I can turn this stuff off and maybe `cargo-deny` will then be happy but I have doubts.

examples/runners/wgpu/src/graphics.rs
```diff
@@ -104,7 +104,8 @@ async fn run(
         format: preferred_format,
         width: size.width,
         height: size.height,
-        present_mode: wgpu::PresentMode::Mailbox,
+        present_mode: wgpu::PresentMode::AutoNoVsync,
```
eddyb, line 107: This change was found by @fu5ha because macOS apparently doesn't support `Mailbox` (did this *ever* work? I think `main` might be broken too).

ValorZard:
All the stuff that was blocking this PR got merged, so does that mean that wgpu can be updated now?

eddyb:
> All the stuff that was blocking this PR got merged, so does that mean that wgpu can be updated now?

@ValorZard To be clear, none of my links above have changed status since I posted them. I was linking *already merged* PRs, some of them just to describe why certain changes where needed.

The only blocker link AFAICT is there's no new *release* containing e.g. the `crossfont`->`ab_glyph` change (https://github.com/rust-windowing/winit/commit/f9b41fd8194903ac9d9db2728d39af2afa812f11 - note how GitHub doesn't list any tags as containing it).

You can also see it's still under [the "Unreleased" section in the CHANGELOG](https://github.com/rust-windowing/winit/blob/bdcbd7d1f99464fe945596b43a922a9632456e44/CHANGELOG.md#unreleased) right now:
> * On Wayland, `wayland-csd-adwaita` now uses `ab_glyph` instead of `crossfont` to render the title for decorations.

We're basically waiting for [`winit` to release `0.28`](https://github.com/rust-windowing/winit/milestone/8), which I'm not seeing a timeline for.

---

For now, this update is not *that* important, at least not for Naga, until these issues are fixed:
* https://github.com/gfx-rs/naga/issues/1575
* https://github.com/gfx-rs/naga/issues/1977#issuecomment-1318334723

We should just make it possible to use the raw SPIR-V (i.e. bypassing Naga entirely) until Naga improves.

examples/runners/wgpu/src/graphics.rs
eddyb: Note that in the latest version of this PR, I've changed it to `AutoVsync` because `AutoNoVsync` was doing excessive redraws (thousands of frames per second).

eddyb:
[`winit 0.28`](https://github.com/rust-windowing/winit/releases/tag/v0.28.0) has since been released (and since then [some further patch releases too](https://github.com/rust-windowing/winit/releases)).

I've further updated `winit`, `wgpu` (and a few other miscellaneous crates), but also changed how we handle `wgpu`'s `naga` dependency, so that we don't have to update `wgpu` to get newer `naga`.

That was in part motivated by a desire to maybe try fixing `naga` bugs ourselves, and being on the the latest from git would make that much easier - but also we could try out e.g. `wgpu` raytracing support, before it lands.

eddyb:
Hmpf, Android build still fails, now with:
```
     = note: ld: error: duplicate symbol: ANativeActivity_onCreate
            >>> defined at lib.rs:159 (examples/runners/wgpu/src/lib.rs:159)
            >>>            /home/runner/work/rust-gpu/rust-gpu/target/aarch64-linux-android/debug/deps/example_runner_wgpu.8xz6nj4o9c8umeb.rcgu.o:(ANativeActivity_onCreate)
            >>> defined at glue.rs:736 (src/native_activity/glue.rs:736)
            >>>            android_activity-6c8fa8b2535fb42a.android_activity.84876fb1-cgu.0.rcgu.o:(.text.ANativeActivity_onCreate+0x0) in archive /home/runner/work/rust-gpu/rust-gpu/target/aarch64-linux-android/debug/deps/libandroid_activity-6c8fa8b2535fb42a.rlib
```

It's possible that me adding this feature-enabling dependency (to fix previous CI errors):
```toml
android-activity = { version = "0.4.1", features = ["native-activity"] }
```
is conflicting with... `ndk-glue`, maybe?

It's conflicting between `android-activity`'s `native_activity::glue` and our:
```rust
#[cfg_attr(target_os = "android", ndk_glue::main(backtrace = "on"))]
pub fn main
```

So it's probably likely that the way Android "Native Activity" apps are supposed to be built has changed a lot.

MarijnS95:
@eddyb correct, `ndk-glue` has been deprecated and `winit` has a migration guide telling you exactly what to change to switch over to `android-activity`: https://github.com/rust-windowing/winit#converting-from-ndk-glue-to-android-activity

eddyb:
Got Android building (and WASM for that matter, but it doesn't work, I get a lot of errors I haven't looked into).

This is what I can get out of `cargo apk run -p example-runner-wgpu --lib --target x86_64-linux-android`, after starting the Android `emulator` with `-no-snapshot -feature Vulkan,GLDirectMem`:



I'm not sure what's wrong with the texture format, but [the `na-winit-wgpu` example](https://github.com/rust-mobile/rust-android-examples/blob/41e72ca9de45e2add4776ec004f916ad979cf781/na-winit-wgpu) has the same problem.

eddyb:
Opened a separate PR with the the addition I was using to test the WASM path:
* #1016

Which will have to wait until the massive problem with `wgpu::backend::web` is fixed:
* https://github.com/gfx-rs/wgpu/issues/3430

So this PR is truly unblocked, just needs a review now!

fu5ha (approved):

PR #926, opened 2022-10-14
Migration from register_attr to register_tool

oisyn:
We need this because the `remove_attr` feature was removed from the Rust nightly early September

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Changelog
+# `rust-gpu` Changelog
 
 All notable changes to this project will be documented in this file.
 
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed 🛠️
 
-- BREAKING: Migrated from `register_attr` to `register_tool`. [More information](docs/src/migration-to-register-tool.md).
+- 🚨BREAKING🚨 Migrated from `register_attr` to `register_tool`. [More information](docs/src/migration-to-register-tool.md).
 
 ## [0.4.0-alpha.15]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,22 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Changed 🛠️
+
+- BREAKING: Migrated from `register_attr` to `register_tool`. [More information](docs/src/migration-to-register-tool.md).
+
+## [0.4.0-alpha.15]
+
+### Added ⭐
+
+- Build-time check for nightly toolchain version to provide user-friendly error messages.
+
+### Changed 🛠️
+
+- Updated rust toolchain to `nightly-2022-08-29`.
--- a/crates/rustc_codegen_spirv/src/symbols.rs
+++ b/crates/rustc_codegen_spirv/src/symbols.rs
@@ -1,7 +1,7 @@
 use crate::attr::{Entry, ExecutionModeExtra, IntrinsicType, SpirvAttribute};
 use crate::builder::libm_intrinsics;
 use rspirv::spirv::{BuiltIn, ExecutionMode, ExecutionModel, StorageClass};
-use rustc_ast::ast::{Attribute, Lit, LitIntType, LitKind, NestedMetaItem};
+use rustc_ast::ast::{AttrKind, Attribute, Lit, LitIntType, LitKind, NestedMetaItem};
 use rustc_data_structures::fx::FxHashMap;
 use rustc_span::symbol::{Ident, Symbol};
 use rustc_span::Span;
@@ -16,6 +16,7 @@ pub struct Symbols {
     // Used by `is_blocklisted_fn`.
     pub fmt_decimal: Symbol,
 
+    pub rust_gpu: Symbol,
     pub spirv: Symbol,
     pub spirv_std: Symbol,
     pub libm: Symbol,
@@ -373,6 +374,7 @@ impl Symbols {
         Self {
             fmt_decimal: Symbol::intern("fmt_decimal"),
 
+            rust_gpu: Symbol::intern("rust_gpu"),
             spirv: Symbol::intern("spirv"),
             spirv_std: Symbol::intern("spirv_std"),
             libm: Symbol::intern("libm"),
@@ -405,22 +407,32 @@ impl Symbols {
 // FIXME(eddyb) find something nicer for the error type.
 type ParseAttrError = (Span, String);
 
+fn attr_is_spirv(a: &Attribute, sym: &Symbols) -> bool {
+    if let AttrKind::Normal(ref normal) = a.kind {
+        let s = &normal.item.path.segments;
+        s.len() == 2 && s[0].ident.name == sym.rust_gpu && s[1].ident.name == sym.spirv
+    } else {
+        false
+    }
+}
+
 // FIXME(eddyb) maybe move this to `attr`?
 pub(crate) fn parse_attrs_for_checking<'a>(
     sym: &'a Symbols,
     attrs: &'a [Attribute],
 ) -> impl Iterator<Item = Result<(Span, SpirvAttribute), ParseAttrError>> + 'a {
     attrs.iter().flat_map(move |attr| {
-        let (whole_attr_error, args) = if !attr.has_name(sym.spirv) {
+        let (whole_attr_error, args) = if !attr_is_spirv(attr, sym) && !attr.has_name(sym.spirv) {
             // Use an empty vec here to return empty
             (None, Vec::new())
         } else if let Some(args) = attr.meta_item_list() {
+            eprintln!("{:?}", args);
             (None, args)
         } else {
             (
                 Some(Err((
                     attr.span,
-                    "#[spirv(..)] attribute must have at least one argument".to_string(),
+                    "#[rust_gpu::spirv(..)] attribute must have at least one argument".to_string(),
                 ))),
                 Vec::new(),
             )
@@ -430,7 +442,11 @@ pub(crate) fn parse_attrs_for_checking<'a>(
             .chain(args.into_iter().map(move |ref arg| {
                 let span = arg.span();
                 let parsed_attr = if arg.has_name(sym.descriptor_set) {
-                    SpirvAttribute::DescriptorSet(parse_attr_int_value(arg)?)
+                    eprintln!("asdfjkhasdlfkjhasldkfsdf");
+                    eprintln!("{:?}", arg);
+                    let my_int = parse_attr_int_value(arg)?;
+                    eprintln!("{}", my_int);
+                    SpirvAttribute::DescriptorSet(my_int)
                 } else if arg.has_name(sym.binding) {
                     SpirvAttribute::Binding(parse_attr_int_value(arg)?)
                 } else if arg.has_name(sym.input_attachment_index) {
@@ -440,10 +456,10 @@ pub(crate) fn parse_attrs_for_checking<'a>(
                         Some(i) => i,
                         None => {
                             return Err((
-                                span,
-                                "#[spirv(..)] attribute argument must be single identifier"
-                                    .to_string(),
-                            ));
+                            span,
+                            "#[rust_gpu::spirv(..)] attribute argument must be single identifier"
+                                .to_string(),
+                        ));
                         }
                     };
                     sym.attributes.get(&name.name).map_or_else(
@@ -498,15 +514,15 @@ fn parse_local_size_attr(arg: &NestedMetaItem) -> Result<[u32; 3], ParseAttrErro
         }
         Some(tuple) if tuple.is_empty() => Err((
             arg.span,
-            "#[spirv(compute(threads(x, y, z)))] must have the x dimension specified, trailing ones may be elided".to_string(),
+            "#[rust_gpu::spirv(compute(threads(x, y, z)))] must have the x dimension specified, trailing ones may be elided".to_string(),
         )),
         Some(tuple) if tuple.len() > 3 => Err((
             arg.span,
-            "#[spirv(compute(threads(x, y, z)))] is three dimensional".to_string(),
+            "#[rust_gpu::spirv(compute(threads(x, y, z)))] is three dimensional".to_string(),
         )),
         _ => Err((
             arg.span,
-            "#[spirv(compute(threads(x, y, z)))] must have 1 to 3 parameters, trailing ones may be elided".to_string(),
+            "#[rust_gpu::spirv(compute(threads(x, y, z)))] must have 1 to 3 parameters, trailing ones may be elided".to_string(),
         )),
     }
 }
@@ -514,7 +530,7 @@ fn parse_local_size_attr(arg: &NestedMetaItem) -> Result<[u32; 3], ParseAttrErro
 // for a given entry, gather up the additional attributes
 // in this case ExecutionMode's, some have extra arguments
 // others are specified with x, y, or z components
-// ie #[spirv(fragment(origin_lower_left))] or #[spirv(gl_compute(local_size_x=64, local_size_y=8))]
+// ie #[rust_gpu::spirv(fragment(origin_lower_left))] or #[rust_gpu::spirv(gl_compute(local_size_x=64, local_size_y=8))]
 fn parse_entry_attrs(
     sym: &Symbols,
     arg: &NestedMetaItem,
@@ -550,7 +566,7 @@ fn parse_entry_attrs(
                                 return Err((
                                     attr_name.span,
                                     String::from(
-                                        "`#[spirv(compute(threads))]` may only be specified once",
+                                        "`#[rust_gpu::spirv(compute(threads))]` may only be specified once",
                                     ),
                                 ));
                             }
@@ -616,7 +632,7 @@ fn parse_entry_attrs(
                             return Err((
                                 attr_name.span,
                                 format!(
-                                    "#[spirv({}(..))] unknown attribute argument {}",
+                                    "#[rust_gpu::spirv({}(..))] unknown attribute argument {}",
                                     name.name.to_ident_string(),
                                     attr_name.name.to_ident_string()
                                 ),
@@ -627,7 +643,7 @@ fn parse_entry_attrs(
                     return Err((
                         attr_name.span,
                         format!(
-                            "#[spirv({}(..))] unknown attribute argument {}",
+                            "#[rust_gpu::spirv({}(..))] unknown attribute argument {}",
                             name.name.to_ident_string(),
                             attr_name.name.to_ident_string()
                         ),
@@ -637,7 +653,7 @@ fn parse_entry_attrs(
                 return Err((
                     arg.span(),
                     format!(
-                        "#[spirv({}(..))] attribute argument must be single identifier",
+                        "#[rust_gpu::spirv({}(..))] attribute argument must be single identifier",
                         name.name.to_ident_string()
                     ),
                 ));
@@ -660,7 +676,7 @@ fn parse_entry_attrs(
                 return Err((
                     arg.span(),
                     String::from(
-                        "The `threads` argument must be specified when using `#[spirv(compute)]`, `#[spirv(mesh_nv)]` or `#[spirv(task_nv)]`",
+                        "The `threads` argument must be specified when using `#[rust_gpu::spirv(compute)]`, `#[rust_gpu::spirv(mesh_nv)]` or `#[rust_gpu::spirv(task_nv)]`",
                     ),
                 ));
             }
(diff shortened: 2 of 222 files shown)


expenses:
:( that sounds annoying. Is it not possible to have something like `use rust_gpu::spirv;`?

oisyn:
@expenses I don't think that's possible for tool attributes. See [Rust playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=6259d6b2385d02c976b4c4f1c3665153)

I was just discussing with @eddyb, we could probably support a `proc_macro_attribute` that converts preludes all `spirv` attributes with `rust_gpu::`. Basically the inverse of what `spirv` was already doing when not compiling for spirv 😉

I converted this PR to draft as I wanted to include a changelog.md, but I think I will address this as well.

eddyb (changes requested):
Nitpick: `[#` everywhere should be replaced with `#[`.
<sub><sub>(the `#` is the attribute syntax, `[` is merely a "container" - this is similar to how some languages have `@foo` and `@bar(123)` attributes but that `@` is `#` in Rust and brackets exist to disambiguate something like `#[allow(overflowing_literals)] (x + 200i8)` which without the brackets would me ambiguous as to whether the parens are attached to `allow` - the languages that don't need that tend to be very limited in *where* you can apply attributes - anyway Rust can look like languages with `[foo]` attribute syntax but AFAIK it's closer to the `@foo` languages despite the aesthetics...)</sub></sub>

oisyn:
Yeah I keep mistyping that. I think I'll also add a attribute called `spriv` just to take care of that common typo 😋

tests/src/main.rs
```diff
@@ -123,8 +123,8 @@ impl Runner {
                 "--crate-type dylib",
                 "-Zunstable-options",
                 "-Zcrate-attr=no_std",
-                "-Zcrate-attr=feature(register_attr,asm_const,asm_experimental_arch)",
-                "-Zcrate-attr=register_attr(spirv)",
+                "-Zcrate-attr=feature(register_tool,asm_const,asm_experimental_arch)",
+                "-Zcrate-attr=register_tool(rust_gpu)",
```
eddyb, line 127: As I was saying elsewhere, we need to just copy this into `spirv-builder` so that users don't need to use themselves (tho we should do only the `register_tool` feature, not `asm_const` - not sure about `asm_experimental_arch`, but if it's needed to use `asm!` *at all*, we might want to inject it anyway)

VZout (approved):

fu5ha (approved):
I didn't look this over exhaustively but the gist of the change is a 👍 from me

examples/shaders/sky-shader/src/lib.rs
```diff
@@ -3,18 +3,16 @@
 #![cfg_attr(
     target_arch = "spirv",
     no_std,
-    feature(register_attr, lang_items),
-    register_attr(spirv)
+    feature(register_tool, lang_items),
+    register_tool(rust_gpu)
```
eddyb, line 7:
```suggestion
```
(the tool module provided by `spirv-builder` and `lang_items` hasn't been needed in a while)

Also, we should make sure we run `cargo +stable check -p sky-shader -p mouse-shader` *at least* in CI (well, it passes before this PR, thanks to `cfg_attr`, but still).

examples/shaders/simplest-shader/src/lib.rs
```diff
@@ -1,16 +1,14 @@
 #![cfg_attr(
     target_arch = "spirv",
     no_std,
-    feature(register_attr),
-    register_attr(spirv)
+    feature(register_tool),
+    register_tool(rust_gpu)
```
eddyb, line 5:
```suggestion
```

examples/shaders/reduce/src/lib.rs
```diff
@@ -1,17 +1,16 @@
 #![cfg_attr(
     target_arch = "spirv",
     no_std,
-    feature(register_attr),
-    register_attr(spirv)
+    feature(register_tool),
+    register_tool(rust_gpu)
```
eddyb, line 5:
```suggestion
```

examples/shaders/mouse-shader/src/lib.rs
```diff
@@ -1,24 +1,22 @@
 #![cfg_attr(
     target_arch = "spirv",
     no_std,
-    feature(register_attr),
-    register_attr(spirv)
+    feature(register_tool),
+    register_tool(rust_gpu)
```
eddyb, line 5:
```suggestion
```

examples/shaders/compute-shader/src/lib.rs
```diff
@@ -1,16 +1,14 @@
 #![cfg_attr(
     target_arch = "spirv",
-    feature(register_attr),
-    register_attr(spirv),
+    feature(register_tool),
+    register_tool(rust_gpu),
```
eddyb, line 4:
```suggestion
```

crates/spirv-std/src/ray_tracing.rs
```diff
@@ -1,5 +1,6 @@
 //! Ray-tracing data types
 use crate::vector::Vector;
+
```
eddyb, line 3: Stray edit? (seems fine either way)

crates/spirv-std/src/lib.rs
```diff
@@ -93,8 +93,9 @@
 //! Core functions, traits, and more that make up a "standard library" for SPIR-V for use in
 //! rust-gpu.
 
-#[cfg_attr(not(target_arch = "spirv"), macro_use)]
+#[macro_use]
 pub extern crate spirv_std_macros as macros;
```
eddyb, line 97: What happens if you remove `pub` from this?

crates/spirv-std/src/lib.rs
```diff
@@ -1,15 +1,15 @@
-#![no_std]
+//#![no_std]
```
eddyb, line 1:
```suggestion
#![no_std]
```

crates/spirv-std/shared/src/lib.rs
```diff
@@ -1,5 +1,4 @@
 //! Small shared crate, to share definitions between `spirv-std`
 //! and `spirv-std-macros`.
-#![no_std]
```
eddyb, line 3: See all the other comments, this needs to be added back in.

crates/spirv-builder/src/lib.rs
```diff
@@ -420,6 +420,9 @@ fn invoke_rustc(builder: &SpirvBuilder) -> Result<PathBuf, SpirvBuilderError> {
     let mut rustflags = vec![
         format!("-Zcodegen-backend={}", rustc_codegen_spirv.display()),
         "-Csymbol-mangling-version=v0".to_string(),
+        "-Zcrate-attr=no_std".to_string(),
```
eddyb, line 423:
```suggestion
```
(as mentioned in my other review comments, this should be kept in the source of the crates, not here, because it's part of Rust not an implementation detail of Rust-GPU)

crates/spirv-std/macros/src/lib.rs
```diff
@@ -171,7 +182,6 @@ pub fn spirv(_attr: TokenStream, item: TokenStream) -> TokenStream {
         .collect::<proc_macro2::TokenStream>()
         .into()
 }
-
```
eddyb, line 174: Stray edit?

crates/spirv-std/macros/src/lib.rs
```diff
@@ -153,7 +160,11 @@ pub fn spirv(_attr: TokenStream, item: TokenStream) -> TokenStream {
                                 && matches!(group.stream().into_iter().next(), Some(TokenTree::Ident(ident)) if ident == "spirv")
                                 && matches!(sub_tokens.last(), Some(TokenTree::Punct(p)) if p.as_char() == '#') =>
```
eddyb, line 161: Hmm this side feels very sketchy but I guess we can't use the proc macro attribute directly on e.g. function arguments?

crates/spirv-std/macros/src/lib.rs
```diff
@@ -138,9 +138,16 @@ pub fn Image(item: TokenStream) -> TokenStream {
...
+/// #[cfg_attr(target_arch="spirv", rust_gpu::spirv(..))].
 #[proc_macro_attribute]
-pub fn spirv(_attr: TokenStream, item: TokenStream) -> TokenStream {
-    let mut tokens = Vec::new();
+pub fn spirv(attr: TokenStream, item: TokenStream) -> TokenStream {
+    let mut tokens: Vec<TokenTree> = Vec::new();
+
+    // prepend with #[rust_gpu::spirv(..)]
+    let attr: proc_macro2::TokenStream = attr.into();
+    tokens.extend(quote! { #[cfg_attr(target_arch="spirv", rust_gpu::spirv(#attr))] });
```
eddyb, line 149: Huh, [this is not what I expected](https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros) (I think the plan used to be exposing which kind of delimiters were used with the attribute but we gave up on that?).

In that case I suppose the code looks fine? (a bit surprised you need a `proc_macro2::TokenStream`, do the builtin ones not implement the trait `quote!` wants?)

eddyb (dismissed):

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -405,13 +407,22 @@ impl Symbols {
...
 type ParseAttrError = (Span, String);
 
+fn attr_is_spirv(a: &Attribute, sym: &Symbols) -> bool {
+    if let AttrKind::Normal(ref normal) = a.kind {
+        let s = &normal.item.path.segments;
+        s.len() == 2 && s[0].ident.name == sym.rust_gpu && s[1].ident.name == sym.spirv
+    } else {
+        false
+    }
+}
```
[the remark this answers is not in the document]
oisyn: That function is already terribly unreadable. I didn't want to mess it up even more by adding a multi-line if statement.

Agreed on the cath-all thing, we should throw an error there.

tests/src/main.rs
oisyn: It *is* in SpirvBuilder (no-std and register_tool), but the tests don't use SpirvBuilder.

tests/ui/spirv-attr/invalid-target.rs
```diff
@@ -30,6 +30,8 @@
 //   * 6 on `_entry_param`
 //   * 1 on `_closure`
 
+use spirv_std::spirv;
+
```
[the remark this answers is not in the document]
oisyn: I will.
I don't understand why this test is failing on CI though, it passes locally.

tests/ui/lang/f32/signum.rs
```diff
@@ -1,3 +1,5 @@
+use spirv_std::spirv;
```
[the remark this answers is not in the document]
oisyn: Well that's the thing. I didn't want to manually change like 120 files so I wrote a simple shell script that prepended that line everywhere 😉. I'll see what I can do.

tests/src/main.rs
```diff
@@ -329,6 +327,10 @@ fn rust_flags(codegen_backend_path: &Path) -> String {
         "-Cembed-bitcode=no",
         &format!("-Ctarget-feature=+{}", target_features.join(",+")),
         "-Csymbol-mangling-version=v0",
+        "-Zcrate-attr=no_std",
+        "-Zcrate-attr=feature(register_tool)",
+        "-Zcrate-attr=register_tool(rust_gpu)",
+        "-Zcrate-attr=allow(unused_attributes)",
```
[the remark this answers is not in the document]
oisyn: The problem here is `spirv-std`, which is `no-std` always. I can make it a `cfg-attr` for `spirv` only. But some random deps might also have `no-std` (`glam` did, for example), that's why I added the `unused_attributes` because the warnings got pretty annoying and you can't always solve them yourself.

crates/spirv-std/macros/src/lib.rs
oisyn: This isn't my code though. But yeah, that's why we had the proc macro attribute implementation in the first place, to strip all those illegal uses of the attribute.

Also it only matches literal `#[spirv()]`. If someone does `#[spirv_std::spirv()]` we don't catch that.

crates/spirv-std/macros/src/lib.rs
oisyn: Yeah `quote!` only works on `proc_macro2` stuff. Really weird.

oisyn: Wait, what exactly did you not expect?

eddyb: > Wait, what exactly did you not expect?

That from `#[attr_name(inputs)]`, the proc macro gets only `inputs`, not `(inputs)` - my understanding was that we would always include the parens used with the attribute, to allow things like `#[attr_name = "value"]` to receive `= "value"` - I'm likely just confused though.

crates/spirv-std/macros/src/lib.rs
eddyb: > Also it only matches literal `#[spirv()]`. If someone does `#[spirv_std::spirv()]` we don't catch that.

Right, this is annoying, but it doesn't look like we can do better for now - I was able to find a crate on `play.rust-lang.org` to test with and yeah: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=b23176b6523a4ae675fd7e65b1fb6d4f

(even on nightly there's no feature-gate suggested so I think it's just not supported at all)

oisyn:
Waiting for #929 first

CHANGELOG.md
```diff
@@ -0,0 +1,22 @@
...
+
+## [0.4.0-alpha.15]
+
+### Added ⭐
+
+- Build-time check for nightly toolchain version to provide user-friendly error messages.
+
+### Changed 🛠️
+
+- Updated rust toolchain to `nightly-2022-08-29`.
```
eddyb, line 22: Opened an issue to track expanding this with older details:
* #930

tests/ui/spirv-attr/invalid-target.rs
```diff
@@ -30,9 +30,9 @@
 //   * 6 on `_entry_param`
 //   * 1 on `_closure`
 
-//use spirv_std::spirv;
+// NOTE(shesp) Directly using `#[rust_gpu::spirv(...)]` because macro attributes are invalid in most contexts
```
eddyb, line 33: Perfect!

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -411,20 +413,44 @@ pub(crate) fn parse_attrs_for_checking<'a>(
...
+                                attr.span,
+                                "#[rust_gpu::spirv(..)] attribute must have at least one argument"
+                                    .to_string(),
+                            ))),
+                            Vec::new(),
+                        )
+                    }
+                } else {
+                    // #[...] but not #[rust_gpu ...]
+                    (None, Vec::new())
```
eddyb, line 448: Heh, like the short comments, makes a decision tree much easier to visually scan! (this is also why I was kinda suggesting using `match` since sometimes it can help flatten some of these things - tho here it's such a mess of conditions that we'll need more than just `match` to make it nice)

crates/spirv-std/src/lib.rs
```diff
@@ -93,8 +91,9 @@
 //! Core functions, traits, and more that make up a "standard library" for SPIR-V for use in
 //! rust-gpu.
 
-#[cfg_attr(not(target_arch = "spirv"), macro_use)]
+#[macro_use]
 pub extern crate spirv_std_macros as macros;
```
eddyb, line 95: You marked my previous comment asking about the `pub` here as resolved, but I don't remember what the explanation was, sorry - it's just that `spirv_std::macros::...` is used elsewhere, right?

docs/src/migration-to-register-tool.md
```diff
@@ -0,0 +1,45 @@
...
+
+This document applies to [PR#926](https://github.com/EmbarkStudios/rust-gpu/pull/926)
+
+## What happened
+
+In a [recent nightly Rust update](https://github.com/rust-lang/rust/commit/76dd5c58a011bb734ad5b8e96fc560374893bc8f), the `register_attr` feature was removed in favor of `register_tool`. Unfortunately, rust-gpu made use of this feature to register the `spirv` attribute.
+
+## What does this mean for you as a shader maintainer
+
+You'll need to import the `spirv` proc macro attribute from `spirv-std` in order for the rust compiler:
```
eddyb, line 11: Sentence looks accidentally cut-off.

docs/src/migration-to-register-tool.md
```diff
@@ -0,0 +1,45 @@
...
+
+```rust
+use spirv_std::spirv;
+```
+
+If your shader code already contains this line but is conditionally only included for non-spirv builds, like so:
```
eddyb, line 17:
```suggestion
If your shader code already contains this line but is conditionally only included for non-SPIR-V builds, like so:
```
or alternatively this:
```
non-`spirv` builds
```

docs/src/migration-to-register-tool.md
```diff
@@ -0,0 +1,45 @@
...
+
+```rust
+#[cfg(not(target_arch = "spirv"))]
+use spirv_std::spirv;
+```
+
+please remove the conditional attribute.
```
eddyb, line 24:
```suggestion
please remove the conditional attribute (the `#[cfg(...)]` line).
```

docs/src/migration-to-register-tool.md
```diff
@@ -0,0 +1,45 @@
...
+
+For this macro attribute to work correctly, it is important that `spirv` is visible in the global score and you use it like you used it before: `#[spirv(..)]`. An attempt to scope the attribute (such as `#[spirv_std::spirv(..)]`) will confuse the macro and likely fail.
```
eddyb, line 26: We don't need to expand on this much more but to be clear this only applies to weird "leaf" things like function args, fields, etc. (which are replaced *while* transforming a larger thing that triggered the proc macro as usual).
Whole-definition attributes (e.g. the whole entry-point or a whole `struct`) will just work through normal name resolution.

Funny thing about that btw, if you *only* need it for a leaf, you need to do:
```rust
#[spirv()]
fn foo(
    #[spirv(ray_tracing_weirdness)] x: T,
) {}
```
(with the whole-`fn` one only serving to handle the inner ones lol - again, I don't think we have to mention this, since we don't have that many attributes to begin with, and so far such a usecase doesn't exist AFAIK, but I wanted to expand on it "for the record" so to speak)

docs/src/migration-to-register-tool.md
```diff
@@ -0,0 +1,45 @@
...
+
+Instead, we opted to implement a proc macro attribute called `spirv` instead[^1]. This macro attribute scans the item it is applied to, and translates any `#[spirv(..)]` it finds into `#[rust_gpu::spirv(..)]` which will be subsequently handled by the codegen backend. Because it is now a proc macro attribute exported from `spirv_std`, you need to do `use spirv_std::spirv` to make it globally visible in your crate. ***Note that we recommend using the `spirv` proc macro attribute itself rather than the `rust_gpu::spirv` attribute it translates to, as the latter is subject to change.***
+
+We've also added the `feature(register_tool)` and `register_tool(rust_gpu)` crate attributes by default when compiling through `SpirvBuilder`. This will silence any error that you would otherwise get for applying a `rust_gpu` scoped attribute.
+
+[^1]: This is not entirely true. In reality, the `spirv` proc macro attribute already existed, but only for non-spirv builds. It was used to turn the `#[spirv(..)]` attribute into a no-op. The proc macro is now used on all platforms, and it emits `#[cfg_attr(target_arch="spirv", rust_gpu::spirv(..))]` for each usage of `#[spirv(..)]`.
```
eddyb, line 45: :100: :100: :100:
<sub>(I wasn't thinking we needed this but now that you wrote it I really like it heh, setting the new standard basically)</sub>

eddyb (commented):
These are the only files I didn't check off as "viewed":
```js
$$('.js-reviewed-checkbox').filter(x=>!x.checked).map(x=>x.parentNode.parentNode.path.value).join('\n')
```
```
tests/ui/arch/all.rs
tests/ui/arch/any.rs
tests/ui/arch/convert_u_to_acceleration_structure_khr.rs
tests/ui/arch/demote_to_helper_invocation.rs
tests/ui/arch/emit_stream_vertex.rs
tests/ui/arch/emit_vertex.rs
tests/ui/arch/end_primitive.rs
tests/ui/arch/end_stream_primitive.rs
tests/ui/arch/execute_callable.rs
tests/ui/arch/ignore_intersection_khr.rs
tests/ui/arch/kill.rs
tests/ui/arch/report_intersection_khr.rs
tests/ui/arch/terminate_ray_khr.rs
tests/ui/arch/trace_ray_khr.rs
tests/ui/image/issue_527.rs
```

But looks like GH decided to merge it *anyway*? :/

crates/spirv-std/src/lib.rs
oisyn: Oh I did this because the things in `spirv-std` now need `spirv_std::spirv` to be visible for all platforms. This way I didn't need to write `use spirv_std::spirv;` in every single file.

docs/src/migration-to-register-tool.md
oisyn: If we do end up having a case where this applies (I don't think we do currently?), I'd suggest supporting `#[spirv]` without the parens.

eddyb: > I'd suggest supporting `#[spirv]` without the parens.

I don't think we have a say in the matter (i.e. both shouldwork), and you likely can't tell them apart (since that's the proc macro, not the tool module one).

PR #928, opened 2022-10-17
Update spirv-tools

Jake-Shadle:
- Update spirv-tools
- Update spirv-tools installs to latest-ish

Updates spirv-tools to 2022.4 (and 2022.3 and 2022.2 because I have been slacking).

---

~~**EDIT**(@eddyb): marked as draft after rebasing onto (and blocking on):~~
* [x] ~~#933~~

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,12 +28,12 @@ jobs:
       # (which should be in the `YYYYMMDD` format and appear in paths) in these
       # listings (NB: they're limited to 1000 results and may need adjustment):
       # https://storage.googleapis.com/spirv-tools/?list-type=2&start-after=artifacts/prod/graphics_shader_compiler/spirv-tools/linux-clang-release/continuous/1700
-      # https://storage.googleapis.com/spirv-tools/?list-type=2&start-after=artifacts/prod/graphics_shader_compiler/spirv-tools/macos-clang-release/continuous/1700
-      # https://storage.googleapis.com/spirv-tools/?list-type=2&start-after=artifacts/prod/graphics_shader_compiler/spirv-tools/windows-msvc-2017-release/continuous/1700
+      # https://storage.googleapis.com/spirv-tools/?list-type=2&start-after=artifacts/prod/graphics_shader_compiler/spirv-tools/macos-clang-release/continuous/1872
+      # https://storage.googleapis.com/spirv-tools/?list-type=2&start-after=artifacts/prod/graphics_shader_compiler/spirv-tools/windows-msvc-2017-release/continuous/1848
       spirv_tools_version: "20220414"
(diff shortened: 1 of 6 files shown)


eddyb:
NB: the validation messages are busted: https://github.com/KhronosGroup/SPIRV-Tools/pull/4831#discussion_r997310010

---

But the rules are real, and we are breaking them, I would suggest just reading the spec:
* [`VUID-StandaloneSpirv-Flat-04744`](https://registry.khronos.org/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-StandaloneSpirv-Flat-04744): `Flat` mandatory on `Fragment` `Input`s other than `f32`
  * they say "integer or double-precision floating-point" but presumably in reality the non-`Flat` default is an interpolation that is only supported for `f32` (and vectors with `f32` elements, right?)
* [`VUID-StandaloneSpirv-Flat-06201`](https://registry.khronos.org/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-StandaloneSpirv-Flat-06201): `Flat`/`NoPerspective`/`Sample`/`Centroid` illegal on `Fragment` `Output`s
* [`VUID-StandaloneSpirv-Flat-06202`](https://registry.khronos.org/vulkan/specs/1.2-extensions/html/vkspec.html#VUID-StandaloneSpirv-Flat-06202): `Flat`/`NoPerspective`/`Sample`/`Centroid` illegal on `Vertex` `Input`s

eddyb:
Now that the whole deal in https://github.com/EmbarkStudios/rust-gpu/pull/928#issuecomment-1281199133 has been fixed, I've bumped the `spirv-tools` version further so we should get better errors on CI now.

---

Also moving https://github.com/EmbarkStudios/rust-gpu/pull/928#discussion_r997334264 here to be able to hide it:
> Hah, you bumped all of these by 116 CI builds, so they *are* in lock-step, just with an offset from their initial setup at the very start? We could use that to our advantage and write a script that finds a set of 3 for a specific date - once we have e.g. the macOS CI build ID, we can derive the others from it and query the bucket for e.g. `".../${macos_id - 12}/${date}-*/install.tgz"`.
>
> It's a bit messy but as long as we can do 3-4 `curl` requests in CI we can avoid hardcoding anything more than that one CI build ID (and even that we might be able to derive from the date if it's a reliably-daily build...).

(i.e. I thought maybe we could automate it - however it is annoying to do this on 3 different platforms when I only trust one of them to have the right tools for quickly script this, without e.g. reaching for Python)

eddyb:
Blocking (and marking as draft) this PR on:
* #933

Which I opened separately to show the effects on tests in isolation, *before* updating `spirv-val`.

eddyb (approved):

eddyb:
I talked to @Jake-Shadle and there's some good ideas around moving away from this model to something that both avoids so many builds of the C++ code *and* avoids dealing with potentially-wildly-different system-installed tooling. (At the very least the system-installed mode could move to linking against the `.a` libraries, which could then be switched from "weird GCP CI artifacts bucket" to our own builds)

For now I want to merge this for the test fixes (when using system-installed SPIRV-Tools on other branches).

PR #929, opened 2022-10-19
Prepare invalid-target test for #926

oisyn:
This PR adds the commented out `use spirv_std::spirv;` to `invalid-target.rs` in preparation for #926

### DIFF
--- a/tests/ui/spirv-attr/invalid-target.rs
+++ b/tests/ui/spirv-attr/invalid-target.rs
@@ -30,6 +30,8 @@
 //   * 6 on `_entry_param`
 //   * 1 on `_closure`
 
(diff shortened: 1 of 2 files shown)


tests/ui/spirv-attr/invalid-target.rs
```diff
@@ -30,6 +30,8 @@
 //   * 6 on `_entry_param`
 //   * 1 on `_closure`
```
eddyb, line 31: Okay good, this accounting (expand to see it above, these are just the last 2 lines of a dozen) is still correct most likely, total error count is still 473.

eddyb (approved):

PR #931, opened 2022-10-19
Changed the location of some `use` statements in tests

oisyn:
See https://github.com/EmbarkStudios/rust-gpu/pull/926#pullrequestreview-1147204065

Basically, not very nice to put the `use spirv_std::spirv;` on the top line before some test comments.


eddyb (approved):
Oh wow I see, most of these needed no imports before!

PR #934, opened 2022-10-26
Update to nightly-2022-09-06

oisyn:
Also updates `glam` to `0.22`. Some tests need to be updated for the removal of `const_mat2`. With the new release of Rust coming out next week I'd like to keep focus on the nightly itself. I'll create an issue for the tests.


tests/ui/arch/all.stderr
```diff
@@ -0,0 +1,7 @@
+error: error:0:0 - Expected operand to be vector bool: All
+  |
+  = note: spirv-val failed
+  = note: module `$TEST_BUILD_DIR/arch/all.stage-id.spv.dir/module`
+
+error: aborting due to previous error
+
```
eddyb, line 7: This (and the `any` one) looks very sus, like something broke? I'll look into it.

eddyb (changes requested):
(only posting this review to avoid accidental merging - I'm also pushing some fixes for the `const_` macros but will then focus on `all`/`any`)

tests/ui/arch/all.stderr
eddyb: Okay this is broken in two different ways:
1. `// build-pass` tests should not pass in combination with "aborting due to previous error"
2. `glam` `0.22` [broke `BVec` types](https://docs.rs/glam/0.22.0/src/glam/bool/bvec2.rs.html#10-13) (lacking the `#[cfg_attr(target_arch = "spirv", repr(simd))]` they had before)

oisyn: Ugh that's annoying

eddyb: For now, to avoid this error, we can disable these:
https://github.com/EmbarkStudios/rust-gpu/blob/097ba40bedd74eeaa296e719ef7e41f2d3d76c23/crates/spirv-std/src/vector.rs#L10-L15

(since the point of these is effectively "has `#[repr(simd)]`")

crates/spirv-std/Cargo.toml
```diff
@@ -12,7 +12,7 @@ bitflags = "1.2.1"
 num-traits = { version = "0.2.14", default-features = false, features = ["libm"] }
 spirv-std-types = { path = "./shared", version = "0.4.0-alpha.15" }
 spirv-std-macros = { path = "./macros", version = "0.4.0-alpha.13" }
-glam = { version = ">=0.17, <=0.21", default-features = false, features = ["libm"], optional = true }
+glam = { version = "=0.22", default-features = false, features = ["libm"], optional = true }
```
eddyb, line 15:
```suggestion
glam = { version = "0.22", default-features = false, features = ["libm"], optional = true }
```
(the reason for the operators before was to get a range, effectively)

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed 🛠️
 
 - 🚨BREAKING🚨 Migrated from `register_attr` to `register_tool`. [More information](docs/src/migration-to-register-tool.md).
+- Updated toolchain to `nightly-2022-09-06`
+- Updated `glam` to `0.22`
+- Removed `glam::BVec` support (they are no longer `#[repl(simd)]` in `glam`, as Rust doesn't support vector of bool types)
```
eddyb, line 15:
```suggestion
- Removed `glam::BVec` support (they are no longer `#[repl(simd)]` in `glam`, as Rust doesn't support SIMD vectors with `bool` elements)
```

tests/ui/arch/any.rs
```diff
@@ -1,9 +1,26 @@
 // build-pass
 
+#![feature(repr_simd)]
+
 use spirv_std::spirv;
+use spirv_std::{scalar::Scalar, vector::Vector};
+/// HACK(shesp). Rust doesn't allow us to declare regular (tuple-)structs containing `bool` members
```
eddyb, line 7:
```suggestion
use spirv_std::{scalar::Scalar, vector::Vector};

/// HACK(shesp). Rust doesn't allow us to declare regular (tuple-)structs containing `bool` members
```
(for consistency with the other test - feel free to just copy-paste between them)

eddyb (dismissed):

eddyb (approved):

Issue #851, opened 2022-02-17
`rustc_codegen_spirv` taking a long time processing my (large) shader


hatoo:
This issue is from Discord conversation https://discordapp.com/channels/750717012564770887/750717499737243679/943879544006381638

### Expected Behaviour

I want to compile [my shader](https://github.com/hatoo/rene) faster.

### Example & Steps To Reproduce

1. clone my repo https://github.com/hatoo/rene
2. `cargo build --release -Z timings`
3. See compile duration

```
Completed rene v0.1.0 build script (run) in 263.2s
```

In console in VSCode and rust-analyzer plugin. It takes more time.

```
Completed rene v0.1.0 build script (run) in 407.3s
```

## System Info

 - Rust: rustc 1.60.0-nightly (1bd4fdc94 2022-01-12)
 - OS: Windows 11
 - CPU: 3950X
 - GPU: RTX2080ti
 - SPIR-V: SPIRV-Tools v2021.4-dev v2021.3-86-g21e3f681

hatoo:
After I added codes to my shader, it takes very longer time to compile.
In [this commit](https://github.com/hatoo/rene/tree/bfcccbb661d97c80cb9783b847f4c802277fa769), it takes about 1000s.
```
Completed rene v0.1.0 build script (run) in 1009.6s
```

And I added one commit to this in [this branch](https://github.com/hatoo/rene/tree/mix).
I've waited for few hours and the build didn't end.

hatoo:
I enabled `-Zself-profile` and build it and I found `rustc` is spending almost all time to link (`link_block_ordering_pass_and_mem2reg`).

## Procedure

1. Enable `-Zself-profile` on spirv-builder. It can be done by like https://github.com/hatoo/rust-gpu/commit/e34bda56b12cc86533e92ae57a754f6ae896e0b1
2. Change `spirv-builder` to use it.
3. `cargo install --git https://github.com/rust-lang/measureme crox flamegraph summarize`
4. `cargo build --release` on `rene`
5. `crox .\rene_shader-xxxx.mm_profdata` and `chrome_profiler.json` will be produced.
6. Open Chrome Developper tool and click the Performance tab. And upload `chrome_profiler.json` by clicking ⬆️ button.

![image](https://user-images.githubusercontent.com/1667746/154833809-92f5a06b-b3dc-46bc-9ea2-d04fe5858bce.png)

eddyb:
[...]

PR #940, opened 2022-11-17
Minimally integrate SPIR-T (replacing structurizer and shuffling passes around).

eddyb:
Please see [the **SPIR-T** repository](https://github.com/EmbarkStudios/spirt) for more information, but to quote the top of its `README`:
> <div align="center">
>
> ### `SPIR-🇹`
> **⋯🢒 🇹arget 🠆 🇹ransform 🠆 🇹ranslate ⋯🢒**
> </div>
>
> **SPIR-🇹** is a research project aimed at exploring shader-oriented IR designs derived from SPIR-V, and producing a framework around such an IR to facilitate advanced compilation pipelines, beyond what existing SPIR-V tooling allows for.

----

This PR integrates it into Rust-GPU in a very limited capacity, in the middle of a sequence of passes operating on SPIR-V, as a `... -> SPIR-V -> SPIR-T -> SPIR-V -> ...` conversion, with the only SPIR-T -> SPIR-T pass being the CFG structurizer (substituting our own structurizer, which we may want to delete?).

That alone is enough to show some compile-time speedups of over 30% in some workloads, while adding only marginal overhead (~0.1%), and it's still being limited by our existing passes (see https://github.com/EmbarkStudios/spirt/issues/6).
<sub>(**TODO**(@eddyb): insert exact timing data from `mouse-shader`, #851, and any other large shaders)</sub>

---

Opening this PR as a draft because there are some discussions that need to be had, i.e.:
* how acceptable is the lossiness we get from the SPIR-T structurizer? (see test changes)
  * the debuginfo we don't guarantee so it's maybe fine? we can also improve it over time
  * `#[spirv(unroll_loops)]` is effectively broken and would need to be rewritten to operate on SPIR-T
* should we just delete code this supersedes? (our old structurizer)
* should we wait for `0.4` to release, so that the SPIR-T dependency can be only `0.5` onward?
  * this would allow more time to refine it before we need to release `spirt 0.1` on crates.io

Review note: the first commit is separate and fixes bugs in the inliner, so that it can *also* run after `mem2reg`, not just before (we end up with two `mem2reg` passes, trying to do as much as possible before, and the rest after, inlining, whereas they it should be interleaved with inlining, as per https://github.com/EmbarkStudios/spirt/issues/6).


fu5ha:
* `#[spirv(unroll_loops)]` is not particularly important -- we don't use it anywhere to my knowledge and gpu shader compilers *should* be able to do this on their own
* the debug info is also not currently used for anything, I'm unaware of any tools that integrate well with it -- we either cross compile parts of gpu code to cpu or use shader disassemblers built into gpu debugging tools.

oisyn:
> should we just delete code this supersedes? (our old structurizer)
Go for it. Its history is on github, having it lingering in the main codebase just confuses things.

> should we wait for 0.4 to release, so that the SPIR-T dependency can be only 0.5 onward?
> this would allow more time to refine it before we need to release spirt 0.1 on crates.io
That is a good question, but to anwer it we first need to know what is required for a proper 0.4, which we should start discussing as well.

eddyb:
Just force-pushed, replacing significant parts of the `linker/inline` commit to fix an edge case I didn't catch before <sub>(inlined calee has `OpPhi`s in the blocks reached from the entry block - likely only possible for the case where a function starts with a loop)</sub>.

OTOH, `linker/test` is still not fully fixed yet (which is what is keeping CI red). If we go with the plan below, fixing the tests that SPIR-T currently implicitly DCEs<sup>1</sup> can be postponed until we delete the non-SPIR-T codepaths.

<sup><sup>1</sup> in SPIR-T, module variables/functions are kept alive by being either directly exported, or (transitively) depended on, by something that itself is exported - exports include linkage-style exports, but also more special-cased features like `OpEntryPoint`s (which are *kind of* "name exports" heh)</sup>

----

Based on all the feedback so far I'm looking at this rough outline of the steps:
1. split prerequisites from this PR into separate PRs (i.e. fixes that pass w/o the SPIR-T switch)
   * this should include `cargo update -p bytemuck` + `deny.toml` changes
2. land SPIR-T integration *disabled* by default, *before* we release 0.4
   * how to control this? if you look for `env` in `crates/rustc_codegen_spirv/src/linker`, you can see a bunch of hacky var names, we could try to clean up that system - at the very least they should all be prefixed
   * if we add something like `SPIRV_BUILDER_RUSTFLAGS` it would be possible to inject some flags that way, though it would be weird to have to use `-C llvm-args=...` (`-C link-args=...` also exists, if we can make use of it)
   * whatever we use to toggle it, should probably also control dumping 
3. maybe duplicate disassembly test outputs (using `compiletest` revisions to run the test twice)
   * i.e. there'd still be one `foo.rs` file <sub>(likely with 2 extra lines to list revisions + the extra flag for the SPIR-T one)</sub> but instead of one `foo.stderr` we'd have e.g. `foo.old.stderr` and `foo.spirt.stderr`
   * should also be possible to dump SPIR-T, à la LLVM's `-print-after-all` (as long as it's clearly labeled as "not the final output, before `spirv-opt`" etc.)
   * might want instead to apply "test with and without SPIR-T" form our wrapper of `compiletest`, to all tests, uniformly - but that doesn't give you the same level of control like explicit "revisions" does, so that's disappointing
     * (not sure if `// only-*` and `// ignore-*` could be used here at all, right now we use them with the target-env component of the target ~~triple~~ quadruple, so that's taken up, but maybe we could have e.g. `spv1.3+spirt` target env values? to get anything to work at all, I mean)
3. release `spirt 0.1.0` on crates.io (see also [corresponding milestone](https://github.com/EmbarkStudios/spirt/milestone/2) - it's mostly docs)
5. release Rust-GPU 0.4 and explicitly mention SPIR-T and how to turn it on for testing
6. turn SPIR-T on by default (or even remove the ability to disable it entirely)
7. release Rust-GPU 0.5
8. (if the old mode is still accessible, this is where we delete it, for 0.6)

eddyb:
Ooops, had to draft again because I forgot to update the CHANGELOG.

oisyn (approved):

PR #941, opened 2022-11-22
Applied workspace inheritance

oisyn:


Cargo.toml
```diff
@@ -39,8 +54,7 @@ incremental = true
 codegen-units = 256
 
 [patch.crates-io]
-spirv-std = { path = "./crates/spirv-std" }
-spirv-std-macros = { path = "./crates/spirv-std/macros" }
 # HACK(eddyb) needed to include this PR (until it's part of a release):
 # * Manishearth/compiletest-rs#249 (for `// only-*` support)
+# [shesp] The commit is part of 0.8 release. We can remove this once we upgrade `compiletest`.
```
eddyb, line 59: Clarification (for everyone else): this is literally just `cargo upgrade compiletest`, it should require no other changes.

Cargo.toml
```diff
@@ -23,6 +23,21 @@ members = [
...
+edition = "2018"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/EmbarkStudios/rust-gpu"
+
+[workspace.dependencies]
+spirv-std = { path = "./crates/spirv-std", version = "=0.4.0-alpha.17" }
+spirv-std-types = { path = "./crates/spirv-std/shared", version = "=0.4.0-alpha.17" }
+spirv-std-macros = { path = "./crates/spirv-std/macros", version = "=0.4.0-alpha.17" }
+spirv-builder = { path = "./crates/spirv-builder", version = "=0.4.0-alpha.17" }
+rustc_codegen_spirv = { path = "./crates/rustc_codegen_spirv", version = "=0.4.0-alpha.17", default-features = false }
```
eddyb, line 38: I think it's a bit confusing to do `default-features = false` here only for this crate, but not e.g. `spirv-builder` - overall, only doing versioning/paths through this mechanism, and leaving `default-features = false` to the individual `Cargo.toml` files, is clearer (and requires fewest amount of changes from the current state of the PR.

If you want to do it:
```suggestion
rustc_codegen_spirv = { path = "./crates/rustc_codegen_spirv", version = "=0.4.0-alpha.17" }
```

crates/spirv-builder/Cargo.toml
```diff
@@ -16,12 +16,12 @@ skip-toolchain-check = ["rustc_codegen_spirv/skip-toolchain-check"]
 watch = ["notify"]
 
 [dependencies]
+rustc_codegen_spirv-types.workspace = true
+# See comment in lib.rs invoke_rustc for why this is here
+rustc_codegen_spirv = { workspace = true }
```
eddyb, line 21: (see the comment on `Cargo.toml` for why this suggestion)
```suggestion
rustc_codegen_spirv = { workspace = true, default-features = false }
```

tests/Cargo.toml
```diff
@@ -14,5 +15,5 @@ use-compiled-tools = ["rustc_codegen_spirv/use-compiled-tools"]
 
 [dependencies]
 compiletest = { version = "0.7.0", package = "compiletest_rs" }
-rustc_codegen_spirv = { path = "../crates/rustc_codegen_spirv", version = "=0.4.0-alpha.17", default-features = false }
+rustc_codegen_spirv = { workspace = true }
```
eddyb, line 18: (see the comment on `Cargo.toml` for why this suggestion)
```suggestion
rustc_codegen_spirv = { workspace = true, default-features = false }
```

eddyb (changes requested):
LGTM, only one issue of consistency.

Cargo.toml
oisyn: That doesn't work. The reason I added `defeault-features=false` in the workspace is because without having it it conflicts with a crate specifying a feature other than the default

oisyn: Wait maybe my wording is not entirely correct, but I did get compile errors when running the tests because of conflicting features in `rustc_codegen_spirv`

eddyb (approved):
Looks like there is a difference with `default-features = false` in workspace vs deps, and `spirv-builder` had to be moved to the workspace too, that's consistent enough for me!

PR #953, opened 2022-11-30
Switch to Rust 2021.

eddyb:
*This PR is built on the following PRs: (**EDIT**: now all of them merged)*
* #949
* #952
  * itself required #950
* #951

---

~~Sadly, this is not ready. The big issue here is that Rust 2021 makes `panic!("...")` go through the same `format_args!("...")` machinery as everything else (i.e. its special-casing was removed - this means that e.g. `panic!("foo: {}")` no long bypasses format string checks, and actually errors).~~

~~While we *should* have the ability to ignore such things, there's enough going on that our panic entry-point special-casing (plus the `fn` param weakening, maybe?) doesn't actually bypass the "zombie" systems enough.~~

~~My best guess so far is that, unlike a string literal constant, `format_args!` actually needs local variables, and the "zombie" system's concept of "dead code" might not include them.~~

~~So far my efforts have focused on improving the granularity of the "zombie" system (hence the other PRs), to better debug such failure modes (when I started, it kept pointing at `ptr::metadata`, which is pretty useless).~~

---

**EDIT**: the new Rust 2021 `panic!` (which always uses `format_args!`) is now supported (by removing simple invocations of `format_args!` feeding directly into a panic entry-point)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed 🛠️
 - [PR#982](https://github.com/EmbarkStudios/rust-gpu/pull/982) updated toolchain to `nightly-2022-12-18`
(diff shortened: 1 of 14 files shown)


eddyb (commented):
Oops, forgot to add a changelog entry, blocking on that.

oisyn:
Oh good point, I also forgot the changelog with the nightly update.

oisyn (approved):

fu5ha (approved):

PR #954, opened 2022-11-30
Update to nightly-2022-10-15

oisyn:
The new `fn_abi` parameter in `BuilderMethods::call()` and `invoke()` scares me a bit. [Here](https://github.com/rust-lang/rust/pull/102551)'s the relevant PR in rust-lang

Tests pass and examples run.

### DIFF
--- a/crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
+++ b/crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
@@ -154,7 +154,7 @@ impl<'tcx> CodegenCx<'tcx> {
             );
         }
         bx.set_span(span);
-        bx.call(entry_func.ty, entry_func, &call_args, None);
+        bx.call(entry_func.ty, None, entry_func, &call_args, None);
         bx.ret_void();
 
         let stub_fn_id = stub_fn.def_cx(self);
--- a/crates/rustc_codegen_spirv/src/target.rs
+++ b/crates/rustc_codegen_spirv/src/target.rs
@@ -1,5 +1,5 @@
 use rspirv::spirv::MemoryModel;
-use rustc_target::spec::{LinkerFlavor, PanicStrategy, Target, TargetOptions};
+use rustc_target::spec::{Cc, LinkerFlavor, PanicStrategy, Target, TargetOptions};
 use spirv_tools::TargetEnv;
 
 const ARCH: &str = "spirv";
@@ -84,7 +84,7 @@ impl SpirvTarget {
         o.dll_suffix = ".spv".into();
         o.dynamic_linking = true;
         o.emit_debug_gdb_scripts = false;
-        o.linker_flavor = LinkerFlavor::Ld;
+        o.linker_flavor = LinkerFlavor::Unix(Cc::No);
         o.panic_strategy = PanicStrategy::Abort;
         o.os = "unknown".into();
         o.env = self.env.to_string().into();
--- a/crates/rustc_codegen_spirv/build.rs
+++ b/crates/rustc_codegen_spirv/build.rs
@@ -10,9 +10,9 @@ use std::process::{Command, ExitCode};
 /// `cargo publish`. We need to figure out a way to do this properly, but let's hardcode it for now :/
 //const REQUIRED_RUST_TOOLCHAIN: &str = include_str!("../../rust-toolchain");
 const REQUIRED_RUST_TOOLCHAIN: &str = r#"[toolchain]
-channel = "nightly-2022-10-01"
+channel = "nightly-2022-10-15"
 components = ["rust-src", "rustc-dev", "llvm-tools-preview"]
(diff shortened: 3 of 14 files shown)


examples/shaders/shared/src/lib.rs
```diff
@@ -40,7 +40,7 @@ pub struct ShaderConstants {
 }
 
 pub fn saturate(x: f32) -> f32 {
-    x.max(0.0).min(1.0)
+    x.clamp(0.0, 1.0)
 }
```
eddyb, line 44: Wait, didn't this use to be weird? Or am I thinking of an interpolation method?
(I really like `clamp`, would've been better if it took `RangeInclusive` because the `..=` syntax is unambiguous, but this looks fine too)

crates/rustc_codegen_spirv/src/target.rs
```diff
@@ -84,7 +84,7 @@ impl SpirvTarget {
         o.dll_suffix = ".spv".into();
         o.dynamic_linking = true;
         o.emit_debug_gdb_scripts = false;
-        o.linker_flavor = LinkerFlavor::Ld;
+        o.linker_flavor = LinkerFlavor::Unix(Cc::No);
```
eddyb, line 87: Interesting, we should revisit this at some point, because we basically have our own custom linker, and we may want to treat parameters to it as "linker flags" etc.

tests/ui/dis/ptr_copy.normal.stderr
```diff
@@ -1,7 +1,7 @@
 error: Cannot memcpy dynamically sized data
-    --> $CORE_SRC/intrinsics.rs:2431:9
+    --> $CORE_SRC/intrinsics.rs:2449:9
      |
-2431 |         copy(src, dst, count)
+2449 |         copy(src, dst, count)
```
eddyb, line 4: Maybe we should use the regex "normalization" feature to erase away these line numbers lolsob.

crates/rustc_codegen_spirv/src/linker/test.rs
```diff
@@ -77,7 +76,6 @@ fn assemble_and_link(binaries: &[&[u8]]) -> Result<Module, String> {
         output_file: None,
         output_dir: None,
         file_loader: None,
-        diagnostic_output: DiagnosticOutput::Raw(Box::new(write_diags)),
```
eddyb, line 80: This breaks the tests, we lost an override and have to rewrite this system, which will block your PR on some of the cleanups I've been doing on the SPIR-T integration side.

eddyb (changes requested):
LGTM, the only addressable comment would be the one about not passing `None` (it might not matter *now* but we wouldn't want weird bugs in the future if we start relying on it).

**EDIT**: oops, you pushed the change in tests, which will *definitely* break at least 4 of them (try running `cargo test -p rustc_codegen_spirv --lib` to see what I mean).

crates/rustc_codegen_spirv/src/linker/test.rs
oisyn: I'm particularly having issues with this. After removing it, RA says it's missing that field (go figure). When I go the definition, the field is in there as well. So I don't understand the compile error in the first place.

oisyn: I'll go back to 2022-10-15 then until this is fixed

examples/shaders/shared/src/lib.rs
oisyn: There was something about `x.step(y)` being intuitively reversed, are you thinking of that?

crates/rustc_codegen_spirv/src/codegen_cx/entry.rs
```diff
@@ -154,7 +154,7 @@ impl<'tcx> CodegenCx<'tcx> {
             );
         }
         bx.set_span(span);
-        bx.call(entry_func.ty, entry_func, &call_args, None);
+        bx.call(entry_func.ty, None, entry_func, &call_args, None);
```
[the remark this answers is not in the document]
eddyb: What happened here? I see thumbs up but also marked as resolved and no change?

eddyb (changes requested):
This looks ready but I'm not sure if you want to pass `Some(entry_fn_abi)` from that `entry.rs` function or not.

examples/shaders/shared/src/lib.rs
eddyb: Nope, pretty sure it was `lerp`.

eddyb:
> EDIT Back to 10-15 because of a blocked issue (`diagnostic_output` being removed from `Config`)

Opened a PR to move us off of `diagnostic_output`:
* #956

oisyn:
Apparently forgot to use the param in the actual call 😬

eddyb (approved):

PR #958, opened 2022-12-01
Update to nightly-2022-10-29

oisyn:
With this we are on schedule for Rust 1.66 to be released on stable on 15 December.

### DIFF
--- a/crates/rustc_codegen_spirv/src/linker/test.rs
+++ b/crates/rustc_codegen_spirv/src/linker/test.rs
@@ -3,8 +3,7 @@ use crate::codegen_cx::SpirvMetadata;
 use pipe::pipe;
 use rspirv::dr::{Loader, Module};
 use rustc_errors::registry::Registry;
-use rustc_span::edition::Edition;
-use std::{io::Read, thread};
+use std::io::Read;
 
 // https://github.com/colin-kiegel/rust-pretty-assertions/issues/24
 #[derive(PartialEq, Eq)]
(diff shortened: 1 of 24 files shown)


crates/rustc_codegen_spirv/src/linker/test.rs
```diff
@@ -80,67 +118,62 @@ fn assemble_and_link(binaries: &[&[u8]]) -> Result<Module, PrettyString> {
...
-            sopts.unstable_opts.threads,
-            || {
-                let mut sess = rustc_session::build_session(
-                    sopts,
-                    None,
-                    None,
-                    Registry::new(&[]),
-                    rustc_session::DiagnosticOutput::Default,
-                    Default::default(),
+        run_in_thread_pool_with_globals(sopts.edition, sopts.unstable_opts.threads, || {
```
eddyb, line 121:
```suggestion
        rustc_span::create_session_globals_then(sopts.edition, || {
```
Does this work? If so we wouldn't need `run_in_thread_pool_with_globals` (I would've avoided it if I had noticed it got privated, I should've double-checked...).

eddyb (changes requested):
It would be nice if we didn't need the private function copied over.

crates/rustc_codegen_spirv/src/linker/test.rs
oisyn: Lol yeah much simpler

eddyb (approved):

PR #959, opened 2022-12-02
Migrate `rustc_codegen_spirv` env vars to codegen args.

eddyb:
This PR adds a pair of centralized env vars to `spirv-builder`:
* `RUSTGPU_RUSTFLAGS`: sets `RUSTFLAGS` *only* for shader crates (and their deps)
  * i.e. all `rustc` invocations using `rustc_codegen_spirv` as the codegen backends
  * we've long-needed something like this - currently functionality like `-Z time-passes` or `-Z self-profile` requires *editing `spirv-builder`'s source by hand*
* `RUSTGPU_CODEGEN_ARGS`: shorthand for `-Cllvm-args=...` in the above variable
  * (note that `-C llvm-args` is a bad name that does not fit with non-LLVM codegen backends, even if they all have similar concepts of backend-specific flags, and we want to avoid showing it to users, if possible)
  * this will allow easily using various `rustc_codegen_spirv`-specific flags (most of them for debugging) that used to be individual *untracked*¹ env vars

¹ Both of these are properly registered with Cargo in `spirv-builder` (to trigger rebuilds when they change), which is an improvement over the old env vars, which neither rustc nor Cargo knew anything about.

---

The docs have more information on the new "codegen args", but here's a quick preview:
```console
$ RUSTGPU_CODEGEN_ARGS="--help" cargo run -p example-runner-wgpu
...
Usage: RUSTGPU_CODEGEN_ARGS="..." with `...` from:

Options:
    -h, --help          Display this message
        --module-output [single|multiple]
                        single output or multiple output
        --disassemble   print module to stderr
        --disassemble-fn NAME
                        print function to stderr
...
        --no-dce        disables running dead code elimination
        --no-compact-ids 
                        disables compaction of SPIR-V IDs at the end of
                        linking
        --no-structurize 
                        disables CFG structurization
        --dump-post-merge FILE
                        dump the merged module immediately after merging, to
                        FILE
        --dump-post-split DIR
                        dump modules immediately after multimodule splitting,
                        to files in DIR
...
```
(most of the output omitted, but hopefully what was kept is representative)

---

I couldn't find a way to ban `env` in `rustc_codegen_spirv` via e.g. clippy, so instead I did it in a CI script.


oisyn (approved):
Kudos for the docs 👍. That reminds me that we probably need to do a pass over them for `0.4.0`

PR #974, opened 2022-12-21
Trying to fix doc build

oisyn:


repi (approved):

eddyb:
Does this mean we should run "Deploy docs" on PRs too, but only to check it? Maybe add it to "Lint" (which is fast enough - wait, why is "cargo-deny" separate from "Lint"?).

PR #978, opened 2022-12-28
Update docs for manual shader crate building

Bobo1239:
Add some additional flags based on [crates/spirv-builder/src/lib.rs](https://github.com/EmbarkStudios/rust-gpu/blob/8fcb61e82a3bd74df67204c9b4fb1f6f71d20c73/crates/spirv-builder/src/lib.rs#L429-L438) which weren't updated here yet. (causing compilation errors) Also update the wording so it doesn't seem like the user is supposed to call `cargo -Zbuild-std`.


oisyn:
Hey thanks, looks good to me!

PR #982, opened 2023-01-04
Update to nightly-2022-12-18

oisyn:
This brings us on par with 1.67 due for release on January 26.

### DIFF
--- a/crates/spirv-builder/src/lib.rs
+++ b/crates/spirv-builder/src/lib.rs
@@ -312,7 +312,7 @@ impl SpirvBuilder {
         match self.print_metadata {
             MetadataPrintout::Full | MetadataPrintout::DependencyOnly => {
                 leaf_deps(&metadata_file, |artifact| {
-                    println!("cargo:rerun-if-changed={}", artifact);
+                    println!("cargo:rerun-if-changed={artifact}");
                 })
                 // Close enough
                 .map_err(SpirvBuilderError::MetadataFileMissing)?;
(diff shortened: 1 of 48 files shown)


eddyb (approved):

eddyb:
The mergify stuff we forgot to fully remove is the only thing passing, actual GHA looks busted.
Will try to trigger GHA by closing and reopening the PR, can't really find any other controls :/.

crates/spirv-builder/src/lib.rs
```diff
@@ -406,9 +406,7 @@ fn join_checking_for_separators(strings: Vec<impl Borrow<str>>, sep: &str) -> St
         let s = s.borrow();
         assert!(
             !s.contains(sep),
-            "{:?} may not contain separator {:?}",
-            s,
-            sep
+            "{}", "{s:?} may not contain separator {sep:?}"
```
eddyb: This seems wrong, it can only print the literal string you wrote, without formatting. I believe the issue is that we're still on Rust2018 (instead of Rust2021) and the lint might be misfiring?

eddyb (changes requested):
I found several instances of `"{}", "` - that indicates likely a bad replacement, where instead of formatting, you're printing the format string verbatim.

This is likely due to the editions, maybe we need to revive a partial version of [my PR to switch to Rust 2021](https://github.com/EmbarkStudios/rust-gpu/pull/953).

eddyb:
Looks like clippy got fixed in:
* https://github.com/rust-lang/rust-clippy/pull/10055

which got sync'd into `rust-lang/rust` in:
* https://github.com/rust-lang/rust/pull/105827

So theoretically updating to `nightly-2022-12-18` should get us fixed clippy (that will not suggest incorrect changes on pre-Rust2021 code).

We could allow `#![allow` the lint for now, or try to land some version of #953 - I don't have a strong preference.

oisyn:
I'll try updating to `nightly-2022-12-18`. If that gives us headaches we can always just allow.

eddyb (approved):

PR #985, opened 2023-01-10
allow libm_intrinsics fallback

djdisodo:
allow using libm software implementation as fallback


crates/rustc_codegen_spirv/src/builder/libm_intrinsics.rs
```diff
@@ -348,6 +349,8 @@ impl Builder<'_, '_> {
                 self.zombie(undef.def(self), "Scalbn not supported yet");
                 undef
             }
-        }
+             */
+            _ => return None
+        })
```
eddyb, line 354: Commenting this stuff out doesn't seem the best. I think I understand what's going on, but maybe each of these should be evaluated to see if there is a GLSL fallback and removed from the `LimCustomIntrinsic` type entirely, if not.

It would be nice if the only software operations kept would be the ones that are relatively cheap (and mostly just operate on the binary representation).

eddyb (changes requested):
I feel like it would be less effort to not have entries in the `libm_intrinsics` map for any function we don't want to hook at all.

Looking a bit closer at it, several `LibmCustomIntrinsic`s are implemented by expanding them to GLSL ops, so I think the reason the remaining ones panic is that they were meant to be implemented the same way.

Do you have a specific subset in mind that you need?

djdisodo:
> I feel like it would be less effort to not have entries in the `libm_intrinsics` map for any function we don't want to hook at all.
> 
> Looking a bit closer at it, several `LibmCustomIntrinsic`s are implemented by expanding them to GLSL ops, so I think the reason the remaining ones panic is that they were meant to be implemented the same way.
> 
> Do you have a specific subset in mind that you need?

(i need erf and erff)
as `libm_intrinsics` is not listed somewhere but generated,
it needs some work to add exceptions, and i'm not sure what's the best way to do that

djdisodo:
i need erf* functions and it doesn't seem to be on GLOps(i'm not sure)
since currently match has branch for all variant
should i remove erf* branches and make default branch or leave erf branches there

djdisodo:
@eddyb

PR #987, opened 2023-01-11
Removed return type inference from `Image` API

oisyn:
By referring to concrete 'sibling' vector types from same vector library based on input.

@eddyb this currently causes an ICE in `tests/ui/image/gather_err.rs`

### DIFF
--- a/tests/ui/image/gather_err.stderr
+++ b/tests/ui/image/gather_err.stderr
@@ -9,26 +9,90 @@ error[E0277]: the trait bound `Image<f32, 0, 2, 0, 0, 1, 0>: HasGather` is not s
               Image<SampledType, 3, DEPTH, ARRAYED, 0, SAMPLED, FORMAT>
               Image<SampledType, 4, DEPTH, ARRAYED, 0, SAMPLED, FORMAT>
 note: required by a bound in `Image::<SampledType, DIM, DEPTH, ARRAYED, _, SAMPLED, FORMAT>::gather`
-   --> $SPIRV_STD_SRC/image.rs:164:15
+   --> $SPIRV_STD_SRC/image.rs:173:15
     |
-164 |         Self: HasGather,
+173 |         Self: HasGather,
     |               ^^^^^^^^^ required by this bound in `Image::<SampledType, DIM, DEPTH, ARRAYED, _, SAMPLED, FORMAT>::gather`
 
-error[E0277]: the trait bound `Image<f32, 2, 2, 0, 0, 1, 0>: HasGather` is not satisfied
-   --> $DIR/gather_err.rs:16:34
-    |
-16  |     let r2: glam::Vec4 = image3d.gather(*sampler, v3, 0);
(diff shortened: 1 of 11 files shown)


tests/ui/image/gather_err.stderr
```diff
@@ -9,26 +9,90 @@ error[E0277]: the trait bound `Image<f32, 0, 2, 0, 0, 1, 0>: HasGather` is not s
...
-    = help: the following other types implement trait `HasGather`:
-              Image<SampledType, 1, DEPTH, ARRAYED, 0, SAMPLED, FORMAT>
-              Image<SampledType, 3, DEPTH, ARRAYED, 0, SAMPLED, FORMAT>
-              Image<SampledType, 4, DEPTH, ARRAYED, 0, SAMPLED, FORMAT>
-note: required by a bound in `Image::<SampledType, DIM, DEPTH, ARRAYED, _, SAMPLED, FORMAT>::gather`
-   --> $SPIRV_STD_SRC/image.rs:164:15
-    |
-164 |         Self: HasGather,
-    |               ^^^^^^^^^ required by this bound in `Image::<SampledType, DIM, DEPTH, ARRAYED, _, SAMPLED, FORMAT>::gather`
+error: internal compiler error: /rustc/0468a00ae3fd6ef1a6a0f9eaf637d7aa9e604acc/compiler/rustc_middle/src/ty/relate.rs:638:13: var types encountered in super_relate_consts: Const { ty: usize, kind: Param(N/#8) } Const { ty: usize, kind: Infer(Var(_#5c)) }
```
eddyb, line 17: Hmmm, so this seems relevant:
* https://github.com/rust-lang/rust/issues/106240

~~Except, if you look at https://github.com/rust-lang/rust/issues/106240#issuecomment-1367108233 - we're not on a nightly that far ahead, so why would this happen sooner?~~

Talking to the relevant people, those linked PRs are actually not the original cause. Will work on reducing a testcase but either way the fix is the same:
* https://github.com/rust-lang/rust/pull/106322

Which hasn't been merged yet... also it sounds like the bug is in a trick I suggested a few weeks back for taking advantage of inference logic for diagnostic suggestion code.

---

However, if the problem is needing `const N: usize` params, we can probably do better by making the `ImageCoordinate` helper trait mirror the `VectorTypeLib` associated type from the `Vector` trait, reducing the need for the extra `Vector` bounds.

But also, I have even more bad news, that I only noticed while looking for the `ImageCoordinate` impls: https://github.com/EmbarkStudios/rust-gpu/blob/a5e05e4410cd83738db37d3060d5e729b3b87af9/crates/spirv-std/src/image/params.rs#L63-L70

`C` params (before this PR) could be individual scalars, for 1D images. I don't know at this moment how to make both that and the return type stuff, work.

It may require adding a generic parameter for the "vector type lib" itself and setting things up that for vector `C`s, `ImageCoordinate` constraints it (forcing the image op return type to be inferred) while for scalar `C`s, it's unconstrained _and can somehow be inferred from the image op return type_.

eddyb (changes requested):

oisyn:
I'm closing this as we're taking a completely different approach to this.

PR #990, opened 2023-02-10
Removed return type inference from `Image` API

oisyn:
Return types are now always `glam` vectors. This also means that `glam` is required; the "glam" feature toggle is made mandatory rather thn removed as we may want to support other specific vector libraries in the future, and this way people get a clear error when they use something else.

I considered removing the `Vector` trait, but that ripples down to some of the API functions like `any()` so I left it in place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -->
 
+## [Unreleased]
+
+### Changed 🛠
+- [PR#990](https://github.com/EmbarkStudios/rust-gpu/pull/990) Removed return type inference from `Image` API and made `glam` usage mandatory.
+
 ## [0.5.0]
 
 ### Added ⭐
--- a/crates/spirv-std/src/image/params.rs
+++ b/crates/spirv-std/src/image/params.rs
@@ -2,60 +2,85 @@ use super::{Arrayed, Dimensionality, ImageFormat};
 use crate::{integer::Integer, scalar::Scalar, vector::Vector};
 
 /// Marker trait for arguments that accept single scalar values or vectors
-/// of scalars.
-pub trait SampleType<const FORMAT: u32>: Scalar {}
+/// of scalars. Defines 2-, 3- and 4-component vector types based on the sample type.
+pub trait SampleType<const FORMAT: u32>: Scalar {
+    /// A 2-component vector of this sample type
+    type Vec2: Default;
 
-impl SampleType<{ ImageFormat::Unknown as u32 }> for i8 {}
-impl SampleType<{ ImageFormat::Unknown as u32 }> for i16 {}
-impl SampleType<{ ImageFormat::Unknown as u32 }> for i32 {}
-impl SampleType<{ ImageFormat::Unknown as u32 }> for i64 {}
-impl SampleType<{ ImageFormat::Unknown as u32 }> for u8 {}
-impl SampleType<{ ImageFormat::Unknown as u32 }> for u16 {}
-impl SampleType<{ ImageFormat::Unknown as u32 }> for u32 {}
-impl SampleType<{ ImageFormat::Unknown as u32 }> for u64 {}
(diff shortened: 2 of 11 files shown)


crates/spirv-std/src/lib.rs
```diff
@@ -116,6 +116,13 @@ pub use runtime_array::*;
 #[cfg(feature = "glam")]
 pub use glam;
 
+// HACK(shesp) As we removed support for generic user-configurable vector types in the Image API,
+// glam is now required. In the future we might want to add other popular vector math libraries,
+// so we keep the "glam" feature toggle intact for now. As the code won't currently compile
+// without it, present the user with a somewhat friendly error message.
+#[cfg(not(feature = "glam"))]
+compile_error!("`spriv-std` now requires the use of `glam`. Make sure the "glam" feature is specified for `spirv-std` in `Cargo.toml`");
```
eddyb, line 124: Neat! I suspect you haven't tested this because of the syntax error. If you don't want to escape the double quotes here you can use a raw string, e.g.:
```suggestion
compile_error!(
    r#"`spriv-std` now requires the use of `glam`. Make sure the "glam" feature is specified for `spirv-std` in `Cargo.toml`"#
);
```

crates/spirv-std/src/image/params.rs
```diff
@@ -2,60 +2,85 @@ use super::{Arrayed, Dimensionality, ImageFormat};
...
-impl SampleType<{ ImageFormat::R64i as u32 }> for i64 {}
+    /// A 3-component vector of this sample type
+    type Vec3: Default;
+
+    /// A 4-component vector of this sample type
+    type Vec4: Default;
+}
+
+/// Helper macro to implement `SampleType` of various formats for various scalar types.
+macro_rules! sampletype_impls {
```
eddyb, line 18: Nit: `FooBar` typically turns into `foo_bar` when switching to snake case (i.e. `sample_type_impls`).

CHANGELOG.md
```diff
@@ -27,6 +27,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -->
 
+## [Unreleased]
+
+### Changed 🛠
+- [PR#990](https://github.com/EmbarkStudios/rust-gpu/pull/990) Removed return type inference from `Image` API and made `glam` usage mandatory.
```
eddyb, line 33: Should we mark this explicitly as backwards-incompatible?
Also, should we remove the `optional = true` from here?
https://github.com/EmbarkStudios/rust-gpu/blob/fd4da169e41fc42ed3a9ede926a3116dddb7019d/crates/spirv-std/Cargo.toml#L15

eddyb (changes requested):
The bulk of the changes LGTM, only having some questions. Also, are we sure we're doing this? It makes a lot of sense, from a technical standpoint, but I'm not sure what the decision process should be like.

CHANGELOG.md
oisyn: > Should we mark this explicitly as backwards-incompatible?
Not sure what you mean?

> Also, should we remove the optional = true from here?
Yeah I thought about that. The reason I kept it as non-default is that I don't want to silently force glam onto people that are currently NOT using glam. Especially if we want to add support for some other vector library, you'd need to consciously make the choice which one you want. Do you agree with this reasoning?

crates/spirv-std/src/lib.rs
oisyn: Typical test - make last minute change to error message - commit 🙄

oisyn:
> but I'm not sure what the decision process should be like.

In what way, exactly? I've openly discussed making glam mandatory in #rust-gpu, everybody seemed to agree. I don't think the technical change ended up being that big after all. I did play around with more drastic proof of concepts but kept scratching those as I wasn't yet familiar with all the intrinsic details (if I did wanted to go through with one of those I would definitely ran it by you and the others first before committing to a full implementation). It's fully backwards compatible (well, only if you're using `glam` obviously 😉).

PR #991, opened 2023-02-15
Update to nightly-2023-01-21

oisyn:
This nightly contains all changes that have been branched off into beta and will become stable as Rust 1.68 on March 9

### DIFF
--- a/crates/rustc_codegen_spirv/src/lib.rs
+++ b/crates/rustc_codegen_spirv/src/lib.rs
@@ -142,7 +142,8 @@ fn is_blocklisted_fn<'tcx>(
                 Some(assoc) if assoc.ident(tcx).name == sym::fmt => match assoc.container {
                     ty::ImplContainer => {
                         let impl_def_id = assoc.container_id(tcx);
-                        tcx.impl_trait_ref(impl_def_id).map(|tr| tr.def_id)
+                        tcx.impl_trait_ref(impl_def_id)
+                            .map(|tr| tr.subst_identity().def_id)
                             == Some(debug_trait_def_id)
                     }
                     ty::TraitContainer => false,
--- a/crates/rustc_codegen_spirv/src/lib.rs
+++ b/crates/rustc_codegen_spirv/src/lib.rs
@@ -143,7 +143,7 @@ fn is_blocklisted_fn<'tcx>(
                     ty::ImplContainer => {
                         let impl_def_id = assoc.container_id(tcx);
                         tcx.impl_trait_ref(impl_def_id)
-                            .map(|tr| tr.subst_identity().def_id)
+                            .map(|tr| tr.skip_binder().def_id)
                             == Some(debug_trait_def_id)
                     }
                     ty::TraitContainer => false,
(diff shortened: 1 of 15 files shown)


crates/rustc_codegen_spirv/src/lib.rs
```diff
@@ -142,7 +142,8 @@ fn is_blocklisted_fn<'tcx>(
                 Some(assoc) if assoc.ident(tcx).name == sym::fmt => match assoc.container {
                     ty::ImplContainer => {
                         let impl_def_id = assoc.container_id(tcx);
-                        tcx.impl_trait_ref(impl_def_id).map(|tr| tr.def_id)
+                        tcx.impl_trait_ref(impl_def_id)
+                            .map(|tr| tr.subst_identity().def_id)
```
eddyb, line 146: `subst_identity` is not necessarily cheap, since it replaces one encoding of generics with another (the name seems misleading, since "identity substitutions" should, well, produce the same result, while this doesn't).

A `.def_id()` method should exist on this `EarlyBinder<TraitRef>` type just like one exists on `Binder<TraitRef>`, but sadly it seems to be lacking. Instead, you can do this for now:

```suggestion
                        tcx.impl_trait_ref(impl_def_id)
                            .map(|tr| tr.skip_binder().def_id)
```
(`skip_binder` is "bad" because it can lead to mistakes by ignoring important contextual information, but `.def_id` is just an ID, not something in the typesystem that cares about that context)

rust-toolchain
```diff
@@ -1,14 +1,13 @@
...
-channel = "nightly-2022-12-18"
+channel = "nightly-2023-01-21"
 components = ["rust-src", "rustc-dev", "llvm-tools-preview"]
-# commit_hash = 0468a00ae3fd6ef1a6a0f9eaf637d7aa9e604acc
+# commit_hash = 5ce39f42bd2c8bca9c570f0560ebe1fce4eddb14
 
 # Whenever changing the nightly channel, update the commit hash above, and make
-# sure to change REQUIRED_TOOLCHAIN in crates/rustc_codegen_spirv/src/build.rs also.
-
+# sure to change REQUIRED_TOOLCHAIN in crates/rustc_codegen_spirv/src/build.rs also
```
eddyb, line 13: I'm confused, are the punctuation changes here intentional? The removal of a `,`, not just `.` makes it feel accidental.

eddyb (changes requested):

rust-toolchain
oisyn: Ugh. This is VSCode keeping thinking that that file is markdown instead of toml. I thought I disabled format-on-write though, but I definitely not changed those things by hand :)

crates/rustc_codegen_spirv/src/lib.rs
oisyn: I mainly got the pattern from the respective PR where they made similar changes. I'm still not entirely sure what EarlyBinder is supposed to do exactly.

eddyb (approved):

PR #1001, opened 2023-03-15
Made `glam` a default feature in `spirv-std`

oisyn:
We need this to be able to publish `spirv-std`, otherwise it will compile without glam and it will fail

The rationale behind doing it this way was because glam was optional in the past while it is mandatory currently. In the future, we might want to support other vector libraries as well, in which case we need to put glam behind a feature toggle again.

Also, people currently would expect glam not to be included when not specifying it or depending on it with `no-default-features`, so I wanted it to be an in-your-face error when you did that. Unfortunately, this also means that `cargo publish` fails as it compiles without features, so as a quick fix I made it default for the 0.6 release for now. We might want to readdress this.

### DIFF
--- a/crates/spirv-std/Cargo.toml
+++ b/crates/spirv-std/Cargo.toml
@@ -15,4 +15,4 @@ num-traits = { version = "0.2.14", default-features = false, features = ["libm"]
 glam = { version = "0.22", default-features = false, features = ["libm"], optional = true }
 
 [features]
-default = []
+default = ["glam"]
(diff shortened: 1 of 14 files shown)


repi (commented):
think would be cleaner to just remove `optional = true` on the glam dependency and remove the `glam` cargo feature?

oisyn:
I did a quick test with that but that rippled through to other crates that added "glam" as a feature within rust-gpu. But I guess I should just remove those as well.

repi:
yup one would have to remove those crates enabling the `glam` feature, but at least it is fully explicit then that it is required now.

or one can leave the feature but just not enable anything with it, but aren't that many users of this crate so breaking change should be fine

repi (approved):

fu5ha (approved):

PR #1016, opened 2023-03-22
`run-wasm` helper (`cargo-run-wasm`) for testing the wgpu runner on the web.

eddyb:
~~*Opening as draft because this is blocked on `wgpu::backend::web` being completely broken:*~~
*  ~~https://github.com/gfx-rs/wgpu/issues/3430~~
    * ~~<sub>the short version is that they made a `Copy` type convertible to a type that implement `Drop`, leading to a kind of UAF ("use after free") but specifically in the context of `wasm-bindgen`'s Rust->JS value ownership</sub>~~

**TODO**(@eddyb): consider hot reloading.

**EDIT**: blocked on:
* #1036

---

**EDIT**: this is now possible thanks to:
* https://github.com/gfx-rs/wgpu/pull/3657

Had to work around some sRGB complications (and push constants) but it finally works!

* `cargo run-wasm --package example-runner-wgpu --bin wgpu_runner`
  * (defaults to serving the result of the build on `http://localhost:8000`)
* `google-chrome-unstable --enable-features=Vulkan --enable-unsafe-webgpu http://localhost:8000`

image

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,5 +1,9 @@
 [alias]
 compiletest = "run --release -p compiletests --"
+run-wasm = ["run", "--release", "-p", "run-wasm", "--"]
+
+[target.'cfg(target_arch = "wasm32")']
+rustflags = ["--cfg=web_sys_unstable_apis"]
 
 [target.'cfg(all())']
 rustflags = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -203,6 +203,16 @@ dependencies = [
 
 [[package]]
 name = "base64"
+version = "0.9.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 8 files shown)


eddyb:
So I found a hackaround for the `wgpu` misuse of `wasm-bindgen` thing, see https://github.com/gfx-rs/wgpu/issues/3430#issuecomment-1480625569.
(Yupp, it's just leaking - can't UAF if there's no `free` - and it can be done quite fine-grained!)

Not sure how one would automate injecting it, but it's easy enough to edit the `.js` file by hand for testing.

---

On another note, we sadly can't use push constants, but that may change soon (Chrome seems to have them anyway?).

oisyn (approved):

eddyb:
I added `PushConstant`-as-read-only-`StorageBuffer` emulation, and using that and #1018 I get:

image

PR #1023, opened 2023-04-03
Always inline calls with "not obviously legal" pointer args (instead of the opposite).

### LINKED ISSUE #1021: Inliner heuristic for "illegal pointer argument" is incomplete.
This is technically not a new issue, and most of it was fixed in:
* https://github.com/EmbarkStudios/rust-gpu/issues/554
* https://github.com/EmbarkStudios/rust-gpu/pull/777

But if we take the test for that (`tests/ui/lang/core/ref/member_ref_arg.rs`) and add another function `g`:
```rust
struct S {
    x: u32,
    y: u32,
}

fn f(x: &u32) {}

fn g(xy: (&u32, &u32)) {}

#[spirv(fragment)]
pub fn main() {
    let s = S { x: 2, y: 2 };
    f(&s.x);
    g((&s.x, &s.y));
}
```
Then only the (existing) `f` call is inlined, but the `g` call still produces:
```
error: error:0:0 - Pointer operand '22[%22]' must be a memory object declaration
```

This is because #777 banned specifically `OpAccessChain` function arguments (i.e. `&s.x`), but in the case of `g`, the pair is created and then (because it has `ScalarPair` ABI) the two pointer fields are extracted to call `g`.

Based on my reading of the SPIR-V standard, the only way to be fully correct here is to *only* allow `OpVariable`s and `OpFunctionParameter`s (instead of banning one thing or another).

cc @Shfty (who ran into this while doing some type/trait metaprogramming which I guess involved ref-pairs)

eddyb:
* fixes #1021

Before this PR, the behavior of the inliner wrt pointer types in `fn` signatures (and pointer args) was to special-case a illegal cases and force inlining. This PR flips some of that on its head, treating everything it's not sure about as illegal, forcing inlining in all the cases not explicitly deemed legal.

Specifically for fixing #1021, the necessary change is obeying the SPIR-V spec about pointer call arguments having to be "memory object declarations" (i.e. an `OpVariable`, or one of the caller's own `OpFunctionParameter`s, which transitively must be a whole `OpVariable` etc.).

With this PR, any pointer argument other than an `OpVariable` (either global or in the caller) or an `OpFunctionParameter`, will force inlining (even if some of those edge cases may actually be legal, we'll still be correctly conservative).

(The several commits should be reviewed separately, as the first 3 only refactor `linker::inline`'s existing logic)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed 🩹
+- [PR#1023](https://github.com/EmbarkStudios/rust-gpu/pull/1023) fixed [#1021](https://github.com/EmbarkStudios/rust-gpu/issues/1021) by always inlining calls with "not obviously legal" pointer args (instead of only inlining calls with "obviously illegal" pointer args)
 - [PR#1009](https://github.com/EmbarkStudios/rust-gpu/pull/1009) fixed [#1008](https://github.com/EmbarkStudios/rust-gpu/issues/1008) by reinstating mutability checks for entry-point parameters pointing into read-only storage classes (e.g. `#[spirv(uniform)] x: &mut u32` is now again an error)
 - [PR#995](https://github.com/EmbarkStudios/rust-gpu/pull/995) fixed [#994](https://github.com/EmbarkStudios/rust-gpu/issues/994) by using `OpAtomicFAddEXT` instead of `OpAtomicFMaxEXT` in `atomic_f_add`.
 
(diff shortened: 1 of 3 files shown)


Shfty:
This has been tested with the [MRP](https://gist.github.com/Shfty/4a72122d97995d748dccf17d4baa12ae) + [(SPIR-T HTML preview)](https://htmlpreview.github.io/?https://gist.githubusercontent.com/Shfty/4a72122d97995d748dccf17d4baa12ae/raw/b56d3045fb71c0ef395899ab1a395ff6a49bb661/shader.spirt.html&dark) that instigated #1021, and compiles it correctly.

Anecdotally, it's also been tested with the [practical case](https://github.com/Bevy-Rust-GPU/rust-gpu-sdf) that consumes the [cons borrowing trait](https://github.com/Bevy-Rust-GPU/type-fields/blob/master/src/cons.rs#L57) which highlighted the bug, is functioning as expected there as well :+1:

oisyn (approved):

PR #1029, opened 2023-04-06
fixed SampledImage::sample() fns being unnecessarily marked as unsafe

Firestar99:
Seems to be a forgotten leftover from a refactor in https://github.com/EmbarkStudios/rust-gpu/pull/359

---

**EDIT**(@eddyb): removed quoted Discord messages (referring to https://github.com/EmbarkStudios/rust-gpu/commit/bc2a121cca0da4a4020e6630c2e429d85cb4fdf9) in favor of PR comments (see below)


eddyb (approved):
Context for what went wrong here: in https://github.com/EmbarkStudios/rust-gpu/commit/bc2a121cca0da4a4020e6630c2e429d85cb4fdf9
(intermediary state during the original `Image` PR), the `S` method-level parameter was replaced with the `Image` "sampled type" (which, crucially, didn't exist initially in the `Image` PR).

But the signature change wasn't accompanied by removing the then-inaccurate safety aspects (the UB only ever stemmed from the `OpTypeImage` "Sampled Type" and that `S` parameter being able to differ).

PR #1030, opened 2023-04-07
Add vector-or-scalar traits

BeastLe9enD:
This PR adds the traits `ScalarVector`, `NumberVector`, `IntegerVector`, `UnsignedIntegerVector`, `SignedIntegerVector`, `FloatVector`.
This traits can be either a scalar or a vector type. (Not sure how to name them, I'm open for changes there).

Motivation:
I am currently adding support for subgroup operations, and for example `OpGroupNonUniformIAdd`s result type must be scalar or vector of integer type, so it would be useful to have traits that represent types like this.

### DIFF
--- a/crates/spirv-std/src/float.rs
+++ b/crates/spirv-std/src/float.rs
@@ -9,17 +9,50 @@ use core::arch::asm;
 /// # Safety
 /// Implementing this trait on non-primitive-float types breaks assumptions of other unsafe code,
 /// and should not be done.
-pub unsafe trait Float: num_traits::Float + crate::scalar::Scalar + Default {
+pub unsafe trait Float: num_traits::Float + crate::scalar::Scalar {
     /// Width of the float, in bits.
     const WIDTH: usize;
 }
 
-unsafe impl Float for f32 {
-    const WIDTH: usize = 32;
+/// Abstract trait representing a SPIR-V floating point type. It may also be a vector type.
+///
+/// # Safety
+/// Implementing this trait on non-primitive-float types breaks assumptions of other unsafe code,
+/// and should not be done.
+pub unsafe trait FloatVector: crate::scalar::ScalarVector {
+    /// Width of the float, in bits.
+    const WIDTH: usize;
 }
 
-unsafe impl Float for f64 {
-    const WIDTH: usize = 64;
+macro_rules! impl_floats {
+    (unsafe impl Float for $($typ:ty)*;) => {
+        $(
+            unsafe impl Float for $typ {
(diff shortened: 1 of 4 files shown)


crates/spirv-std/src/vector.rs
```diff
@@ -23,3 +25,54 @@ unsafe impl Vector<u32, 4> for glam::UVec4 {}
...
+///
+/// # Safety
+/// Implementing this trait on non-simd-vector/scalar types breaks assumptions of other unsafe code, and
+/// should not be done.
+pub unsafe trait ScalarOrVector {
+    /// The scalar type
+    type Scalar;
+
+    /// The dimension of the vector, or 1 if it is a scalar
+    const DIM: usize;
```
LegNeato, line 39: Can you use [std](https://doc.rust-lang.org/stable/std/index.html)::[num](https://doc.rust-lang.org/stable/std/num/index.html)::[NonZeroUsize](https://doc.rust-lang.org/stable/std/num/struct.NonZeroUsize.html#) here?

BeastLe9enD:
@oisyn Yea the thing is, my plan is to add support for subgroup operations after this PR has been merged, and there are some ops like `OpGroupNonUniformIAdds` for which this trait can be really useful.

crates/spirv-std/src/vector.rs
BeastLe9enD: Yes, theoretically you could do it, but I am not sure if it is currently supported in Rust-Gpu.

oisyn:
Right! I would suggest adding this trait as part of that work.

crates/spirv-std/src/vector.rs
```diff
@@ -23,3 +25,54 @@ unsafe impl Vector<u32, 4> for glam::UVec4 {}
...
+    for f64, 4: unsafe impl ScalarOrVector for glam::DVec4;
+
+    for u32, 2: unsafe impl ScalarOrVector for glam::UVec2;
+    for u32, 3: unsafe impl ScalarOrVector for glam::UVec3;
+    for u32, 4: unsafe impl ScalarOrVector for glam::UVec4;
+
+    for i32, 2: unsafe impl ScalarOrVector for glam::IVec2;
+    for i32, 3: unsafe impl ScalarOrVector for glam::IVec3;
+    for i32, 4: unsafe impl ScalarOrVector for glam::IVec4;
+}
```
eddyb, line 78: It would be nicer to make this one per line, like `i32: IVec2 => 2, IVec3 => 3, IVec4 => 4;`
Also, you can make the macro implement both `Vector` and `ScalarOrVector` at the same time.

PR #1046, opened 2023-04-22
Update platform-support.md

TrueDoctor:
Remove internal and unsupported compilation targets from public docs


eddyb (changes requested):
This is not ideal, the reality is much more complicated. Some of these still work, and they mostly come from SPIRV-Tools, so it will take more to rectify it.

Issue #1053, opened 2023-04-25
Unable to compile `spirv-std-macros` with old `Cargo.lock`.


dsvensson:
When trying to get [this project](https://github.com/hatoo/ash-raytracing-example) working with more recent versions of dependencies I get the following with spirv-std bumped to 0.7.0

```
error[E0599]: no method named `token` found for struct `LitStr` in the current scope
   --> C:\....\.cargo\registry\src\index.crates.io-6f17d22bba15001f\spirv-std-macros-0.7.0\src\lib.rs:790:65
    |
790 | ...                   new_t.push(TokenTree::Literal(l.token()));
    |                                                       ^^^^^ method not found in `LitStr`
```

I bumped the rust-toolchain.toml to the same as in this repo fwiw.

eddyb:
Does this go away when you run `cargo update -p syn`? If so, the issue is that we only require an old version:
https://github.com/EmbarkStudios/rust-gpu/blob/a64857af0247dabdf08103b9c91ce0355c6530b8/crates/spirv-std/macros/Cargo.toml#L17
But some other dependencies are forcing newer versions:
https://github.com/EmbarkStudios/rust-gpu/blob/a64857af0247dabdf08103b9c91ce0355c6530b8/Cargo.lock#L2386-L2388

Thankfully, this shouldn't affect newer projects, as Cargo will always pick the newest version when creating the `Cargo.lock`, but it is a mistake on our part and we should be careful with dependencies like that.

This is probably not strong enough, but we should at least [run `cargo update -Z minimal-versions`](https://github.com/rust-lang/cargo/issues/4100) in CI, to force dependency versions in `Cargo.toml` files to be large enough to allow building & passing tests (cc @repi).

But I just tried that and it fails to compile `anyhow 1.0.0`, so that probably means the ecosystem isn't necessarily readily to do something that broad. Also, it would perhaps be nicer if we just forced version bumps in `Cargo.lock` to be reflected in `Cargo.toml`, instead, but I can't find a command that would make the two equal.

eddyb:
@dsvensson I've opened a tentative fix:
* #1054

But note that the simplest fix for that project is still likely running `cargo update -p syn`.

dsvensson:
Aha, doh, so deleting Cargo.lock and rebuilding, should have tried. :man_facepalming: Thanks.

dsvensson:
[...]

PR #1054, opened 2023-04-26
ci: add custom lint for `spirv-std` building with `-Z minimal-versions`.

eddyb:
This tries to prevent issues like:
* #1053

Note that the separation into commits is to aid review, and also I'm going to open this PR without the fix for the failure, so that I can demonstrate CI failing with the new check.


eddyb:
[CI fails as expected](https://github.com/EmbarkStudios/rust-gpu/actions/runs/4805654088/jobs/8552282020?pr=1054), will push fix commit now.

crates/spirv-std/macros/Cargo.toml
```diff
@@ -14,4 +14,4 @@ proc-macro = true
 spirv-std-types.workspace = true
 proc-macro2 = "1.0.24"
 quote = "1.0.8"
-syn = { version = "1.0.58", features = ["full", "visit-mut"] }
+syn = { version = "1.0.85", features = ["full", "visit-mut"] }
```
eddyb, line 17: Should we consider switching to [`syn 2.0.0`](https://github.com/dtolnay/syn/releases/tag/2.0.0)?
I guess even if we do, I still want to land the CI check separately.

repi: yeah we can definitely switch

repi (approved):

Issue #1048, opened 2023-04-24
`DebugPrintf` prevents panics' infinite loops from being (unsoundly) optimized away.


charles-r-earp:
#768

See https://github.com/charles-r-earp/rust-gpu/tree/debug_printf_panic.

A minimal example that prints and panics. 
```rust
#![no_std]
#![feature(asm_experimental_arch)]

use spirv_std::macros::spirv;

#[spirv(compute(threads(1)))]
pub fn main() {
    unsafe {
        spirv_std::macros::debug_printfln!();
    }
    // panic becomes infinite loop, normally optimized away
    // but with debug_printf it is not.
    panic!();
}
```

```
cargo run -p debug-printf-panic
// output
; SPIR-V
; Version: 1.5
; Generator: Google rspirv; 0
; Bound: 22
; Schema: 0
               OpCapability Shader
               OpCapability VulkanMemoryModel
               OpExtension "SPV_KHR_non_semantic_info"
          %1 = OpExtInstImport "NonSemantic.DebugPrintf"
               OpMemoryModel Logical Vulkan
               OpEntryPoint GLCompute %2 "main"
               OpExecutionMode %2 LocalSize 1 1 1

               ; Debug Information
          %3 = OpString "
"

               ; Types, variables and constants
       %void = OpTypeVoid
          %7 = OpTypeFunction %void

               ; Function 2
          %2 = OpFunction %void None %7
          %8 = OpLabel
         %17 = OpExtInst %void %1 1 %3
               OpBranch %18
         %18 = OpLabel
               OpLoopMerge %21 %18 None
               OpBranch %18
         %21 = OpLabel
               OpUnreachable
               OpFunctionEnd

#version 450
layout(local_size_x = 1, local_size_y = 1, local_size_z = 1) in;

void main()
{
    // unimplemented ext op 12
    for (;;)
    {
    }
}
```

This causes a crash or hang when running the shader.

eddyb:
> ```rust
>     // panic becomes infinite loop, normally optimized away
>     // but with debug_printf it is not.
> ```

> This causes a crash or hang when running the shader.

Yes, that's the intended behavior from the infinite loop, that a panic hangs.
It's actually a bug that the infinite loop gets optimized away sometimes (maybe it should get a control barrier or something, to force it to stick around).

[...]

PR #1055, opened 2023-04-26
Added barrier to panic loop.

charles-r-earp:
Related to #1048.

On `panic!`, rustc_codegen_spirv emits a `loop {}`, but this can be optimized away with spirv-opt due to the AgressiveDCE pass. This PR inserts a memory_barrier inside the loop which prevents it from being removed. 

This ensures that shaders that panic, ie from out of bounds accesses, will not silently continue execution but hang the app. 

Unfortunately the wgpu compute example errors with `UnsupportedInstruction(Function, MemoryBarrier)`, likely naga does not support barriers. 

It does work using `create_shader_module_spirv` and enabling the features `Features::SPIRV_SHADER_PASSTHROUGH | Features::TIMESTAMP_QUERY_INSIDE_PASSES`. 

Edit:
Replaced with https://github.com/charles-r-earp/rust-gpu/tree/panic-abort.

The idea is for panic! to emit OpKill / OpTerminateInvocation and replace that with a return from the entry point after inlining. Additionally it will use debug_printfln! if available so that it's easier to detect panics while developing.

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/intrinsics.rs
+++ b/crates/rustc_codegen_spirv/src/builder/intrinsics.rs
@@ -363,7 +363,7 @@ impl<'a, 'tcx> IntrinsicCallMethods<'tcx> for Builder<'a, 'tcx> {
             let uint = builder.type_int(32, 0);
             let memory = builder.constant_u32(uint, memory);
             let semantics = builder.constant_u32(uint, semantics);
-            builder.memory_barrier(memory, semantics).unwrap()
+            builder.memory_barrier(memory, semantics).unwrap();
         }
         self.br(abort_loop_bb);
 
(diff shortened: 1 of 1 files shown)


eddyb:
Is it safe to say this PR has been obsoleted by the following?
* https://github.com/EmbarkStudios/rust-gpu/pull/1070
* https://github.com/EmbarkStudios/rust-gpu/pull/1080
* https://github.com/EmbarkStudios/rust-gpu/pull/1082

charles-r-earp:
> Is it safe to say this PR has been obsoleted by the following?
> 
>     * [Lower aborts (incl. panics) to "return from entry-point", instead of infinite loops. #1070](https://github.com/EmbarkStudios/rust-gpu/pull/1070)
> 
>     * [Add `debugPrintf`-based panic reporting, controlled via `spirv_builder::ShaderPanicStrategy`. #1080](https://github.com/EmbarkStudios/rust-gpu/pull/1080)
> 
>     * [Show `panic!` messages via `debugPrintf`, even including some runtime arguments (`{u,i,f}32` as `{}` or `{:?}`). #1082](https://github.com/EmbarkStudios/rust-gpu/pull/1082)

Yes

eddyb:
Alright then, closing in that case - thanks for prompting all of this, btw, Rust-GPU panics would've probably stayed in a pretty bad shape for a while longer if not for #1048 and the ensuing discussions.

PR #1059, opened 2023-05-05
Inliner: rewrite IDs on decorations fix

SiebenCorgie:
Currently the inliner only rewrites IDs in the 'inlined_callee_blocks'. Therefore 'OpDecorate %id' on any '%id' that is rewritten is lost in this pass.

This fix applies the rewrite rules to all decorations as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     (`0.11.1` for `wgpu 0.15`, `0.12.1` for `wgpu 0.16`, and any later versions)
 
 ### Fixed 🩹
-- [PR#1059](https://github.com/EmbarkStudios/rust-gpu/pull/1059) Fix bug where the `inline` pass would not rewrite `OpDecorate` IDs, which would subsequently loose those decorations in later DCE passes.
+- [PR#1059](https://github.com/EmbarkStudios/rust-gpu/pull/1059) fixed the `inline` pass not copying `OpDecorate`s in the callee (which led to their loss).
 
 ## [0.7.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     you may be able to `cargo update -p naga` to update to a fixed `naga` version  
     (`0.11.1` for `wgpu 0.15`, `0.12.1` for `wgpu 0.16`, and any later versions)
 
+### Fixed 🩹
+- [PR#1059](https://github.com/EmbarkStudios/rust-gpu/pull/1059) Fix bug where the `inline` pass would not rewrite `OpDecorate` IDs, which would subsequently loose those decorations in later DCE passes.
+
 ## [0.7.0]
 
 ### Added ⭐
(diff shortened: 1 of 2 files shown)


eddyb (changes requested):
You need to *duplicate* annotations on inlining - you should try testing with multiple call sites. (also, you should update the CHANGELOG)

crates/rustc_codegen_spirv/src/linker/inline.rs
```diff
@@ -350,6 +352,24 @@ impl Inliner<'_, '_> {
...
 
+    ///Applies all rewrite rules to the decorations in the header.
+    fn apply_rewrite_for_decorations(&mut self, rewrite_rules: &FxHashMap<Word, Word>) {
+        // NOTE(siebencorgie): We don't care *what* decoration we rewrite atm. AFAIK there is no case where rewriting
+        // the decoration on inline wouldn't be valid.
+        for annotation_idx in 0..self.annotations.len() {
+            if self.annotations[annotation_idx].class.opcode == Op::Decorate {
+                if let Some(id) = self.annotations[annotation_idx].operands[0].id_ref_any_mut() {
+                    if let Some(&rewrite) = rewrite_rules.get(id) {
+                        //Copy decoration instruction and push it.
```
eddyb, line 363:
```suggestion
                        // Copy decoration instruction and push it.
```

crates/rustc_codegen_spirv/src/linker/inline.rs
```diff
@@ -350,6 +352,24 @@ impl Inliner<'_, '_> {
         result
     }
 
+    ///Applies all rewrite rules to the decorations in the header.
```
eddyb, line 355:
```suggestion
    /// Applies all rewrite rules to the decorations in the header.
```

CHANGELOG.md
```diff
@@ -38,6 +38,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     you may be able to `cargo update -p naga` to update to a fixed `naga` version  
     (`0.11.1` for `wgpu 0.15`, `0.12.1` for `wgpu 0.16`, and any later versions)
 
+### Fixed 🩹
+- [PR#1059](https://github.com/EmbarkStudios/rust-gpu/pull/1059) Fix bug where the `inline` pass would not rewrite `OpDecorate` IDs, which would subsequently loose those decorations in later DCE passes.
```
eddyb, line 42:
```suggestion
- [PR#1059](https://github.com/EmbarkStudios/rust-gpu/pull/1059) fixed the `inline` pass not copying `OpDecorate`s in the callee (which led to their loss).
```

crates/rustc_codegen_spirv/src/linker/inline.rs
```diff
@@ -350,6 +352,24 @@ impl Inliner<'_, '_> {
...
+    fn apply_rewrite_for_decorations(&mut self, rewrite_rules: &FxHashMap<Word, Word>) {
+        // NOTE(siebencorgie): We don't care *what* decoration we rewrite atm. AFAIK there is no case where rewriting
+        // the decoration on inline wouldn't be valid.
+        for annotation_idx in 0..self.annotations.len() {
+            if self.annotations[annotation_idx].class.opcode == Op::Decorate {
```
eddyb, line 360: Hmm this is enough for your usecase but it should be possible to extend it to all decorate instructions (which I'll need for late zombie reporting).

eddyb (changes requested):

eddyb (approved):

eddyb:
Sorry for the force push to your branch, had to rebase it to land it (I wonder if it's because of the merge commit you had - next time, use `git pull --rebase` instead of `git pull`, to avoid creating merge commits).

PR #1069, opened 2023-06-03
Add `rustfmt.toml` with `version = "Two"` to enable formatting array patterns.

eddyb:
Was checking something about SPIR-T CI and realized I'd forgotten I ended up with a `rustfmt.toml` for SPIR-T that contained `version = "Two"` because of this PR:
* https://github.com/rust-lang/rustfmt/pull/4994

I'm not really sure why a decision was taken to not enable formatting array/slice patterns by default, making them nearly-unusable in non-trivial cases, but this is the result.

You'll notice there's many more changes than just array/slice patterns, looks like `version = "Two"` also enables a few more quality-of-life improvements as well, and I'm not sure if it's the direct cause but I was also able to remove some `#[rustfmt::skip]`s which didn't do anything anymore (see the second commit).


repi (approved):
nice, didn't know about this either. but is another nightly rustfmt feature but sounds fine to use here!

PR #1064, opened 2023-05-26
Add `custom_insts` for our own `OpExtInst`s, and use it for some debuginfo.

eddyb:
This PR introduces a Rust-GPU-specific custom "extended instruction set" (`OpExtInstImport`/`OpExtInst` - note that unlike the requirements imposed by the SPIR-V specification, we're only concerned about `rspirv`/`spirt` compatibility, as long as we always rewrite such instructions before the final SPIR-V output).

Longer-term, we'll likely want such custom instructions for more advanced usecases (e.g. one idea being floated around was adding our own `OpAbort` that gets transformed to standard SPIR-V), and the debuginfo needs will likely get superseded by adopting `NonSemantic.Shader.DebugInfo.100`, but for now this is easier for some debuginfo.

The main feature this PR adds is the ability to have location *ranges*, not just the starting point:
<table>
<tr><th>before this PR (since #1040)</th><th>after this PR</th></tr>
<tr><td><sub><pre>
error: cannot memcpy dynamically sized data
    --> $CORE_SRC/intrinsics.rs:2460:9
     |
2460 |         copy(src, dst, count)
     |         ^
     |
note: used from within `core::intrinsics::copy::<f32>`
    --> $CORE_SRC/intrinsics.rs:2447:21
     |
2447 | pub const unsafe fn copy<T>(src: *const T, dst: *mut T, count: usize) {
     |                     ^
note: called by `ptr_copy::copy_via_raw_ptr`
    --> $DIR/ptr_copy.rs:28:18
     |
28   |         unsafe { core::ptr::copy(src, dst, 1) }
     |                  ^
note: called by `ptr_copy::main`
    --> $DIR/ptr_copy.rs:33:5
     |
33   |     copy_via_raw_ptr(&i, o);
     |     ^
note: called by `main`
    --> $DIR/ptr_copy.rs:31:1
     |
31   | #[spirv(fragment)]
     | ^
</pre></sub></td><td><sub><pre>
error: cannot memcpy dynamically sized data
    --> $CORE_SRC/intrinsics.rs:2460:9
     |
2460 |         copy(src, dst, count)
     |         ^^^^^^^^^^^^^^^^^^^^^
     |
note: used from within `core::intrinsics::copy::<f32>`
    --> $CORE_SRC/intrinsics.rs:2447:21
     |
2447 | pub const unsafe fn copy<T>(src: *const T, dst: *mut T, count: usize) {
     |                     ^^^^
note: called by `ptr_copy::copy_via_raw_ptr`
[...]

PR #1070, opened 2023-06-05
Lower aborts (incl. panics) to "return from entry-point", instead of infinite loops.

eddyb:
We currently map the `abort` intrinsic (used almost exclusively for panic) to infinite loops, and they either:
- are optimized out by `spirv-opt` or drivers (i.e. treated as UB)
  - this obviously changes the semantics, and if a conditional panic protected against some UB, now that UB is unconditional and could affect more code
  - worst case, unwanted side-effects could run (e.g. writes of corrupted values, to buffers)
- are preserved by *both* `spirv-opt` *and* drivers, and cause a timeout when used
  - see these previous discussions for what it take to *intentionally* cause this (cc @charles-r-earp):
    - https://github.com/EmbarkStudios/rust-gpu/issues/1048
    - https://github.com/EmbarkStudios/rust-gpu/pull/1055
  - however, this is only a reliable option cross-hardware only for compute-only Vulkan queues (not just compute *shaders*, but compute shaders on a non-graphical queue), where the timeout can't block other work
  - worst case, one can easily accidentally hang *their entire* desktop if GPU compositing is involved
    - (yes I've done this to myself while trying out the barrier trick, no it wasn't fun)

With infinite loops being so terrible, I propose we move towards a "well-defined invocation exit" approach, where we keep the "abort" as a custom instruction (using our "extended instruction set", added in #1064), and then effectively emulate the semantics of [`OpTerminateInvocation`](https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#OpTerminateInvocation) for it by:
- inlining any function that uses our custom `Abort` (either directly, or transitively through some functions *it* calls) - in the end, we should end up with `Abort`s only used directly from entry-points
- in entry-points, we rewrite `Abort`s to a plain `OpReturn` (from the entry-point, i.e. exiting the invocation)
  - we could potentially have a mode where we generate a `debugPrintf` call at this point, with the same inlining-aware "backtrace" we use elsewhere, so that the user gets some feedback if they have the validation layers enabled (and/or try to extract a panic message when we generate the abort in the first place, too)

This PR implements that proposal (but without any `debugPrintf` conveniences), and so far it seems to work great, but I haven't tested the performance impact (i.e. where before the infinite loops were optimized away, now we're seeing an actual cost to various e.g. bounds checks, that need to *do something* at all).

There are also other ways of implementing this, and we could do the `Abort` -> `OpReturn` rewriting very late (if we think it would be better than letting the SPIR-T structurizer see it), so there's some room to explore mitigations to perf issues, if they arise.

~~(I will leave this PR as draft until we're sure about the perf aspects)~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed 🛠
+- [PR#1070](https://github.com/EmbarkStudios/rust-gpu/pull/1070) made panics (via the `abort` intrinsic)
+  early-exit (i.e. `return` from) the shader entry-point, instead of looping infinitely
 - [PR#1071](https://github.com/EmbarkStudios/rust-gpu/pull/1071) updated toolchain to `nightly-2023-05-27`
 
 ## [0.8.0]
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -2457,8 +2457,7 @@ impl<'a, 'tcx> BuilderMethods<'a, 'tcx> for Builder<'a, 'tcx> {
                         let is_standard_debug = [Op::Line, Op::NoLine].contains(&inst.class.opcode);
                         let is_custom_debug = inst.class.opcode == Op::ExtInst
                             && inst.operands[0].unwrap_id_ref() == custom_ext_inst_set_import
-                            && [CustomOp::SetDebugSrcLoc, CustomOp::ClearDebugSrcLoc]
-                                .contains(&CustomOp::decode_from_ext_inst(inst));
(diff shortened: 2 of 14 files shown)


repi (approved):
approving, but with the reservation that there are no major performance regressions with this on esp. AMD and Nvidia GPUs in `ark`. which I believe @VZout was going to try and verify?

VZout:
No noticable perf difference in ark. near identical timings.

VZout (approved):

eddyb:
Thanks for the confirmation! I will leave this open until I have anything else to land (as I'd rather not stack too many PRs), to give more time to anyone who may have perf-sensitive shaders (e.g. @charles-r-earp @pema99 @Shfty) to test it, but we're likely fine as-is based on what I'm seeing.

eddyb:
Decided to merge this despite wanting to dig further into @pema99's usecase, where this PR does seem to have a perf impact - ideally we will move towards more flexibility, which will make comparisons easier.

PR #1079, opened 2023-07-14
spirv-builder: try to get docs to build on `docs.rs`, and improve them slightly.

eddyb:
See each commit for more details, but the original motivation was the first commit (I was working on another change that requires `spirv-builder` additions and it would be great if I could link to `docs.rs` docs).

Quick diff for the second commit:
|Before|After|
|-|-|
|image|image|

The potentially controversial change is removing mentions of `0.4.0-alpha.*` prereleases - they keep feeling out of place after so many more releases, and IMO if someone is updating from a `0.4.0-alpha.*` prerelease, they should probably update to `0.4` proper first, and only *then* look at what was released since.


repi (approved):
looks good!

PR #1080, opened 2023-07-14
Add `debugPrintf`-based panic reporting, controlled via `spirv_builder::ShaderPanicStrategy`.

eddyb:
This is a followup to my earlier PR about panics (and the only thing I want to really get into `0.9`):
* #1070

There are better docs on `spirv_builder::ShaderPanicStrategy`, but this is the basic usage example:
```rust
    SpirvBuilder::new("examples/shaders/sky-shader", "spirv-unknown-vulkan1.1")
        .shader_panic_strategy(spirv_builder::ShaderPanicStrategy::DebugPrintfThenExit {
            print_inputs: true,
            print_backtrace: true,
        })
```
<sub>(that is, to enable emitting all the information we can right now - `print_inputs` will show values of entry-point inputs, to identify the invocation, while `print_backtrace` will collect the static "inlined call frames" information we track for diagnostics, into the message)</sub>

And this is what it looks like with an artificially induced panic in `sky-shader` (plus some extra functions that just call the next one in the chain, to show off a non-trivial backtrace):

---

If you look at the diff, you may notice most of it is to the inliner: yupp it had more bugs, but not only that, the entire strategy for tracking debuginfo was wrong because it's fundamentally an outside-in inliner, instead of the usual inside-out model (I'm honestly surprised any of the past diagnostics demos worked at all).

The other changes I've made were to the `ash` runner, hopefully they're not too disruptive (cc @DJMcNab).
This is how I was testing (after sprinkling `sky-shader` with extra panics):
```sh
VK_LAYER_ENABLES=VK_VALIDATION_FEATURE_ENABLE_DEBUG_PRINTF_EXT cargo run -p example-runner-ash --release -- -d
```

Haven't been able to break anything with it, so I'm relatively confident the inliner changes aren't semantically meaningful (after all, the whole point is to tweak debuginfo to make it more accurate, not to change anything else).


repi (approved):

eddyb:
(marked as draft to avoid accidentally merging w/o changelog)

fu5ha (approved):

eddyb:
For anyone seeing this PR later, the output, `wgpu` compatibility, and docs, were improved in:
* https://github.com/EmbarkStudios/rust-gpu/pull/1036

PR #1081, opened 2023-07-19
Add specialization constants via `#[spirv(spec_constant(id = 123))] x: u32` entry-point inputs.

eddyb:
Quoting some of the documentation I added to `docs/src/attributes.md`:

> Entry point inputs also allow access to [SPIR-V "specialization constants"](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#SpecializationSection),
> which are each associated with an user-specified numeric "ID" (SPIR-V `SpecId`),
> used to override them later ("specializing" the shader):
> * in Vulkan: [during pipeline creation, via `VkSpecializationInfo`](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap10.html#pipelines-specialization-constants)
> * in WebGPU: [during pipeline creation, via `GPUProgrammableStage`<i>`#constants`</i>](https://www.w3.org/TR/webgpu/#gpuprogrammablestage)
>   * note: WebGPU calls them ["pipeline-overridable constants"](https://gpuweb.github.io/gpuweb/wgsl/#pipeline-overridable)
> * in OpenCL: [via `clSetProgramSpecializationConstant()` calls, before `clBuildProgram()`](https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clSetProgramSpecializationConstant.html)
> 
> If a "specialization constant" is not overriden, it falls back to its *default*
> value, which is either user-specified (via `default = ...`), or `0` otherwise.
> 
> While only "specialization constants" of type `u32` are currently supported, it's
> always possible to *manually* create values of other types, from one or more `u32`s.
> 
> Example:
> 
> ```rust
> #[spirv(vertex)]
> fn main(
>     // Default is implicitly `0`, if not specified.
>     #[spirv(spec_constant(id = 1))] no_default: u32,
> 
>     // IDs don't need to be sequential or obey any order.
>     #[spirv(spec_constant(id = 9000, default = 123))] default_123: u32,
> 
>     // Assembling a larger value out of multiple `u32` is also possible.
>     #[spirv(spec_constant(id = 100))] x_u64_lo: u32,
>     #[spirv(spec_constant(id = 101))] x_u64_hi: u32,
> ) {
>     let x_u64 = ((x_u64_hi as u64) << 32) | (x_u64_lo as u64);
> }
> ```

---

There's not much else to it, quite bare-bones, but I wanted to have something for @VZout before the release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,11 +30,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added ⭐
+- [PR#1081](https://github.com/EmbarkStudios/rust-gpu/pull/1081) added the ability
+  to access SPIR-V specialization constants (`OpSpecConstant`) via entry-point
+  inputs declared as `#[spirv(spec_constant(id = ..., default = ...))] x: u32`  
+  (see also [the `#[spirv(spec_constant)]` attribute documentation](docs/src/attributes.md#specialization-constants))
 - [PR#1036](https://github.com/EmbarkStudios/rust-gpu/pull/1036) added a `--force-spirv-passthru` flag to `example-runner-wgpu`, to bypass Naga (`wgpu`'s shader translator),
   used it to test `debugPrintf` for `wgpu`,  and updated `ShaderPanicStrategy::DebugPrintfThenExit` docs to reflect what "enabling `debugPrintf`" looks like for `wgpu`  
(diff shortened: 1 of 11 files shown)


eddyb:
To test this for myself, I made it so that in `example-runner-ash`, you can now NumPad +/- to control `sky-shader`'s "sun intensity", through a new specialization constant (and yes, it will rebuild the pipeline, and require recompiling the shaders in the driver, etc., every single time).

|low "sun intensity"|high "sun intensity"|
|-|-|
| image | image |

fu5ha (approved):
Cool :D

PR #1091, opened 2023-08-30
Update to nightly-2023-08-29

LegNeato:
**EDIT**(@eddyb): initial PR by @LegNeato but I added to it to avoid creating another PR (https://github.com/EmbarkStudios/rust-gpu/pull/1091#commits-pushed-9536c3e).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed 🛠
+- [PR#1091](https://github.com/EmbarkStudios/rust-gpu/pull/1091) updated toolchain to `nightly-2023-08-29`
 - [PR#1085](https://github.com/EmbarkStudios/rust-gpu/pull/1085) updated toolchain to `nightly-2023-07-08`
 
 ## [0.9.0]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2041,7 +2041,6 @@ dependencies = [
  "spirv-tools",
  "syn",
  "tempfile",
- "termcolor",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2041,6 +2041,7 @@ dependencies = [
  "spirv-tools",
  "syn",
  "tempfile",
+ "termcolor",
 ]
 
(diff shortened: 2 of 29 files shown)


LegNeato:
This builds successfully for me locally but panics when using GPU stuff (CPU example works fine). I have never touched the compiler or this project so I don't really know what I am doing 😆

LegNeato:
I'm fairly sure the problems are in the type pointer stuff. That machinery was removed in rustc in https://github.com/rust-lang/rust/commit/04303cfb3af3a8019ef913b1a07a94cfb683a5ab

LegNeato:
I managed to get it further along by adding `let ptr = self.pointercast(ptr, self.type_ptr_to(ty));`
to various places. It looks like the above change removes those before calling things like `load` and `store`.

LegNeato:
I also needed to add a new intrinsic `compare_bytes` due to https://github.com/rust-lang/rust/commit/502af03445f045dade14f14e754803f02c2c5e24.

LegNeato:
When I don't add the pointer casts, it compiles but I get:
```
TODO: intcast not implemented yet: val=SpirvValue { kind: Def(14981), ty: 54 } val.ty=Adt { def_id: Some(DefId(0:34076 ~ core[e34c]::option::Option)), align: Align(4 bytes), size: Some(Size(8 bytes)), field_types: [4, 4], field_offsets: [Size(0 bytes), Size(4 bytes)], field_names: None } dest_ty=Integer(32, true) is_signed=false
```

When I do, it gets all the way to compiling the shader but bails on converting a `*void` to a `*i32` IIRC.

eddyb:
Hi, sorry to barge in with those commits, but I went ahead and finished this PR - it turned out to be pretty nasty and I'm not that happy with my approach (it's now conservative when creating pointers, but aggressive when accessing them).

So I really hope you didn't waste too much time trying to make it all work (also, feel free to poke me on Discord etc. any time, esp. if you want any feedback, I tend to mainly get GH notifs for a direct ping, i.e. `@eddyb`).

eddyb (approved):

LegNeato:
Thanks completing this! I like to dive into the deep end in codebases to try to learn them, so the time I spent poking at this was time well spent. Doubly so now that I've seen your changes and can compare them to the mental model I built up.

PR #1101, opened 2023-11-27
Add ignore and no_run to get `cargo test` to pass.

LegNeato:
Without these running `cargo test` on a clean checkout fails.


fornwall (approved):
Thanks!

PR #1102, opened 2023-11-27
rustup: update to nightly-2023-11-26.

LegNeato:


LegNeato:
I'll take a look at the failures today... all tests passed on my machine

LegNeato:
Ah, I thought you were just updating adhoc. If rust-gpu requires nightly, is there a compelling reason to tie it to cutpoints? Perhaps my mental model is wrong about what versions are used where w.r.t rust-gpu and end users. In my mind when folks are using nightly the whole point is to stay close to HEAD...an "old" nightly is the worst of both worlds (not "stable" / tested, not recent).

Tracking HEAD would also make breakages and changes from rustc show up right away and they would more likely be treated as a regression. My impression is that often rust-gpu will be broken, and then weeks down the line you point it out to rustc. By that time the change has stuck and the onus is on you to convince them to revert rather than for them to prove that it should stay in rather than treated as a regression.

Regardless, it might be worth setting up a CI job that updates to the most recent nightly and tries to build rust-gpu automatically to catch issues as they are introduced.

LegNeato:
Ok, for now I am going to just keep updating https://github.com/EmbarkStudios/rust-gpu/pull/1109. Closing this.

Issue #911, opened 2022-08-18
Refactor rustc_codegen_spirv is compiled


watjurk:
# Summary
`rustc_codegen_spirv` is compiled and discovered in a very hacky way. Cargo compiles `rustc_codegen_spirv` because it is inside `spirv-builder`'s Cargo.toml as a dependency, and then `spirv-builder` discovers the `rustc_codegen_spirv`'s dylib by this super hacky function: [`find_rustc_codegen_spirv`](https://github.com/EmbarkStudios/rust-gpu/blob/main/crates/spirv-builder/src/lib.rs#L378-L391), I think this should be avoided.

# Motivation
When I've tried to build something using `spirv-builder` this hacky discovery path was not working. Also `rustc_codegen_spirv` is being build using the same profile as `spirv-builder` and we end up with `rustc_codegen_spirv` being build _without_ optimizations (if we are in debug mode). This is a big problem because compiling `rustc_codegen_spirv` is a one time cost, but while developing one would recompile thier shaders many times.

# Solution
Let's move `rustc_codegen_spirv` into another trait wich will compile `rustc_codegen_spirv` and provide path to it's dylib file.
I've crated an example of how this could be implements: https://github.com/watjurk/spirv-builder-alternative, but here is a short summary:

All of this happens inside [crates/rustc_codegen_spirv](https://github.com/EmbarkStudios/rust-gpu/tree/main/crates/rustc_codegen_spirv):
### `Cargo.toml`:
```toml
[package]
name = "rustc_codegen_spirv_compiler"
```
### `src/lib.rs`:
```rust
pub fn dylib_path() -> PathBuf {
// Compile rustc_codegen_spirv using cargo and return path to it's dylib file.
}
```
### `src/rustc_codegen_spirv`
```
Folder with the `rustc_codegen_spirv` crate.
```

# Future
In the future this approach could be modified without breaking code that relies on `dylib_path` function, for example we can serve pre-builed dylib's of rustc_codegen_spirv and `dylib_path` instead of building `rustc_codegen_spirv` will download one of the pre-build dylib's depending on user's hardware.

eddyb:
I agree it's *noticeably* suboptimal right now, and we have been wondering for a while about alternatives.

However, there are more variations to what we could do (cc @repi @oisyn).

#### How should `rustc_codegen_spirv` be built?
*(:stop_sign: used below to represent "downsides" aka "cons" in a easy-to-spot way)*
[...]

PR #1103, opened 2023-11-27
spirv-builder: build `rustc_codegen_spirv` with the right toolchain via build script.

eddyb:
This has been a long-time coming, I've mostly been putting it off because of how annoying it'd be to implement.
(turns out I was a bit wrong and there's so many more things that *don't* work, than the few that do)

For some background, also see previous discussion in:
- https://github.com/EmbarkStudios/rust-gpu/issues/911

Rough summary of the new approach:
- unknown toolchain builds `spirv-builder` with no special features
  - `spirv-builder`'s build script kicks in and creates a private temporary helper workspace, which:
    - depends on the exact `spirv-builder` being built (via path dependency, to wherever it is)
    - enables the `internal-unstable-trigger-self-build-for-backend` special feature, which in turn is what enables the `rustc_codegen_spirv` optional dependency
  - *however*, the `rust-toolchain.toml` (and `Cargo.lock`) that `rustc_codegen_spirv` needs are still unknown!
    - `spirv-builder`'s build script uses `cargo_metadata` to find its own `rustc_codegen_spirv` dependency, *from the perspective of the helper workspace that enables it via the special feature*
      - <sub>(the reason for this indirection is a bit subtle - roughly speaking, Cargo pretends to not even know about non-workspace packages that are optional dependencies, unless *something in the workspace* causes them to be enabled - so the original workspace using `spirv-builder` is useless for this)</sub>
    - `rust-toolchain.toml` and `Cargo.lock` are copied from the `rustc_codegen_spirv` found with `cargo_metadata`
  - the correct toolchain is then asked to build that temporary helper workspace, causing a nested build of `spirv-builder`, including the `rustc_codegen_spirv` dylib, whose path can be embedded in the outer build
    - of note is that we control this build, and can always make it `--release`, incremental, etc.
- the resulting (outer) `spirv-builder` will use information from its build script, to run the correct `rustup` toolchain, with the correct `rustc_codegen_spirv` dylib path, regardless of the outer toolchain/env vars/etc.

Benefits:
- users never have to worry about `rust-toolchain.toml` or nested workspaces/helper binaries again
- overriding profiles in `Cargo.toml` is no longer needed to ensure the codegen backend gets optimized
- there's no duplicate build of the backend even if both a build script and hot reloading use `spirv-builder`
  - (in fact, I've been tempted to make the build script side mandatory for hot reloading, because it would allow us to enforce the shader interface, and disallow hot reloading when it'd make the types incompatible)
  - or at the very least we could have `SpirvBuilder` generate some APIs to conveniently rebuild *all the same shaders* (making the SPIR-V blobs almost like "self-updating values")

Downsides:
- working on this repository gets weirder/more contrived
  - specifically around `rustc_codegen_spirv`, which now has its own nested workspace
- users can accidentally cause more copies of `rustc_codegen_spirv` to be built than they'd want
- users still get all the language/libcore limitations of the current Rust-GPU toolchain *in their shader code*, but without having a `rust-toolchain.toml` around to run `cargo check` on the host with the same toolchain
  - this might not be a huge deal, and we could add `cargo +nightly-... check` suggestions to errors etc.
- spurious rebuilds can occur, until we sort out all the env vars and directories that such nested builds can be sensitive to

---

If you want to give this a go, remove `rust-toolchain.toml` (and anything you added to e.g. `Cargo.toml` for overriding profiles for build dependency, any weird workspace indirection executables etc.) and add:
```toml
[patch.crates-io]
spirv-builder = { git = "https://github.com/LykenSol/rust-gpu", branch = "auto-self-build" }
```
(assuming you're already depending on `spirv-builder = "0.9"`)


crates/spirv-builder/build.rs
```diff
@@ -0,0 +1,233 @@
...
+    .unwrap();
+    const NESTED_SELF_BUILDER_MAIN_RS: &str = r#"
+fn main() {
+    let dylib_path = spirv_builder::codegen_backend::codegen_backend_dylib_path();
+    println!("cargo:rerun-if-changed={}", dylib_path.display());
+    println!("cargo:rustc-env=SPIRV_BUILDER_CODEGEN_BACKEND_DYLIB_PATH={}", dylib_path.display());
+    println!("cargo:rustc-env=SPIRV_BUILDER_RUSTUP_TOOLCHAIN={}", env!("RUSTUP_TOOLCHAIN"));
+}
+"#;
+    let ws_src_main_rs = ws.join("src/main.rs");
```
eddyb, line 105: New idea (from https://github.com/EmbarkStudios/rust-gpu/issues/911#issuecomment-1966153512) to consider trying:

The diff I'm commenting on _generates a pair of `Cargo.toml` and `src/main.rs`_ - it could be a `-Z script` combined `.rs` file, _forcing Cargo to help out with caching_.

That is, we could manage `~/.cache/rust-gpu` (something respecting `$XDG_CACHE_HOME` in reality), under which we generate a `-Z script` `.rs` file per _choice of contents_ (as in, we can content-address it, i.e. the files just take a name given by their own hash).

That directory of `.rs` files would be relatively miniscule, but Cargo should be able to e.g. automatically GC the target dirs _it created itself (under `~/.cargo`) to build `-Z script` packages_.

PR #1109, opened 2023-12-17
Update to latest nightly

LegNeato:
I'm going to periodically update this to newer versions of nightlies as different commits.

If you intend to use this version, please comment so we can see how many people want the project to track newer nighties in `main`.


LegNeato:
The latest update has some tests failing. I'll look into them in a couple days.

PR #1118, opened 2024-01-23
WIP: update rspirv 0.11 to 0.12

Firestar99:
still very WIP, just opening a PR to be able to ref it in `CHANGELOG.md` :D


crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
```diff
@@ -1357,6 +1358,61 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
...
+            },
+            (OperandKind::LoadCacheControl, Some(word)) => match word.parse() {
+                Ok(x) => inst.operands.push(dr::Operand::LoadCacheControl(x)),
+                Err(()) => self.err(format!("unknown LoadCacheControl {word}")),
+            },
+            (OperandKind::StoreCacheControl, Some(word)) => match word.parse() {
+                Ok(x) => inst.operands.push(dr::Operand::StoreCacheControl(x)),
+                Err(()) => self.err(format!("unknown StoreCacheControl {word}")),
+            },
+            (OperandKind::LiteralFloat, Some(word)) => todo!(),
```
Firestar99, line 1415: I could copy-paste all of the above, but I have no idea what to do here, as there is no such thing as an `dr::Operand::LiteralFloat`

crates/rustc_codegen_spirv/src/symbols.rs
```diff
@@ -118,8 +118,8 @@ const BUILTINS: &[(&str, BuiltIn)] = {
         ("layer_per_view_nv", LayerPerViewNV),
         ("mesh_view_count_nv", MeshViewCountNV),
         ("mesh_view_indices_nv", MeshViewIndicesNV),
-        ("bary_coord_nv", BaryCoordNV),
-        ("bary_coord_no_persp_nv", BaryCoordNoPerspNV),
+        ("bary_coord", BaryCoordKHR),
+        ("bary_coord_no_persp", BaryCoordNoPerspKHR),
```
Firestar99, line 122: I've just removed the `_nv` suffix, which may break some code. Should we instead still allow the suffix and just replace it with the KHR version internally?

Firestar99:
While porting, I've noticed that `rustc_codegen_spirv::spirv_type_constraints::instruction_signatures(Op)` somehow seems to validate a lot of SPIR-V Ops on their input and output types, and that `spirv/grammar/autogen_table.rs` seems to also contain a giant List of auto-generated instructions with all their inputs and outputs. Maybe something could be simplified there?

crates/rustc_codegen_spirv/src/symbols.rs
eddyb: NV aliases are still there in `rspirv::BuiltIn`, no change was needed here: <https://docs.rs/spirv/0.3.0+sdk-1.3.268.0/spirv/enum.BuiltIn.html#associatedconstant.BaryCoordNV>.

Well except this abuses glob imports, hah. The `use BuiltIn::*;` must've been why it started erroring.
But that's not new! You can see that some are referenced `BuiltIn::` without relying on the glob.

eddyb (changes requested):

Firestar99:
Basically ready to merge, apart from `LiteralFloat` still being a `todo!`. If it's essentially unused, should it just be replaced with some kind of error message, and done with it?

eddyb (changes requested):
Ahh, I see you got around to fixing up the `BaryCoord` stuff before I did, nice!

Leaving this review more as a note that I will check the subtler aspects in more depth (so we don't accidentally merge too soon), but what I've seen looks great, thank you so much for working on this!

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
```diff
@@ -1357,6 +1358,61 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
...
+            },
+            (OperandKind::LoadCacheControl, Some(word)) => match word.parse() {
+                Ok(x) => inst.operands.push(dr::Operand::LoadCacheControl(x)),
+                Err(()) => self.err(format!("unknown LoadCacheControl {word}")),
+            },
+            (OperandKind::StoreCacheControl, Some(word)) => match word.parse() {
+                Ok(x) => inst.operands.push(dr::Operand::StoreCacheControl(x)),
+                Err(()) => self.err(format!("unknown StoreCacheControl {word}")),
+            },
+            (OperandKind::LiteralFloat, Some(_word)) => todo!(),
```
eddyb, line 1415: Wow I didn't realize this was so manual. Longer-term might be worth switching to SPIR-T's view of `OperandKind`, which handles immediate operands in a much more dynamic way, and contains a lot of the same information that `.parse()` uses in these `match` arms (but then I'm not sure you could create the typed `rspirv` `Operand` value out of that).

eddyb, line 1415:
```suggestion
            (OperandKind::LiteralFloat, Some(word)) => match word.parse() {
                Ok(v) => inst.operands.push(dr::Operand::LiteralBit32(f32::to_bits(v))),
                Err(e) => self.err(format!("invalid float: {e}")),
            },
```
(or more like the `SpirvType::Float(32) =>` case, either way it would do the same thing - also maybe this should be moved to just under `LiteralNumber` handling?)

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
```diff
@@ -1357,6 +1358,61 @@ impl<'cx, 'tcx> Builder<'cx, 'tcx> {
...
+                Err(()) => self.err(format!("unknown InitializationModeQualifier {word}")),
+            },
+            (OperandKind::LoadCacheControl, Some(word)) => match word.parse() {
+                Ok(x) => inst.operands.push(dr::Operand::LoadCacheControl(x)),
+                Err(()) => self.err(format!("unknown LoadCacheControl {word}")),
+            },
+            (OperandKind::StoreCacheControl, Some(word)) => match word.parse() {
+                Ok(x) => inst.operands.push(dr::Operand::StoreCacheControl(x)),
+                Err(()) => self.err(format!("unknown StoreCacheControl {word}")),
+            },
```
eddyb, line 1414: Wow I didn't realize this was so manual. Longer-term might be worth switching to SPIR-T's view of `OperandKind`, which handles immediate operands in a much more dynamic way, and contains a lot of the same information that `.parse()` uses in these `match` arms (but then I'm not sure you could create the typed `rspirv` `Operand` value out of that).

crates/rustc_codegen_spirv/src/builder/spirv_asm.rs
Firestar99: added that just under `LiteralInteger`

eddyb:
@Firestar99 #1127 just landed, you'll need to rebase (but I should be able to review concurrently).

PR #1122, opened 2024-01-28
spirv-std: only enable `glam`&`num-traits`'s `libm` features on SPIR-V targets.

### LINKED ISSUE #1121: adding spirv-std anywhere in the worskspace, slows down the CPU code due to libm feature
https://github.com/EmbarkStudios/rust-gpu/blob/1a9c8b8ba1dfb4b7dbe26d0fd55265de6b36d00f/crates/spirv-std/Cargo.toml#L14-L15

it was slowing down my cpu code by 3x, specifically the glam Vec3::distance.
it's really bad because new users might enable it in the shaders crate within the workspace early in development, without realizing the impact in performance.

Possible solution:
spirv-std crate should only enable libm for spirv target in Cargo.toml

eddyb:
- fixes #1121

Draft because I'm not sure if we should do some kind of automated testing for this, I can't think of an obvious way to know "slower floating-point algorithms are being used" (also there's https://github.com/EmbarkStudios/rust-gpu/pull/1115 which may conflict).

Testing-wise the best we can do is probably likely this:
```console
$ cargo tree -e features -p spirv-std | rg libm
├── glam feature "libm"
│       └── libm v0.2.8
├── num-traits feature "libm"
│       └── libm feature "default"
│           └── libm v0.2.8
```
(the above is *before* this PR, and there are no `libm` results *after* this PR)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,10 @@ spirv-builder = { path = "./crates/spirv-builder", version = "=0.9.0", default-f
 rustc_codegen_spirv = { path = "./crates/rustc_codegen_spirv", version = "=0.9.0", default-features = false }
 rustc_codegen_spirv-types = { path = "./crates/rustc_codegen_spirv-types", version = "=0.9.0" }
 
+# External dependencies that need to be mentioned more than once.
+num-traits = { version = "0.2.15", default-features = false }
+glam = { version = ">=0.22, <=0.24", default-features = false }
+
 # Enable incremental by default in release mode.
 [profile.release]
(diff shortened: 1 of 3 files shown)


fornwall (approved):

eddyb:
Accidentally found that this is really broken *without* `resolver = "2"`/`edition = "2021"`.
_The ability itself_ to have per-target feature controls was _added then_ (I now remember struggling in 2020 etc.).

Not sure how to control the resolver from `spirv-builder` but I will look around for a solution.

PR #1019, opened 2023-03-25
WIP: VK_EXT_mesh_shader

BeastLe9enD:
This PR adds support for VK_EXT_mesh_shader. At the moment, it depends on my unofficial updated rspirv version. 

An example mesh shader looks like this:
```Rust
#[spirv(mesh_ext(
    threads(1),
    output_vertices = 3,
    output_primitives_ext = 1,
    output_triangles_ext
))]
pub fn my_mesh_shader(
    #[spirv(position)] positions: &mut [Vec4; 3],
    #[spirv(primitive_triangle_indices_ext)] indices: &mut [UVec3; 1],
) {
    unsafe {
        set_mesh_outputs_ext(3, 1);
    }

    positions[0] = Vec4::new(-0.5, 0.5, 0.0, 1.0);
    positions[1] = Vec4::new(0.5, 0.5, 0.0, 1.0);
    positions[2] = Vec4::new(0.0, -0.5, 0.0, 1.0);

    indices[0] = UVec3::new(0, 1, 2);
}
```

PR #1125, opened 2024-02-01
VK_EXT_mesh_shader, version 2

Firestar99:
**!!! builds on top of [PR #1118](https://github.com/EmbarkStudios/rust-gpu/pull/1118), which needs to be merged first**

Adds support for `VK_EXT_mesh_shader`. Rebased from the abandoned [PR #1019](https://github.com/EmbarkStudios/rust-gpu/pull/1019) by @BeastLe9enD. Note that I have split the PR in two parts: [PR #1118](https://github.com/EmbarkStudios/rust-gpu/pull/1118) handles updating to a more recent rspirv version, whereas this PR only handles the additions needed for `VK_EXT_mesh_shader` support.

(Sorry if the diff currently looks horrid, as it currently shows 2 PRs in one. As soon as the other PR is merged, I'll rebase this one and the diff should be way easier to review on github)


crates/spirv-std/src/arch/mesh_shading.rs
```diff
@@ -0,0 +1,66 @@
...
+/// Ceases all further processing: Only instructions executed before
+/// *OpEmitMeshTasksEXT* have observable side effects.
+///
+/// This instruction must be the last instruction in a block.
+///
+/// This instruction is only valid in the *TaskEXT* Execution Model.
+#[spirv_std_macros::gpu_only]
+#[doc(alias = "OpEmitMeshTasksEXT")]
+#[inline]
+pub unsafe fn emit_mesh_tasks_ext(group_count_x: u32, group_count_y: u32, group_count_z: u32) -> ! {
```
Firestar99, line 58: The documentation on `OpEmitMeshTasksEXT` mentions an optional payload pointer, which does not exist on this function. The compiletests contain examples on mesh and task shaders with payloads. But I'm unsure how rust-gpu handles payloads in general, and if this is the correct way to handle them or not.

Firestar99: Resolved it by adding an explicit `emit_mesh_tasks_ext_payload()` that takes a payload param simiarly to raytracing.

PR #1127, opened 2024-02-05
Update spirv-tools

Jake-Shadle:
- Update to spirv-tools 0.10.0
- Use pre-built binaries from spirv-tools-rs


tests/ui/lang/core/ref/member_ref_arg-broken.stderr
```diff
@@ -3,6 +3,7 @@ warning: `#[inline(never)]` function `member_ref_arg_broken::h` needs to be inli
 warning: `#[inline(never)]` function `member_ref_arg_broken::h_newtyped` needs to be inlined because it has illegal argument or return types
 
 error: error:0:0 - OpLoad Pointer <id> '$ID[%$ID]' is not a logical pointer.
+         %39 = OpLoad %uint %38
```
eddyb, line 6: Love to finally have this resolved! We still should clean up some of the formatting so it's easier to understand that this is a `spirv-val` error, but at least now we won't have test output differ between "compiled vs installed" tools.

crates/rustc_codegen_spirv/Cargo.toml
```diff
@@ -56,7 +56,7 @@ sanitize-filename = "0.4"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 smallvec = { version = "1.6.1", features = ["union"] }
-spirv-tools = { version = "0.9", default-features = false }
+spirv-tools = { version = "0.10", default-features = false }
```
eddyb, line 59: If you move this to `Cargo.toml` and have `spirv-tools.workspace = true` instead here, it would theoretically be possible to have the CI script not hardcode `0.10.0`, and instead scan `Cargo.toml` for the version.

There's also stuff like this (`jq` would ofc be replaced by e.g. using the `cargo_metadata` crate etc.):
```console
$ cargo tree -p spirv-tools --depth 0
spirv-tools v0.10.0
$ cargo metadata --format-version 1 | jq -r '.packages[] | select(.name == "spirv-tools").version'
0.10.0
```

For now it seems fine to land this PR as-is, just thinking out loud mostly.

.github/install-tools/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "install-tools"
+edition = "2021"
+version = "0.1.0"
+publish = false
+
+[dependencies]
+tar = "0.4"
+zstd = "0.13"
+
+[workspace]
```
eddyb, line 11: Minor nit: missing newline at the end (also I would put `[workspace]` at the top so it's easy to see it self-excludes from the outer workspace).

(and maybe call it `install-spirv-tools`? the new `cargo -Zscript` stuff would be great for this kind of helper, but we don't have to switch right away ofc)

eddyb (changes requested):
Mostly just the `Cargo.toml` nitpick, but you also need to add a changelog entry, can you make sure to mention "Vulkan SDK 1.3.275" as well in it?

(I want to be more cognizant of specific Vulkan SDK versions just so as much as possible can be kept in sync without having to guess or approximate etc.)

CHANGELOG.md
```diff
@@ -30,15 +32,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed 🛠
+
 - [PR#1101](https://github.com/EmbarkStudios/rust-gpu/pull/1101) Add `ignore` and `no_run` to documentation to make `cargo test` pass.
 - [PR#1112](https://github.com/EmbarkStudios/rust-gpu/pull/1112) updated wgpu and winit in example runners
 - [PR#1100](https://github.com/EmbarkStudios/rust-gpu/pull/1100) updated toolchain to `nightly-2023-09-30`
 - [PR#1091](https://github.com/EmbarkStudios/rust-gpu/pull/1091) updated toolchain to `nightly-2023-08-29`
 - [PR#1085](https://github.com/EmbarkStudios/rust-gpu/pull/1085) updated toolchain to `nightly-2023-07-08`
+- [PR#1127](https://github.com/EmbarkStudios/rust-gpu/pull/1127) updated `spirv-tools` to `0.10.0`, which follows `vulkan-sdk-1.3.275`.
```
eddyb, line 41: Don't remember explicitly deciding on this, but the entire changelog is most-recent-first.

eddyb (changes requested):

CHANGELOG.md
Jake-Shadle: Yah, I noticed that, sorry.

eddyb (approved):

PR #1128, opened 2024-02-05
Use single test job for branch protections

Jake-Shadle:


.github/workflows/ci.yaml
```diff
@@ -127,6 +127,14 @@ jobs:
           cargo apk build -p example-runner-wgpu --lib --features use-installed-tools --no-default-features
           echo "::endgroup::"
 
+  # This just allows us to branch protect on this one job rather needing to change
+  # it if the test matrix changes
+  test_success:
```
eddyb, line 132: We could probably give this a name like "All CI" and include Lint/cargo-deny too (and maybe put it first), remove mergify, etc. - I'll land this as-is so we can do it later.

eddyb (approved):

PR #1129, opened 2024-02-07
Fix bit rotates by using `== 0` instead of `!= 0` for `mask_is_zero`.

### LINKED ISSUE #1062: u64 bitwise rotate works incorrectly
### Expected Behaviour
Bitwise rotating works correctly.

I was porting SHA512 and this function worked differently on CPU and GPU:
```rust
fn big_sigma0(x: u64) -> u64 {
    x.rotate_right(28) ^ x.rotate_right(34) ^ x.rotate_right(39)
}
```

## System Info
 - Rust: rustc 1.69.0-nightly (44cfafe2f 2023-03-03)
 - OS: NixOS Linux 5.15.110
 - GPU: AMD Radeon RX 6800 XT
 - GPU Driver: amdvlk
 - SPIR-V: v2023.2 unknown hash, 1980-01-01T00:00:00 (incorrect time because of NixOS)

eddyb:
- fixes #1062 

(no idea what happened here, the variable name alone should've made this easy to spot, but it took @chrisliebaer seeing the broken expression in SPIRV-Cross output, to narrow it down)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,12 +30,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed 🛠
-- [PR#1127](https://github.com/EmbarkStudios/rust-gpu/pull/1127) updated `spirv-tools` to `0.10.0`, which follows `vulkan-sdk-1.3.275`.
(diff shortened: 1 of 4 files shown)


fornwall (approved):
Nice!

PR #1134, opened 2024-02-17
Update to Rust 1.76.0

LegNeato:
This updates to the last nightly tagged as rust 1.76, which was released to stable.

This is a version of https://github.com/EmbarkStudios/rust-gpu/pull/1109 that stops on the last nightly of the current stable release rather than trying to track the most recent nightly.


LegNeato:
Not sure about the error, ` RUSTDOCFLAGS=-Dwarnings cargo doc --no-deps` passes when run locally.

LegNeato:
This is a version of https://github.com/EmbarkStudios/rust-gpu/pull/1109 that stops on the last nightly of the current stable release rather than trying to track the most recent nightly.

LegNeato:
it looks like this nightly might be a bit buggy as it is ICEing on non-rust-gpu stuff. Should we fast-forward to `nightly-2023-12-31` (which we know works from the other PR) or some other version?

LegNeato:
Actually, the forward porting touched this area...so either they cut on a buggy commit or I got the porting wrong. Need to investigate.

PR #1143, opened 2024-04-12
Account for signedness in memset const pat

### LINKED ISSUE #1061: Array init with 0i32 uses 0u32 instead.
Initializing arrays seems to replace 0i32 with 0u32 when using [0i32; N].

Example (see https://github.com/charles-r-earp/rust-gpu/tree/array-init for compiletests).
```
; SPIR-V
; Version: 1.3
; Generator: Google rspirv; 0
; Bound: 11
; Schema: 0
               OpCapability Shader
               OpCapability Float64
               OpCapability Int64
               OpCapability Int16
               OpCapability Int8
               OpCapability ShaderClockKHR
               OpExtension "SPV_KHR_shader_clock"
               OpMemoryModel Logical Simple
               OpEntryPoint Fragment %1 "main" %o
               OpExecutionMode %1 OriginUpperLeft
          %3 = OpString "/home/charles/Documents/rust/rust-gpu/tests/ui/lang/core/array/init_array_i32.rs"
               OpSource Unknown 0 %3 "// Test creating an array.
// build-pass

use spirv_std::macros::spirv;

#[spirv(fragment)]
pub fn main(o: &mut i32) {
    let array = [0i32; 4];
    *o = array[0];
}
"
               OpName %o "o"
               OpDecorate %o Location 0
        %int = OpTypeInt 32 1
%_ptr_Output_int = OpTypePointer Output %int
       %void = OpTypeVoid
          %7 = OpTypeFunction %void
          %o = OpVariable %_ptr_Output_int Output
       %uint = OpTypeInt 32 0
     %uint_0 = OpConstant %uint 0
          %1 = OpFunction %void None %7
         %10 = OpLabel
               OpLine %3 9 4
               OpStore %o %uint_0
               OpNoLine
               OpReturn
               OpFunctionEnd
```
A 0u32 is stored to `o` instead of the expected 0i32, which fails to compile. 

What works:
- f32, this seems to be an issue with signed integers
- `<[T; N]>::default()`
- An explicit list, `[0i32, 0i32, ..]`
-  Using 1 instead of 0, `[1i32; 1]`

JulianKnodt:
Fixes #1061

Thanks to LegNeato for pointing out what was wrong

### DIFF
--- a/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
+++ b/crates/rustc_codegen_spirv/src/builder/builder_methods.rs
@@ -190,7 +190,7 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
         match *ty {
             SpirvType::Void => self.fatal("memset invalid on void pattern"),
             SpirvType::Bool => self.fatal("memset invalid on bool pattern"),
-            SpirvType::Integer(width, _signedness) => match width {
+            SpirvType::Integer(width, false) => match width {
                 8 => self.constant_u8(self.span(), fill_byte).def(self),
                 16 => self
                     .constant_u16(self.span(), memset_fill_u16(fill_byte))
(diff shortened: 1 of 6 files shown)


crates/rustc_codegen_spirv/src/builder/builder_methods.rs
```diff
@@ -315,7 +338,7 @@ impl<'a, 'tcx> Builder<'a, 'tcx> {
         } else {
             for index in 0..count {
                 let const_index = self.constant_u32(self.span(), index as u32);
-                let gep_ptr = self.gep(pat.ty, ptr, &[const_index]);
+                let gep_ptr = self.inbounds_gep(pat.ty, ptr, &[const_index]);
```
JulianKnodt, line 341: I added this because I assume that const memsets will always be in a valid range, but I'm not sure.

LegNeato:
Awesome, that was quick!!!! 😍

